PR #1, opened 2017-08-28
WIP: scalable k-means

TomAugspurger:
Implements [`k-means||`](https://arxiv.org/abs/1203.6402), a scalable k-means initialization alternative to K-Means++

### DIFF
--- a/benchmarks/k_means_kdd.py
+++ b/benchmarks/k_means_kdd.py
@@ -1,0 +1,158 @@
+"""
+Script for running K-Means clustering on the KDD-Cup dataset.
+"""
+import argparse
+import glob
+import logging
+import os
+import string
+import sys
+from timeit import default_timer as tic
+
+import dask.array as da
+import dask.dataframe as dd
+import pandas as pd
+import requests
+import s3fs
+import sklearn.cluster as sk
+from daskml.cluster import KMeans
+from distributed import Client
+from sklearn.datasets import get_data_home
+
+logger = logging.getLogger()
+
+try:
+    import coloredlogs
+except ImportError:
+    pass
+else:
+    coloredlogs.install()
+
+URL = "http://archive.ics.uci.edu/ml/machine-learning-databases/kddcup99-mld/kddcup.data.gz"  # noqa
+
+
+def parse_args(args=None):
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument('-s', '--scheduler-address', default=None,
+                        help="Address for the scheduler node. Runs locally by "
+                             "default.")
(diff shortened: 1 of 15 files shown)


benchmarks/k_means.py
```diff
@@ -0,0 +1,109 @@
...
+"""
+import argparse
+import logging
+import sys
+try:
+    import coloredlogs
+except ImportError:
+    pass
+else:
+    coloredlogs.install()
```
mrocklin:
```
mrocklin@carbon:~/workspace/dask-ml$ python benchmarks/k_means_kdd.py 
  File "benchmarks/k_means_kdd.py", line 12, in <module>
    import coloredlogs
ModuleNotFoundError: No module named 'coloredlogs'
```

benchmarks/k_means_kdd.py
```diff
@@ -0,0 +1,138 @@
...
+        logger.info("Using distributed mode")
+        client = Client(args.scheduler_address)
+        logger.info(client.scheduler_info())
+        fs = s3fs.S3FileSystem()
+        if fs.exists("dask-data"):
+            logger.info("Using cached dataset")
+        else:
+            logger.info("Uploading to cloud storage")
+            upload(local, fs)
+        path = "dask-data/kddcup/kdd.parq/"
```
mrocklin: I tried running this with a dask cluster on localhost and ran into errors here

TomAugspurger: Whoops, missing the `s3://`.

Some context, the kdd-cup dataset is the largest dataset used in the k-means|| paper, but it isn't that big... I can cluster the entire dataset just fine on my laptop.

I just setup a cluster to benchmark k-means on the airlines dataset.

mrocklin: We shouldn't necessarily need to engage s3 here though, no?  I wanted to run this locally just to see the diagnostic dashboard (things look great by the way) and ran into issues here

TomAugspurger:
Oh, sorry mixing up my examples. Yes... that should be `path = local` I
think. I'll push an update.

benchmarks/k_means_kdd.py
TomAugspurger: Oh I remember now. I assumed that using the distributed scheduler implied a remote cluster. I'll have to refactor it a bit more then.

TomAugspurger:
For some reason, travis is pickup up https://github.com/dask/dask-ml/commit/a2541e4cc775f27e9cef9e5c8cb4378f5ac821bd as the merge commit, which is for #15 . Opening a new PR.

PR #9, opened 2017-09-27
MinMaxScaler

daniel-severo:
- [X] Implement `fit` method.
- [X] Implement `inverse_transform` method.
- [x] Implement `_handle_zeros_in_scale` method.
- [X] Add support for dask dataframes.
- [x] Implement columns slicer.


daniel-severo:
@TomAugspurger I still need to implement a few things (see the checklist). My main concern is, how should the [`_handle_zeros_in_scale`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/preprocessing/data.py#L61-L76) function be implemented? I'm sure this can be done in a dasky way, I just can't get my head around it. All the solutions I came up with were not to my liking

TomAugspurger:
> My main concern is, how should the _handle_zeros_in_scale function be implemented?

IIUC, then this should work

```python
def _handle_zeros_in_scale(scale, copy=True):
    scale = scale.copy()  # maybe raise an error if copy is False
    scale[scale == 0.0] = 1.0  # this requires a recent version of dask
    return scale
```

We want to avoid calls like `np.isscalar(X)`, since that would evaluate the graph.

> I think that most arguments in sklearn use an underscore at the end, so I took the liberty of automating the tests a bit more in this sense. Maybe we can evolve this in a way that we end up writing fewer code for the same testing coverage?

Yep, I planned to do that soon. If you can start that'd be great, otherwise I'll get to it next week.

> The implementation of the column slicer (apply scaler to supplied columns only) I think can be handled by masking. Will try it tomorrow.

Indeed. No worries if you can't get to this.

> Ah ... need to get the python2 support down also.

Yeah, maybe we just go py3 only :)

daniel-severo:
>    scale[scale == 0.0] = 1.0  # this requires a recent version of dask

I wasn't aware I could slice and dice things with dask. Ty.

> Yep, I planned to do that soon. If you can start that'd be great, otherwise I'll get to it next week.

I've never automated tests in this way. I want to take a wack at it. I'll do it in another PR and we can discuss it.

> Yeah, maybe we just go py3 only :)

Ok, I'm gonna ignore py2 support for now (thank god).

TomAugspurger:
> Ok, I'm gonna ignore py2 support for now (thank god).

For now, you can just edit `.travis.yml` to remove "PYTHON=2.7" from the matrix.

daniel-severo:
I think the best option to handle column-aware scalers is to program oriented towards dask arrays and use df.values to convert dataframes to arrays. But first we must [fix this](https://github.com/dask/dask/issues/2727) and

daniel-severo:
@TomAugspurger what do you think? A few notes:

1. The variable `self.n_seen_samples_` must exist for compatibility with sklearn. I recently found out that dataframes don't carry around the meta information regarding the number of records and that `len(df)` actually computes the dataframe. Hence, I made a change such that `self.n_seen_samples = np.nan` instead of `len(X)`.

2. The cleanest way I found for implementing the column-aware scaler was using `dd.merge`. Not sure if this is the best way performance-wise.

If dataframes had the meta info. of it's record size, we could cast it to a dask array and life would be easier. 

:/

OctavioBR:
LGTM. Well done 👌

daskml/preprocessing/data.py
```diff
@@ -45,3 +48,71 @@ def inverse_transform(self, X, copy=None):
         if self.with_mean:
             X += self.mean_
         return X
+
+
+class MinMaxScaler(skdata.MinMaxScaler):
+
+    def __init__(self, feature_range=(0, 1), copy=True, columns=None):
+        super().__init__(feature_range, copy)
+        self._columns = columns
```
TomAugspurger: I think this can be a public attribute `.columns`.

daskml/preprocessing/data.py
```diff
@@ -45,3 +48,71 @@ def inverse_transform(self, X, copy=None):
...
+    def partial_fit(self, X, y=None):
+        raise NotImplementedError()
+
+    def transform(self, X, y=None, copy=None):
+        _X = slice_columns(X, self._columns)
+        _X *= self.scale_
+        _X += self.min_
+
+        if isinstance(_X, dd.DataFrame) and self._columns:
+            return dd.merge(_X, X.drop(self._columns, axis=1),
```
TomAugspurger: How about

```python
for column in self._columns:
    X[column] = _X[column]
```

daskml/preprocessing/data.py
```diff
@@ -45,3 +48,71 @@ def inverse_transform(self, X, copy=None):
...
+    def inverse_transform(self, X, y=None, copy=None):
+        if not hasattr(self, "scale_"):
+            raise Exception("This %(name)s instance is not fitted yet. "
+                            "Call 'fit' with appropriate arguments before "
+                            "using this method.")
+        _X = slice_columns(X, self._columns)
+        _X -= self.min_
+        _X /= self.scale_
+
+        if isinstance(_X, dd.DataFrame) and self._columns:
```
TomAugspurger: Likewise here.

daskml/utils.py
```diff
@@ -0,0 +1,18 @@
+import dask.dataframe as dd
+import dask.array as da
+
+
+def slice_columns(X, columns):
+    if isinstance(X, dd.DataFrame):
+        return X[columns or list(X.columns)]
```
TomAugspurger: Maybe

```python
if columns is None:
    columns = X.columns
```

Otherwise this will fail for `columns=Index(['a', 'b'])`, since `bool(columns)` is ambiguous.

daskml/utils.py
```diff
@@ -0,0 +1,18 @@
...
+    else:
+        return X
+
+
+def handle_zeros_in_scale(scale):
+    scale = scale.copy()
+    if isinstance(scale, da.Array):
+        scale[scale == 0.0] = 1.0
+    elif isinstance(scale, dd.Series):
+        scale = scale.map(lambda x: 1 if x == 0 else x)
```
TomAugspurger: I suspect it'll be faster to use `scale.where`:

```python
In [43]: s = dd.from_pandas(pd.Series([0, 1, 0, 2, 3]), 2)

In [44]: s.where(s != 0, 1).compute()
Out[44]:
0    1
1    1
2    1
3    2
4    3
dtype: int64
```

daniel-severo:
Was the overall ~architecture~ how you expected it to be? I really wish that I could abstract the type (`da.Array` and `dd.DataFrame`) from the class (i.e. in a way that we wouldn't have to do if elses for everything).

But I think the fact that dataframes don't have meta information on the record size makes it quite a hassle.

TomAugspurger:
>  I really wish that I could abstract the type (da.Array and dd.DataFrame) from the class (i.e. in a way that we wouldn't have to do if elses for everything).

We could use the multipledispatch library. I'm inclined to put up with the isinstance checks for the moment though.

I'll take one last look later today, but I think things are good.

daskml/preprocessing/data.py
```diff
@@ -45,3 +48,73 @@ def inverse_transform(self, X, copy=None):
...
+                 handle_zeros_in_scale(data_range))
+
+        to_persist["data_min_"] = data_min
+        to_persist["data_max_"] = data_max
+        to_persist["data_range_"] = data_range
+        to_persist["scale_"] = scale
+        to_persist["min_"] = feature_range[0] - data_min * scale
+        to_persist["n_samples_seen_"] = np.nan
+
+        values = persist(*to_persist.values(), cache=self._cache)
```
mrocklin: Cache is almost never used any longer with Dask.  I'm curious what the objective here was and if there is a way to replace this with something else.

TomAugspurger: I used it in my implementation of `StandardScaler`. Looking back now, it seems unnecessary. The call to `persist` should be sufficient to ensure that the values don't have to be re-computed, correct?

mrocklin: Yes, `persist` will return dask collections that are equivalent to what were put in, except that now all of the parts will be fully completed.

daniel-severo: Great. I'll remove it from both of them. Ty.

TomAugspurger:
I just pushed a couple commits redoing some of the testing stuff. I think your earlier version iterated over `self` instead of one of the estimators.

I added an `assert_estimator_equal` method, along with some tests for it to ensure we're doing the right thing. I'll rewrite the older tests to use this, hopefully this weekend.

tests/test_utils.py
```diff
@@ -13,6 +16,45 @@
...
+        exclude = set(exclude)
+
+    assert (set(left_attrs) - exclude) == set(right_attrs) - exclude
+
+    for attr in set(left_attrs) - exclude:
+        l = getattr(left, attr)
+        r = getattr(right, attr)
+        if isinstance(l, (np.ndarray, da.Array)):
+            assert_eq_ar(l, r, **kwargs)
+        elif isinstance(l, (pd.core.generic.NDFrame, dd._Frame)):
```
daniel-severo: Any reason why we can't compare against `pd.DataFrame` instead of `pd.core.generic.NDFrame`?
Same for `dd.DataFrame` and `dd._Frame`.

TomAugspurger:
Those are the base classes that include Series too.

________________________________

TomAugspurger:
Thanks @daniel-severo, I'll rewrite the other tests in a new PR.

PR #11, opened 2017-09-29
Added Imputer with mean and median strategies.

daniel-severo:


daniel-severo:
Waiting for https://github.com/dask/dask-ml/pull/9 to finish so we can use the newly created utils.

TomAugspurger:
All done with #9!

TomAugspurger:
I'm going to merge #16 if that's OK. You should just have to adjust the import of `assert_estimator_equal`.

daniel-severo:
@TomAugspurger yeah. This week has been crazy at the office. I'm only gonna finish this PR sometime next week.

TomAugspurger:
No worries!

daniel-severo:
@TomAugspurger a few notes:

1. In order to compute the `statistics_` I used masked arrays for array inputs, but for dataframes I "hacked" it by inserting `np.nan` values. It works since dataframe methods end up ignoring `np.nan` in most cases.
2. I noticed you used `da.vstack(X).T` when implementing the `QuantileTransformer`, why not use `da.hstack(X)`?
3. Looks like dask doesn't have a `median` method :(

mrocklin:
> Looks like dask doesn't have a median method :(

Median is very hard to do in parallel.  Consider the approximate quantile functions.

dask_ml/preprocessing/imputation.py
```diff
@@ -0,0 +1,90 @@
...
+        return da.isnull(X) if v == "NaN" or np.isnan(v) else X == v
+
+
+class Imputer(skimputation.Imputer):
+    def __init__(self, missing_values="NaN", strategy="mean",
+                 axis=0, verbose=0, copy=True, columns=None):
+        super().__init__(missing_values="NaN", strategy="mean",
+                         axis=0, verbose=0, copy=True)
+        self.columns = columns
+        self.missing_values = missing_values
```
TomAugspurger: Does calling `super()` set all these for you?

dask_ml/preprocessing/imputation.py
```diff
@@ -0,0 +1,90 @@
...
+
+        values = persist(*to_persist.values())
+        for k, v in zip(to_persist, values):
+            setattr(self, k, v)
+
+        return self
+
+    def _dense_fit(self, _masked_X, strategy, missing_values):
+        if strategy == "mean":
+            estimator = (lambda x: x.mean() if isinstance(x, dd._Frame)
```
TomAugspurger: It'd be nice to avoid this lambda if possible, they can lead to strange issues (e.g. [here](https://github.com/dask/dask/issues/2774#issuecomment-336918011))

How about something like

```python
    def _dense_fit(self, _masked_X, strategy, missing_values):
        if strategy == "mean":
            if isinstance(_masked_X, dd._Frame):
                return _masked_X.mean()
            else:
                return da.ma.filled(_masked_X.mean(0))
        else:
            raise NotImplementedError()
```

though this may get a bit unweildy if we have many strategies.

dask_ml/preprocessing/imputation.py
daniel-severo: > Does calling super() set all these for you?

I'll check.

I was about to finish up the implementation of median when I noticed that if your series has `np.nan` values and you call `dd.Series.quantile` the output is also `np.nan` . Looks like dask doesn't drop them automatically (pandas does) @mrocklin .

I'm gonna insert a call to `dropna` before I compute the quantile, that should solve it for dataframes at least.

But yeah, I'll finish up and we can finish the other methods some other time. It would be nice to let people include any estimator as the `strategy` argument (as long as it is dask-compatible). I want to start working on the ALS estimator :)

TomAugspurger:
Thanks, I opened https://github.com/dask/dask/issues/2786 for the
Series.quantile issue. Dropping them explicitly for now seems fine.

daniel-severo:
Since we are at it: what do you think is the best way to calculate the `statistics_` for dataframe inputs when `strategy == "median"`? I realize that I need to apply it column-wise, but I can't seem to wrap my head around the correct usage of a dask dataframe series.

I wanted something like:

```python
construct_dask_series({c:df[c].dropna().quantile(0.5) for c in df.columns})
```

But I think this isn't supported in dask. I can use a `pd.Series` object and use `dd.from_pandas`, but I don't think this is wise.

daniel-severo:
I had to do some engineering to get the `statistics_` attribute as a pandas series. Support for dask array inputs with median strategy doesn't look to straight forward to me. The `da.ma` module is still in it's infancy.

Basically this PR will give us the Imputer scaler for a mean and median strategies, but only for dataframes in the later case.

TomAugspurger:
@daniel-severo I pushed a few changes:

1. `statistics_` is now concrete, instead of a lazy dask.array / series
2. `.transform` now returns an array / dataframe of the same shape as the input. I think this will be a more useful convention of users. We apply the transformation to a subset of the columns, and all other columns are passed through untouched
3. Changed the chunks in the test (`chunks=2` had 50 blocks, each with a chunksize of 2). This will make the tests a bit snapier
4. Changed the type of `.statistics_` when the strategy is a `median`. Always having an array or Series here will be useful.

This will fail, until https://github.com/dask/dask/pull/2810 and https://github.com/dask/dask/pull/2791 are merged (later today probably). We could put conditional logic in there, but at the moment I'm comfortable requiring dask master for the test suite to pass.

And finally finally, this will fail since I broke your handling of `missing_values` other than NaN with dataframes. I'm OK with just supporting NaN when the input is a DataFrame, since I think that's all that pandas users will expect.

TomAugspurger:
FYI, I think I'm going to do the release before merging this. We'll keep a quick pace of releases over the month or two though.

daniel-severo:
Great! Let's wait for the PRs to merge. I can finish up whatever extra mods we need tomorrow.

My contribution throughput will be pretty much stable with some fading every 2 weekends. Looking forward to see this project take flight :)

daniel-severo:
Waiting on https://github.com/dask/dask/pull/2810 and https://github.com/dask/dask/pull/2791 so that tests will pass and we can continue.

For now I'll leave it be.

TomAugspurger:
Great, thanks!

MaxBenChrist:
Hi there, 

I am in search of an imputer on dask data frames and stumbled over this pr. Now, before I implement this by myself from scratch, maybe I can help finishing this pr here? 
From what I understand, the two quoted dask prs are already merged into the dask master, right?
What needs to be done on this imputer pr to finish it?

TomAugspurger:
Rebased, we'll see if everything still passes.

mrocklin:

daniel-severo:
I'm back. Had some personal problems. Hope I can help out here again.

If I'm not mistaken we rebased this PR and if nothing broke we could merge it.

I'll check it out.

PR #12, opened 2017-09-30
Added scipy to setup.py

daniel-severo:
While trying to run the tests I noticed that we need scipy and it wasn't in setup.py.


TomAugspurger:
Thanks.

PR #13, opened 2017-10-02
TST: Refactor to use new testing tools

TomAugspurger:


daniel-severo:
I think we might have made a mix-up with `tests/test_utils.py`

There are functions that are testing `daskml/utils.py` (e.g. `test_slice_columns`, `test_handle_zeros_in_scale`). Others are functions that are utilities for writing tests (e.g. `assert_estimator_equal`)

TomAugspurger:
Yep, I decided to throw the definition of `assert_estimator_equal` into `tests/test_utils` at the last moment, so that pytest's assertion rewriting would take place.

I found out afterwards that pytest lets you register specific modules for rewriting asserts. I'll make a PR tomorrow moving stuff so that the definition of `assert_estimator_equal` is in `daskml/utils.py`.

PR #19, opened 2017-10-09
ENH: Adds K-Means

TomAugspurger:
The benchmarks are still a WIP.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,6 @@ docs/day_1
 *.parquet
 *.gz
 docs/source/examples/susy
+.DS_Store
+/benchmarks/k_means.csv
+/benchmarks/k_means.png
--- a/benchmarks/base.py
+++ b/benchmarks/base.py
@@ -1,0 +1,12 @@
+import argparse
+
+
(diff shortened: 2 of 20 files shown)


TomAugspurger:
@mrocklin do you have any clues on [this failure](https://travis-ci.org/dask/dask-ml/jobs/286114439#L823)? Is having non-str keys in `dask.Delayed.dask` problematic? If so, I'll try to create a minimal example.

The dask in question comes from https://github.com/dask/dask-ml/pull/19/files#diff-91207982b39f2eb6ed758c872befe481R20

mrocklin:
That check is designed to verify that the ShareDict has sensible names.  Note that these keys don't correspond to the keys of the task graph, but rather to the over-arching keys of the sharedict.  When we don't use sharedict.merge well these end up resorting to numeric ids, which is a sign of graph uncleanliness.   If you wanted to change that check to check that the keys are not numeric then that might be a good compromise.

TomAugspurger:
I'm going to merge this and continue to tinker with it as needed. I'll do some benchmarks later today.

PR #23, opened 2017-10-11
ENH: Added QuantileTransformer

TomAugspurger:
This mimics the scikit-learn implementation. The primary difference is that our
version uses *apporiximate* quantiles. I suspect this is a good default for
daskml. It'd be nice to implement an exact version as well, though I'll wait for
interest from users first.

Benchmark:

```python
In [1]: import numpy as np
   ...: import dask.array as da
   ...:
   ...: import sklearn.preprocessing as pp
   ...: import daskml.preprocessing as dp
   ...:
   ...: X = da.random.random(size=(10_000_000, 10), chunks=1_000_000)
   ...: x = X.compute()
   ...:
   ...: qt = pp.QuantileTransformer()
   ...: dqt = dp.QuantileTransformer()
   ...:

In [2]: %time qt.fit(x)
CPU times: user 5.92 s, sys: 961 ms, total: 6.88 s
Wall time: 7.01 s
Out[2]:
QuantileTransformer(copy=True, ignore_implicit_zeros=False, n_quantiles=1000,
          output_distribution='uniform', random_state=None,
          subsample=100000)

In [3]: %time dqt.fit(X)
CPU times: user 9.87 s, sys: 702 ms, total: 10.6 s
Wall time: 1.91 s
Out[3]:
QuantileTransformer(copy=True, ignore_implicit_zeros=False, n_quantiles=1000,
          output_distribution='uniform', random_state=None,
          subsample=100000)

In [4]: %time _ = qt.transform(x)
CPU times: user 23.9 s, sys: 4.52 s, total: 28.4 s
Wall time: 29.1 s

In [5]: %time _ = dqt.transform(X).compute()
CPU times: user 33.3 s, sys: 6.1 s, total: 39.4 s
Wall time: 8.4 s
```

I haven't profiled how much of that is due to parallelism vs. approximate
quantiles.

I also want to look a bit more closely at the difference in
quantiles (presumably due to using approximate quantiles). The maximum
difference is just below 0.02 on the test dataset.

### DIFF
--- a/daskml/preprocessing/data.py
+++ b/daskml/preprocessing/data.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+import multiprocessing
 
 import dask.array as da
 import dask.dataframe as dd
@@ -135,9 +136,8 @@ class QuantileTransformer(skdata.QuantileTransformer):
         if isinstance(X, (pd.DataFrame, dd.DataFrame)):
             X = X.values
         if isinstance(X, np.ndarray):
-            # TODO: What should the package policy be on inferring
-            # chunksizes?
-            X = da.from_array(X, chunks=10000)
+            C = len(X) // multiprocessing.cpu_count()
+            X = da.from_array(X, chunks=C)
 
         rng = check_random_state(self.random_state)
         # TODO: non-float dtypes?
--- a/daskml/preprocessing/data.py
+++ b/daskml/preprocessing/data.py
@@ -1,10 +1,14 @@
 from collections import OrderedDict
 
-from dask import persist
 import dask.array as da
 import dask.dataframe as dd
+from dask import persist, compute
 import numpy as np
+import pandas as pd
+from scipy import stats
 from sklearn.preprocessing import data as skdata
+from sklearn.utils.validation import check_random_state
+
 from daskml.utils import handle_zeros_in_scale, slice_columns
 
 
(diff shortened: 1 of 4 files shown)


TomAugspurger:
cc @ogrisel if you're interested.

FWIW, the way QuantileTransformer was structured made it very easy to subclass and override the private methods that do the actual work. Not sure if that was the intent :)

daskml/preprocessing/data.py
```diff
@@ -116,3 +120,93 @@ def inverse_transform(self, X, y=None, copy=None):
...
+    __doc__ = __doc__ + '\n'.join(
+        skdata.QuantileTransformer.__doc__.split("\n")[1:])
+
+    def _check_inputs(self, X, accept_sparse_negative=False):
+        if isinstance(X, (pd.DataFrame, dd.DataFrame)):
+            X = X.values
+        if isinstance(X, np.ndarray):
+            # TODO: What should the package policy be on inferring
+            # chunksizes?
+            X = da.from_array(X, chunks=10000)
```
TomAugspurger: @mrocklin do you have any thoughts on choosing good, default chunk sizes? Is this even something we should be doing?

The use-case here is someone with (not too big) NumPy arrays, who want to drop this transformer in for the parallelism. Perhaps I should add a `chunks` keyword to the `__init__`, and require that if a NumPy array is used?

mrocklin: Perhaps divide the shape by the number of logical cores on the machine?

ogrisel:
This looks good, but the fact that it's easy to subclass is not guaranteed in the future as it's considered private API (although it's unlikely to change).

I think it would be worth to add a CI job configuration to test against any pre-prelease of scikit-learn (e.g. using `pip install --pre scikit-learn` or even against a nightly snapshot of the master although we do not provide nightly builds for linux yet. scikit-learn is probably cheap enough to build from source though.

In the travis conf of scikit-learn we habe an "allowed failure" build with dev versions of our upstream dependencies numpy and scipy to be able to detect any problem caused by an upstream change before it's released in a stable version.

TomAugspurger:
Understood about the subclassing :) I wouldn't expect scikit-learn to try
to avoid breaking there. At the moment, I'd rather have dask-ml break
loudly when something changes in scikit-learn, rather than subtle
differences creeping in due to us not copying over changes correctly.

I will set up a CI run against scikit-learn master.

PR #30, opened 2017-10-12
Reorganize documentation

### LINKED ISSUE #29: Docs reorganization
I've been walking through the documentation and had a few notes.  I'd be happy to make these changes but wanted to check in before I submitted anything.  

The separation between single machine and distributed learning seems odd to me.  Many of the topics listed in single machine (grid search, pipelining, possibly even incremental learning) are still relevant when on a cluster.  

I might instead flatten the TOC to just remove the single-machine/distributed distinction, and give all of the subsections their own home.  This might flatten the TOC to something like the following:

1.  Pipelines
2.  Hyper-parameter search
3.  Incremental learning
4.  Generalized Linear Models
5.  Joblib
6.  XGBoost
7.  Clustering
8.  Examples
9.  API

I've also found it pleasant recently to start sections with the API relevant for that section.  [Futures docs](http://dask.pydata.org/en/latest/futures.html) for an example.  This gives a quick list at the beginning of each section on the API relevant for that section.  Those functions still link to the main API doc page.

Any thoughts or objections to this reorganization @TomAugspurger ?

mrocklin:
This breaks out sections based on algorithm type rather than on
single/distributed.

Fixes #29

This still needs substantial work, both in fixing up API docs and
fleshing out content in the various sections.

I've pushed this to the `dask` fork rather than my own as a signal that others are more than welcome to collaborate on this.


mrocklin:
Your review of recent changes would be welcome @TomAugspurger

daskml/xgboost.py
```diff
@@ -0,0 +1 @@
+from dask_xgboost import *  # noqa
```
TomAugspurger: Thoughts on making xgboost optional? AFAICT, xgboost doesn't upload wheels.

docs/source/joblib.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Dask.distributed integrates with Joblib_ by providing an alternative
+cluster-computing backend, alongside Joblib's builtin threading and
+multiprocessing backends.
+
+Joblib_ is a library for simple parallel programming primarily developed and
+used by the Scikit Learn community.  As of version 0.10.0 it contains a plugin
+mechanism to allow Joblib code to use other parallel frameworks to execute
+computations.  The ``dask.distributed`` scheduler implements such a plugin in
+the ``distributed.joblib`` module and registers it appropriately with Joblib.
```
TomAugspurger: Maybe add, "when imported".

daskml/xgboost.py
mrocklin: I think that it is.  We don't import this in `daskml/__init__`

docs/source/joblib.rst
mrocklin: Resolved

PR #34, opened 2017-10-13
Rename to dask_ml

### LINKED ISSUE #33: Rename to dask_ml?
Most dask packages follow the following naming convention:

1.  dask-foo for package name on PyPI, and github
2.  dask_foo for import name in Python

The `dask-ml` package is currently imported as `daskml` rather than `dask_ml`.  The underscore is a bit of a pain, so I can see how the current name would be preferred.  Still though, I thought I'd bring this up.  We may want to switch for conformity's sake.

cc @TomAugspurger @jcrist

TomAugspurger:
Closes #33 

Might want to wait on #30, and then I'll rebase on that.

cc @mrocklin


TomAugspurger:
Also cc @daniel-severo. Sound OK to you?

daniel-severo:
I'm all for conventions.

LGTM

PR #54, opened 2017-10-20
CI: Test against py27

TomAugspurger:


.circleci/config.yml
```diff
@@ -18,22 +18,62 @@ jobs:
             conda update -q conda
             conda install conda-build anaconda-client --yes
             conda config --add channels conda-forge
-            conda env create -f ci/environment.yml --name=${ENV_NAME}
+            conda env create -f ci/environment.yml --name=${ENV_NAME} python=${PYTHON}
```
jakirkham: Might want to add `-q` to this and the other install commands. Otherwise there are many lines related to progress bars that changed slightly. 😜

TomAugspurger: Thanks :D

PR #60, opened 2017-10-24
Datasets wraps

### LINKED ISSUE #58: @wraps breaks introspection
https://github.com/dask/dask-ml/blob/cb3d0ea392a277cade35405f6b156f22a1616551/dask_ml/datasets.py#L11

The use of `@wraps` here leads to wrong behavior of some introspection tools, like `??` in Jupyter (will show you the code for the sklearn version).

I feel that it's best to wrap thing manually (assigning the `__name__`, `__doc__` and `__module__` attributes for `inner` manually), after line 17, before the `return`.

This issue is making some unpleasant workarounds necessary in the `xarray_filters` codebase (same `datasets.py` file) (where we are trying to give more helpful docstrings users beyond what's in sklearn). The issue for me revolves around knowing whether I'm going to wrap the `make_blobs` from `dask_ml` or from `sklearn`. I could do it by inspecting the signature, or the `__module__` attribute, but those attributes are set equal to the ones in sklearn right now.

I'll put links to relevants parts of the `xarray_filters` codebase here as soon as I push the new code up.

TomAugspurger:
Closes #58 

cc @gpfreitas

### DIFF
--- a/dask_ml/cluster/k_means.py
+++ b/dask_ml/cluster/k_means.py
@@ -139,7 +139,7 @@ class KMeans(BaseEstimator):
         if isinstance(X, dd.DataFrame):
             raise TypeError("Unable to fit K-Means on dataframe, since the "
                             "chunk lengths are not known")
-        if isinstance(X,  pd.DataFrame):
+        if isinstance(X, pd.DataFrame):
             X = X.values
 
         if isinstance(X, np.ndarray):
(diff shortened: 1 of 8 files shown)


TomAugspurger:
I'm not sure what the best way of rewriting function signatures is on py2, so that'll be py3 only I guess. Python2 users will see `*args, **kwargs`.

PR #62, opened 2017-10-24
[WIP] RobustScaler

jorisvandenbossche:
We are having a mini coding workshop with Matthew, and @ogrisel and I made a very minimal implementation of a RobustScaler. It's working on a toy dataset, but not yet more extensive testing.

### DIFF
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -210,3 +210,22 @@ class QuantileTransformer(skdata.QuantileTransformer):
             X_col = da.clip(X_col, clip_min, clip_max)
 
         return X_col
+
+
+class RobustScaler(skdata.RobustScaler):
+
+    def _check_array(self, X, copy):
+        return X
+
+    def fit(self, X, y=None):
+        q_min, q_max = self.quantile_range
+        if not 0 <= q_min <= q_max <= 100:
+            raise ValueError("Invalid quantile range: %s" %
+                             str(self.quantile_range))
+
+        quantiles = [da.percentile(col, [q_min, 50., q_max]) for col in X.T]
+        quantiles = da.vstack(quantiles).compute()
+        self.center_ = quantiles[:, 1]
+        self.scale_ = quantiles[:, 2] - quantiles[:, 0]
+        self.scale_ = skdata._handle_zeros_in_scale(self.scale_, copy=False)
+        return self


dask_ml/preprocessing/data.py
```diff
@@ -210,3 +210,22 @@ def _transform_col(self, X_col, quantiles, inverse):
             X_col = da.clip(X_col, clip_min, clip_max)
 
         return X_col
+
+
+class RobustScaler(skdata.RobustScaler):
+
+    def _check_array(self, X, copy):
+        return X
```
TomAugspurger: FYI: adding a `utils.validation.check_array` that works on dask arrays is near the top of my TODO list.

dask_ml/preprocessing/data.py
```diff
@@ -210,3 +210,22 @@ def _transform_col(self, X_col, quantiles, inverse):
...
+
+    def fit(self, X, y=None):
+        q_min, q_max = self.quantile_range
+        if not 0 <= q_min <= q_max <= 100:
+            raise ValueError("Invalid quantile range: %s" %
+                             str(self.quantile_range))
+
+        quantiles = [da.percentile(col, [q_min, 50., q_max]) for col in X.T]
```
ogrisel: It would be great to add the `axis` kwarg to `da.percentile` to reflect what is available in `np.percentile`.

BTW, I think the code `RobustScaler` in scikit-learn should also use `np.percentile` to compute both the median and the interquantile range at the same time as done here.

jorisvandenbossche:
One question that came up is how to handle computed attributes during fit. Here in the PR I `compute` them, but in the other examples already included in `preprocessing/data.py` the approach is not uniform: both `compute` as `persist` are used.

TomAugspurger:
I think for now we want to always compute the attributes (https://github.com/dask/dask-ml/pull/71/files#diff-b3693cf8f80e35f1ef5ec62caeebce81R64). In the future, we may add a `compute` keyword to each estimator.

The fitted attributes here are small correct?

ogrisel:
Yes, 2 float arrays of size `n_features`.

jorisvandenbossche:
Yes, but for the StandardScaler that is the same (similarly small data), but there `persist` is used (but that could be changed to compute for consistency if we choose that option)

TomAugspurger:
Yes, we should change StandardScaler (and MinMaxScaler) to be concrete. Apologies for the confusion.

tests/preprocessing/test_data.py
```diff
@@ -104,6 +104,61 @@ def test_df_column_slice(self):
...
 
 
+class TestRobustScaler(object):
+    def test_fit(self):
+        a = dpp.RobustScaler()
+        b = spp.RobustScaler()
+
+        # bigger data to make percentile more reliable
+        # and not centered around 0 to make rtol work
+        X, y = make_classification(n_samples=1000, chunks=200)
```
ogrisel: Maybe fix the seed: `random_state=0`

tests/preprocessing/test_data.py
```diff
@@ -104,6 +104,61 @@ def test_df_column_slice(self):
...
+
+        # overwriting dask-ml's fitted attributes to have them exactly equal
+        # (the approximate equality is tested above)
+        a.scale_ = b.scale_
+        a.center_ = b.center_
+
+        assert_eq_ar(a.transform(X).compute(), b.transform(X.compute()))
+
+    def test_inverse_transform(self):
+        a = dpp.RobustScaler()
```
ogrisel: Better rename `a` to `scaler` or `rs`.

PR #63, opened 2017-10-25
Add dask_ml.joblib import

mrocklin:


docs/source/joblib.rst
```diff
@@ -9,14 +9,14 @@ Joblib_ is a library for simple parallel programming primarily developed and
...
-the ``distributed.joblib`` module and registers it appropriately with Joblib
+the ``dask_ml.joblib`` module and registers it appropriately with Joblib
 when imported.  As a result, any joblib code (including many scikit-learn
 algorithms) will run on the distributed scheduler if you enclose it in a
 context manager as follows:
 
 .. code-block:: python
 
-   import distributed.joblib
+   import dask_ml.joblib  # import registers joblib plugin
```
TomAugspurger: Could we change the comment to `# joblib plugin registered on import`. Seeing `import ...` threw me off for a second since it looked like the start of another module import statement.

mrocklin: Changed it to `registers joblib plugin`

PR #64, opened 2017-10-25
Refer to dask_ml over dask_glm in docs

mrocklin:
Also, I found that this example failed:

```python
   >>> from dask_ml.linear_model import LogisticRegression
   >>> from dask_ml.datasets import make_classification
   >>> X, y = make_classification()
   >>> lr = LogisticRegression()
   >>> lr.fit(X, y)
```

Chunks wasn't available in the `_wrap_maker` function, presumably because I didn't provide it as a keyword to `make_classification`.  Is this a requirement?  If so, how should I adjust the example?

### DIFF
--- a/dask_ml/datasets.py
+++ b/dask_ml/datasets.py
@@ -70,7 +70,7 @@ def make_counts(n_samples=1000, n_features=100, n_informative=2, scale=1.0,
 
     Examples
     --------
-    >>> X, y = make_classification()
+    >>> X, y = make_counts()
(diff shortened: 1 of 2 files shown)


TomAugspurger:
> Chunks wasn't available in the _wrap_maker function, presumably because I didn't provide it as a keyword to make_classification.

Yep, it's currently required, though we could pick a good default based on the number of CPU cores. I can make a PR doing that later today.

PR #71, opened 2017-10-26
DOC: Add contributing docs

TomAugspurger:


docs/source/contributing.rst
```diff
@@ -0,0 +1,80 @@
...
+
+And then
+
+.. code-block:: none
+
+   python -m pip install -e .
+
+If you have any trouble with the build step, please open an issue.
+
+Conventions
```
TomAugspurger: I'd be curious to hear other's thoughts on here.

At some point, if we add a `compute` keyword to each estimator, we'll add that here.

daniel-severo:
To clarify: the attributes learned in `fit` should be persisted, not computed, right? I think this is the way the scalers have been implemented. Or are we now changing it?

PR #72, opened 2017-10-26
Various k-means fixes

TomAugspurger:
Comes from trying to satisfy check_estimator. But that seems a bit
too restrictive.

### DIFF
--- a/ci/environment-27.yml
+++ b/ci/environment-27.yml
@@ -9,6 +9,7 @@ dependencies:
   - heapdict
   - multipledispatch
   - nbsphinx
+  - nose
   - numpy
   - numpydoc
   - pandas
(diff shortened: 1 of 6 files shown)


TomAugspurger:
Quick update on the changes here:

- [First commit](https://github.com/dask/dask-ml/pull/72/commits/cca305086cbb776a390be9366be508ee57e6c40e): ensure our dtypes match scikit-learn's throughout
- [Second commit](https://github.com/dask/dask-ml/pull/72/commits/6c457c7709bc6bcccc7803cad55f83e2389d3346): Implements a `check_array`. On a `(10_000_000, 2)` array, that adds ~0.03s of time to the fit time (47s) which I think is more than acceptable.

PR #74, opened 2017-10-27
ENH: Add DummyEncoder

TomAugspurger:


dask_ml/preprocessing/data.py
```diff
@@ -370,3 +370,231 @@ def transform(self, X, y=None):
...
+            Dask array or dataframe will return a Dask DataFrame.
+            Numpy array or pandas dataframe will return a pandas DataFrame
+        """
+        if isinstance(X, np.ndarray):
+            X = pd.DataFrame(X, columns=self.transformed_columns_)
+
+        elif isinstance(X, da.Array):
+            # later on we concat(..., axis=1), which requires
+            # known divisions. Suboptimal, but I think unavoidable.
+            unknown = np.isnan(X.chunks[0]).any()
```
TomAugspurger: @mrocklin, I'd be curious to hear your thoughts on how we can better handle unknown chunks/divisions here. The stuff here and further down (directly assigning `._chunks`) is pretty hacky.

In this case, since we're going `da.Array` with unknown chunks -> `dd.DataFrame`, we known we'll end up with a RangeIndex, so as long as we know the length of each chunk we can compute the divsions.

I *think* that having `dd.from_dask_array` took an optional `divisions` argument (or maybe `chunks`?) would be the cleanest way to support this. We could actually require that the argument be provided when there are any unknown divisions, since currently you end up with incorrect results when the dask array as unknown chunks.

mrocklin: Do you need to know the number of rows in every chunk?  Why do you need accurate divisions here?

I agree that modifying `_chunks` feels odd.  We don't yet have a better mechanism.

PR #78, opened 2017-10-30
[WIP]TruncatedSVD/PCA

### LINKED ISSUE #22: PCA, TruncatedSVD Transformers
This would both be useful / not too difficult (hopefully). Can build off the SVD that's in dask.array.

TomAugspurger:
Closes #22

### DIFF
--- a/dask_ml/decomposition/__init__.py
+++ b/dask_ml/decomposition/__init__.py
@@ -1,0 +1,2 @@
+from .pca import PCA                     # noqa
+from .truncated_svd import TruncatedSVD  # noqa
--- a/dask_ml/decomposition/pca.py
+++ b/dask_ml/decomposition/pca.py
@@ -1,0 +1,434 @@
+import dask.array as da
+import numpy as np
+import scipy.sparse as sp
+from dask import compute
+from sklearn.decomposition.base import _BasePCA
+from sklearn.utils.extmath import fast_logdet
+from sklearn.utils.validation import check_is_fitted, check_random_state
+
+from ..utils import svd_flip
+
+
+class PCA(_BasePCA):
+    """Principal component analysis (PCA)
+
+    Linear dimensionality reduction using Singular Value Decomposition of the
+    data to project it to a lower dimensional space.
+
+    It uses the "tsqr" algorithm from Benson et. al. (2013). See the References
+    for more.
+
+    Read more in the :ref:`User Guide <PCA>`.
+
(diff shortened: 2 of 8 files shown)


dask_ml/decomposition.py
```diff
@@ -0,0 +1,362 @@
...
+
+        # compute
+        (U, S, V,
+         explained_variance_, total_var,
+         explained_variance_ratio_,
+         ) = compute(U, S, V,
+                     explained_variance_, total_var,
+                     explained_variance_ratio_)
+        components_ = V
+        singular_values_ = S.copy()
```
mrocklin: Is the compute call above safe?  Are the U and V parts of that computation likely to be large?

Do we ever use the computed `U` matrix explicitly?

TomAugspurger: - `components_` (`V`)  is `(min(n_components, n_features_), n_features)`
- `S` and `U` are both `(min(n_components, n_features)`,)

So as long as the array is not too wide, this will be OK.  I've mainly been assuming this up till now, though decomposition is probably the area we're mostly likely to run into extremely wide arrays.

TomAugspurger:
Gist showing some comparisons on PCA: https://gist.github.com/dced18b10aa28cba434771a37a3576f7

Dask is slower on small datasets (1,000 x 500), but quite a bit faster on multi-GB datasets. For 1 100,000 x 5,000 array of doubles (4GB), dask took 11s vs. 35s for scikit-learn, using the 'randomized' solvers (svd_compressed for dask, Halko et. al for scikit-learn).

Merging.

mrocklin:
Nice results.  Do the two methods have the same accuracy?

cc @ogrisel @marianotepper

TomAugspurger:
> Do the two methods have the same accuracy?

Scikit-learn's solvers seem to have a higher degree of precision. In most of the tests, I needed to adjust the tolerence from something like `1e-5` to `1e-3`.

I wish I had kept better notes on the patches I made to the scikit-learn test suite as I was going through. I have another branch that redid everything in a more careful way, but it was taking a while to re-do everything. https://github.com/TomAugspurger/dask-ml/commit/07bfcb2e5ed82a9e9f63bf840d6025bb176f71af shows what the required differences for accuracy were like. The other commits in that branch do things like remove arpacak-specific solvers, remove tests where `n_samples > n_features`, which the tsqr algorithm doesn't like.

PR #82, opened 2017-11-02
ENH: Nystrom kernel_approximation

TomAugspurger:
This is a building block for several methods, most notably spectral clustering,
SVM, and Gaussian Processes.

This implements the Nystrom approximation that makes it possible to scale those
out to larger datasets.


TomAugspurger:
Quick status update on this: the scikit-learn kernel approximation (which this PR follows) is to reduce the dimensionality of X, so it maybe be useful for very wide inputs, but still suffers from the poor scalability in the number of samples. I'm going to leave this as a WIP for now, to see if there's interest.

PR #91, opened 2017-11-14
[WIP]: Spectral Clustering

TomAugspurger:
Lots of work to do yet on the API, docs, tests, but I think I have the basic algorithm down.

This is following https://www.csie.ntu.edu.tw/~cjlin/papers/psc08.pdf most closely, though I'll add other references in the docs.

http://nbviewer.jupyter.org/gist/TomAugspurger/bd49e706dadd21c70ad34fdaff5bfc25

Here's the main takeaway

download

That shows the running time in seconds of scikit-learn's implementation (which computes the full `n_samples x n_samples` affinity matrix), compared to this PR's implementation which uses an approximation (based on 100 rows in that example, but that's tunable).

Which looks good from a performance standpoint, assuming the correctness is there :)

cc @PeterDSteinberg


TomAugspurger:
I missed an optimization to avoid a large matrix multiply earlier. Here's the updated chart:

scale

TomAugspurger:
@PeterDSteinberg if you have some time, could you point me to a good dataset to test this out on?

Here's some image segmentation examples on images builtin to skimage:

http://nbviewer.jupyter.org/gist/TomAugspurger/a426545268af6beaa32b926fd75deeee

PeterDSteinberg:
Hi @TomAugspurger A high value data set for you to try out here is the one used in this Phase I notebook:

 * https://github.com/ContinuumIO/elm/blob/master/examples/temperature-PDFs-clustering.ipynb

It is, if I recall correctly, the same MERRA2 surface temperature data set used by:
```
Classifying reanalysis surface temperature probability density functions (PDFs) over North America with cluster analysis
P. C. Loikith, B. R. Lintner, J. Kim, H. Lee, J. D. Neelin, and D. E. Waliser

GEOPHYSICAL RESEARCH LETTERS, VOL. 40, 3710–3714, doi:10.1002/grl.50688, 2013
```
(I'll send you that paper privately for copyright reasons).  

It's a bit of set up because it is 1000+ NetCDFs that must be downloaded, but would make a good demo for promotion of these tools.  The idea of the paper is taking 1000+ surface temperature 2D rasters (one NetCDF each), concatenating them in time to 3D, histogram the temperature deviations along the time dimension into 150 bins, and use the log counts within those histogram bins as input features to clustering.  They're using the PDF of each raster cell's temperature time series as a way of classifying climates.

Easier ideas:
   * Earthio is reasonably stable now and you can use its [S3 landsat utilities here](https://github.com/ContinuumIO/earthio/blob/master/earthio/s3_landsat_util.py) .  Those were written quickly for a conference - feel free to see if the downloading can be made more efficient or we should be importing an existing Anaconda/3rd party tool for S3 LANDSAT.  
   * One swath/row/time point of landsat is hardcoded into the datashader examples - https://github.com/bokeh/datashader/blob/master/examples/datasets.yml#L36

PeterDSteinberg:
Also @TomAugspurger I can do an example with dask-ml spectral clustering related to NLDAS data.  Just need about 1 to 2 weeks to clean up this PR: https://github.com/ContinuumIO/Elm-Earthio-NLDAS/tree/post-elm-pr-192

TomAugspurger:
Thanks, I'll give those a shot later.

TomAugspurger:
Thanks, I had some issues downloading the data, probably user-error. I'll try again later.

I've had some issues when the data aren't scaled. For example, this current fails:

```python
from dask_ml.cluster import SpectralClustering
from scipy.misc import face

img = face(gray=True)[:600, 400:1000]
sc1 = SpectralClustering(n_clusters=8)
X1 = img.reshape(-1, 1)
sc1.fit(X1)
```

Normalizing the data fixes things

```python
X2 = (img.reshape(-1, 1) - img.mean()) / img.std()
s1.fit(X2)
```

I haven't discovered if this is a numerical accuracy thing or whether it actually matters for the math to work out correctly.

I'm going to merge this once I get the CI passing.

PeterDSteinberg:
If the data are not normally distributed, e.g. strong positive skew, then you would be more likely to encounter matrix algebra edge cases related to condition number.  The zscore normalization you mentioned would reduce the residuals and condition number.  I would guess the problem is related to the specifics of the input data (`misc` or `gray=True`?).

TomAugspurger:
Apparently the issue I mentioned in https://github.com/dask/dask-ml/pull/91#issuecomment-345832180 was because the data were integers. https://github.com/dask/dask-ml/pull/91/commits/bafd40e3d87a9f685b9e4c5e661107e5c338afde fixed things by converting the array to floats.

TomAugspurger:
Will do a bit more iteration as follow-up PRs.

PR #94, opened 2017-11-15
Implement GLM estimators in dask-ml

TomAugspurger:
Moving them here from dask-glm.

No changes to the API yet.

### DIFF
--- a/dask_ml/datasets.py
+++ b/dask_ml/datasets.py
@@ -82,48 +82,6 @@ def make_counts(n_samples=1000, n_features=100, n_informative=2, scale=1.0,
     return X, y
 
 
-def make_poisson(n_samples=1000, n_features=100, n_informative=2, scale=1.0,
-                 chunks=100):
-    """
-    Generate a dummy dataset for modeling count data.
-
-    Parameters
-    ----------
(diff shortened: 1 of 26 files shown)


PeterDSteinberg:
Thanks @TomAugspurger  - you beat me to it - Is the goal for this PR to make `dask-glm` installation no longer required for `dask-ml` - essentially moving all of its assets to the new subpackage here?  That would be my vote but it's not critical to me.

TomAugspurger:
dask-glm will still be a required dependency, as the optimizers and regularizers will still live there. Just the scikit-learn style estimators are moving here.

dask_ml/datasets.py
```diff
@@ -82,6 +82,48 @@ def make_counts(n_samples=1000, n_features=100, n_informative=2, scale=1.0,
     return X, y
 
 
+def make_poisson(n_samples=1000, n_features=100, n_informative=2, scale=1.0,
```
MLnick: This seems to be exactly the same as `make_counts` above?

TomAugspurger: Whoops, thanks.

Issue #65, opened 2017-10-25
Make chunks optional in the datasets.make_* methods


TomAugspurger:
Just pick a good default.

PR #105, opened 2017-11-18
made chunks optional in datasets.make_* methods

deshpand:
default value: 100

### DIFF
--- a/dask_ml/datasets.py
+++ b/dask_ml/datasets.py
@@ -19,7 +19,9 @@ chunks_doc = """
 def _wrap_maker(func):
 
     def inner(*args, **kwargs):
-        chunks = kwargs.pop('chunks')
+        chunks = 100
+        if 'chunks' in kwargs:
+            chunks = kwargs.pop('chunks')
         X, y = func(*args, **kwargs)
         return (da.from_array(X, chunks=(chunks, X.shape[-1])),
                 da.from_array(y, chunks=chunks))


mrocklin:
Default values can be dangerous.  They need to be robust in all situations.  chunksizes of 100 might not be appropriate in some situations.  My guess is that if we want to support a default chunksize here that we would want to be more clever than just a single value.  Historically we have avoided trying to make these decisions because they can be quite hard generally.

deshpand:
Thank you.  I guess I was being over-eager to get my first commit!  This was for issue #65 which also had the label 'good first issue'.  If you or Tom have suggestions for a 'clever than just a single value', please let me know.  Or I can look at other issues and/or familiarize myself with the project.. thanks

TomAugspurger:
Sorry for mislabeling that issue, I hadn't thought through things fully :)

Another option is to inspect the number of available cpu cores, if we go down this route. Personally though, I don't mind having to think about the number of chunks.

deshpand:
No problem @TomAugspurger 
< I don't mind having to think about the number of chunks.
I can close/cancel this PR for now, say after another day.  Please let me know if there is a mailing list or other way of communication.  Keen on getting involved, hoping for a jump start with holidays/vacations coming up..

TomAugspurger:
Github issues probably work best for now.

Are there any areas in particular you're interested in working on?

On Sun, Nov 19, 2017 at 7:51 AM, Gururaj Deshpande <notifications@github.com

deshpand:
Sure, I will look at the issues, documentation and peruse the code.  Nothing in particular, just looking to contribute.  Relatively new to scikit-learn/ML and would like to keep learning, been coding in Python for a few years.  Also dabbled with spark, keras, Theano etc. recently.  Some Pycon/*con videos led me to dask and now dask-ml.

PR #111, opened 2017-12-28
Updated notes on scalability of various estimators

### LINKED ISSUE #109: How does dask-ml handle large datasets?
From the latest demos of dask-sklearn (and related projects) that I've seen, dask has primarily focused on parallelizing grid/random search and CV, and provides some smart caching mechanisms for transformers and estimators that have already been evaluated on a dataset.

However, that approach breaks down when, say, a slow algorithm is being trained on a very large dataset. With some slower algorithms, training one instance of the algorithm on one fold of the dataset can take hours or days, which makes training the algorithm unviable for all but the most patient practitioner.

Can dask-ml speed up the underlying ML algorithms that it supports?

TomAugspurger:
Closes #109

### DIFF
--- a/docs/source/modules/api.rst
+++ b/docs/source/modules/api.rst
@@ -4,6 +4,10 @@
 API Reference
 =============
 
+This page lists all of the estimators and top-level functions in ``dask_ml``.
+Unless otherwise noted, the estimators implemented in ``dask-ml`` are
+appropriate for parallel and distributed training.
+
 :mod:`dask_ml.model_selection`: Model Selection
 ===============================================
 
@@ -42,6 +46,12 @@ Incremental Learning
 ====================
 
 .. currentmodule:: dask_ml
+
+Some scikit-learn estimators support out-of-core training through the
+``partial_fit`` method. The following estimators wrap those scikit-learn
+estimators, allowing them to be used in Pipelines and on Dask arrays and
+dataframes. Training will still be serial, so these will not benefit from
+a paralell or distributed training any more than the underlying estimator.
 
 .. autosummary::
    :toctree: generated/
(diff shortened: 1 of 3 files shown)


TomAugspurger:
I'm not sure if this is the best approach. The short version is that

1. All the estimators in dask-ml will work in parallel on distributed arrays.
2. The `Incremental*` wrappers will work in serial over the blocks, so they won't benefit from any parallelism, beyond what scikit-learn implements
3. The meta-estimators like `RandomizedSearchCV` inherit whatever parallelism the actual estimator implements. If you search over `dask_ml.linear_model.LogisticRegression`, things should be fine on a large array. If you search over `sklearn.linear_model.LogisticRegression` not so much.

We should probably have something like that in `index.rst` but I don't want to over-complicate things too early.

Adding more scalability examples as suggested by @marcelsan would be great (though we'll want to balance this against the time it takes to do a full doc build).

cc @rhiver

rhiever:
Thank you for the clarification and I think this is a good addition to the docs. Are there plans to implement more estimators in dask-ml in the future?

docs/source/modules/api.rst
```diff
@@ -43,6 +47,12 @@ Incremental Learning
 
 .. currentmodule:: dask_ml
 
+Some scikit-learn estimators support out-of-core training through the
+``partial_fit`` method. The following estimators wrap those scikit-learn
+estimators, allowing them to be used in Pipelines and on Dask arrays and
+dataframes. Training will still be serial, so these will not benefit from
+a paralell or distributed training any more than the underlying estimator.
```
rhiever: typo: paralell --> parallel

TomAugspurger:
> Are there plans to implement more estimators in dask-ml in the future?

Yep, as driven by user / developer interest.

Issue #5, opened 2017-09-24
Distributed TFIDF


daniel-severo:
Greetings!

I recently used dask to implement a distributed version of tfidf. I want to contribute to the dask project by putting it somewhere.

Would this be the correct repo.?

I thought maybe a `feature_extraction` directory would be appropriate.

TomAugspurger:
Things are still a bit in flux, but I think this would be a good home. This would be a great addition!

> I thought maybe a feature_extraction directory would be appropriate.

Perfect. I'm trying to follow scikit-learn's layout as closely as possible.

daniel-severo:
I see.

What's the difference between this repo. and `dask-glm`? Can't the later be a subset of the former?

TomAugspurger:
> What's the difference between this repo. and dask-glm?

I'm probably going to just import the dask-glm estimators into `dask-ml` namespace (likewise with dask-searchcv, dask-patternsearch). For the user, it'd be nice to have a single place to go for all dask-related ML things.

Development will probably still continue in those other repositories.

bendruitt:
I'd be interested in seeing this code. How was this achieved?

daniel-severo:
The way I first implemented it was something like this: (I changed it around so that I removed company private code and this was way before I started contributing to dask so my knowledge was far less than it is today)

```python
import numpy as np
import dask.bag as db

from toolz.curried import frequencies
from toolz.curried import valmap
from toolz.curried import merge_with
from toolz.curried import unique
from toolz import merge_with as _merge_with


def normalize(_dict):
    return {k:v/sum(_dict.values()) for k,v in _dict.items()}

corpus = 2*[
    "a a b b b",
    "a a a a c",
    "c b a d e f g",
    "a b b b c c c d d d a b",
    "c b a d e f g",
    "a b b b c c c d d d a b"
]


base = (
    db
    .from_sequence(corpus, npartitions=4)
    .map(str.split)
)

tf = (
    base
    .map(frequencies)
    .map(normalize)
)

idf = (
    base
    .map(unique)
    .map(frequencies)
    .reduction(merge_with(sum), merge_with(sum), split_every=2)
    .apply(normalize)
    .apply(valmap(np.log10))
    .apply(valmap(np.negative))
)
```

This has a draw-back, since it must pass over the entire dataset once in order to calculate the `idf` step.

[...]

PR #122, opened 2018-01-24
Add HashingVectorizer

rth:
Following the discussion in https://github.com/dask/dask-ml/issues/5 this PR is an attempt to wrap scikit-learn's `HashingVectorizer` for dask_ml.

This is a simpler problem than TF-IDF because scikit-learn's  `HashingVectorizer` is stateless and so it's an embarrassingly parallel problem.  The IDF weighting could be later added as an additional processing step.

Currently, I have two issues,
 * the output is a Bag of sparse scipy CSR arrays (one array per partition with multiprocessing) and I'm not sure how to best convert this to a sparse array that dask would understand.
 * If I understand correctly dask Bag is unordered. Does it mean that when this estimator gets a Bag containing a sequence of filenames, there is no guarantee that the rows of the vectorized document term matrix will match the original position in the sequence? How can we know which row corresponds to which filename then? 

**Todo**

 - [ ] Return dask sparse arrays
 - [ ] Constrain input (e.g. dask array of objects)
 - [ ] Determine which tests are necessary (and if any can/should be taken from scikit-learn)
 - [ ] Adapt the docstrings
 - [ ] Benchmark


mrocklin:
Thanks for your work here!  One major change is that I think all dask-ml methods are expected to consume and produce dask arrays.  Doing this uniformly throughout the project allows operations to be dropped in in various places in pipelines, etc., similar to how numpy arrays are used uniformly in sklearn.

For your second question on sparse arrays you might want to read through http://dask.pydata.org/en/latest/array-sparse.html

dask_ml/feature_extraction/text.py
```diff
@@ -0,0 +1,146 @@
...
+        self.norm = norm
+        self.alternate_sign = alternate_sign
+        self.dtype = dtype
+
+    def partial_fit(self, X, y=None):
+        """Does nothing: this transformer is stateless.
+        This method is just there to mark the fact that this transformer
+        can work in a streaming setup.
+        """
+        return self
```
jorisvandenbossche: Is there a reason to override `partial_fit` and `fit` ? It seems we can just use the parent ones?

TomAugspurger:
In this case, bags might be the right input type since the scikit-learn version accepts a list of strings.

As for the return value of `transform`, I think think returning a single sparse matrix from https://github.com/mrocklin/sparse makes sense, rather than a bag of sparse matrices.

I was working on adapting http://scikit-learn.org/stable/auto_examples/applications/plot_out_of_core_classification.html#sphx-glr-auto-examples-applications-plot-out-of-core-classification-py. Will post a gist later, but something like

```python
documents = stream_reuters_documents()

# convert to jsonl format
file_number = 0
for i, doc in enumerate(documents, 1):
    with open("documents-{:0>4}.jsonl".format(file_number), "a") as f:
        f.write(json.dumps(doc) + "\n")
    if i % 10 == 0:
        file_number += 1

# replacement for get_minibatch that
# 1. takes a bag
# 2. returns just the X_text (haven't figured out y yet)

def get_minibatch2(documents, pos_class=positive_class):
    """Extract a minibatch of examples, return a tuple X_text, y.

    Note: size is before excluding invalid docs with no topics assigned.

    """
    data = [(u'{title}\n\n{body}'.format(**doc), pos_class in doc['topics'])
            for doc in documents if doc['topics']]
    if not len(data):
        return np.asarray([], dtype=int), np.asarray([], dtype=int)
    X_text, y = zip(*data)
    return X_text  # TODO: y
```

And then using it:

```python
import dask_ml.feature_extraction.text
import sklearn.feature_extraction.text
import sparse

b = (db.read_text("*.jsonl")
    .map(json.loads)
    .filter(lambda x: len(x) > 0)
    .map_partitions(get_minibatch2))

v1 = sklearn.feature_extraction.text.HashingVectorizer(
    decode_error='ignore',
    n_features=2**18,
    alternate_sign=False
)

v2 = dask_ml.feature_extraction.text.HashingVectorizer(
    decode_error='ignore',
    n_features=2**18,
    alternate_sign=False
)

r1 = v1.transform(b.take(2))

r2 = v2.transform(b).take(2)
```

```python
>>> r1
<2x262144 sparse matrix of type '<class 'numpy.float64'>'
	with 213 stored elements in Compressed Sparse Row format>

>>> sparse.concatenate(r2)
<COO: shape=(2, 262144), dtype=float64, nnz=213, sorted=False, duplicates=True>
```

mrocklin:
> I think think returning a single sparse matrix from https://github.com/mrocklin/sparse makes sense

Do you mean a dask array of sparse matrices?

rth:
Thank you for the review! 

@mrocklin Thanks, I'll work on adapting the interface. It's true that text vectorizers are one of the few places in scikit-learn that take as input an iterable instead of an array (done so one could stream documents in).

@jorisvandenbossche You are right, will fix it.

@TomAugspurger Interesting, thanks for the ideas. I also went for that scikit-learn Reuters example in this PR after looking at the 20newsgoups and the RCV1 one. Reuters also appears to be the only text dataset that scikit-learn doesn't get pickled after downloading.  I tried to simplify it a bit as IMO the whole `ReutersParser` class and document filtering and parsing is overly complicated (and not very relevant) for an introductory example in dask-ml.  

I wonder if just using some dataset with one file per document would be easier to illustrate things with. For instance, there are a few subsets of the Trec Legal 2009 (Enron ERDM) dataset [here](https://github.com/FreeDiscovery/FreeDiscovery/blob/master/freediscovery/datasets.py#L92) that naturally exist in that format. The nice thing about it is that in later benchmarks one could use the same code with a larger subset of the dataset including the full 0.7M document collection (~7GB uncompressed) which would be more representative of real use cases. It's also used in the [dask word count example](http://dask.pydata.org/en/latest/examples/bag-word-count-hdfs.html). The issue is that one would still need to write a rudimentary fetcher to run the example in CI.. Not sure what would be the best approach.

TomAugspurger:
>   I tried to simplify it a bit as IMO the whole ReutersParser class and document filtering and parsing is overly complicated (and not very relevant) for an introductory example in dask-ml.

Agreed, I saw your test example later, it's much simpler :)

The Enron dataset would be a good test case for these. For testing, I'm happy with whatever is easiest, even just generating random documents on the fly. We have scikit-learn to compare to for the correct output, so we shouldn't need to hardcode any answers from a real dataset.

mrocklin:

rth:
Sorry for the delay @mrocklin, and thanks for the ping. I have blocked time to finish this PR this week.

mrocklin:
Checking in again here.  It looks like this has stalled.  What is the right way to move forward?

dask_ml/feature_extraction/text.py
```diff
@@ -0,0 +1,139 @@
...
+        Returns
+        -------
+        X : scipy.sparse matrix, shape = (n_samples, self.n_features)
+            Document-term matrix.
+        """
+
+        transformer = super(HashingVectorizer, self).transform
+
+        if isinstance(X, Bag):
+            return X.map_partitions(transformer)
```
mrocklin: It looks like this could also be a Dask Series array or anything that implements a `map_partitions` method that applies that function over iterables.  Perhaps something like the following would work instead?

```python
if hasattr(X, 'map_partitions):
    return X.map_partitions(transformer)
```

Also, what is the desired return type here?  Do we want this to be a Dask array of sparse arrays?  My guess is that dask.bag wouldn't give that to us by default

docs/examples/plot_text_vectorization.py
```diff
@@ -0,0 +1,84 @@
...
+from sklearn.datasets import get_data_home
+from sklearn.externals.six.moves.urllib.request import urlretrieve
+
+import dask
+import dask.bag as db
+import dask.multiprocessing
+
+from dask_ml.feature_extraction.text import HashingVectorizer
+
+dask.set_options(get=dask.multiprocessing.get)
```
mrocklin: We'll want to do this on a test-by-test basis rather than globally, otherwise we'll affect other tests

mrocklin: Ah, this is an example.  My mistake.

mrocklin:
I wonder if a broader solution to this problem would be to make a `Stateless` meta-transformer and then compose that with scikit-learn's `HashingVectorizer`

```python
from sklearn... import HashingVectorizer

transformer = HashingVectorizer(...)

from dask_ml... import StatelessTransformer

transformer = StatelessTransformer(transformer)

b = dask.bag.read_text('*.txt')
x = tranformer.transform(b)
```

dask_ml/feature_extraction/text.py
rth: Yes, we want this to be a Dask sparse array. How to convert from dask.bag to dask sparse array is one point I have not figured out yet.

rth:
Yes, sorry this PR has gone stale. I'm still interested in working on this, was just not able to find time for it so far, should be able to at scipy next week. If despite everything it's not the case, a contributor should feel free to take this PR over.

> I wonder if a broader solution to this problem would be to make a Stateless meta-transformer and then compose that with scikit-learn's HashingVectorizer

It depends if you think this will be useful elsewhere. As far as I can tell, other scikit-learn methods that could benefit form such meta-transformer are `sklearn.preprocessing.{FunctionTransformer,Normalizer,Binarizer}` etc.  Since in stateless transformers a `fit` doesn't do anything, wouldn't such stateless meta-transformer be somewhat equivalent to a `dask_ml.wrappers.ParallelPostFit` ?

rth:
Well except that it would need to work with dask.bag which is a particular case of HashingVectorizer. Because this particularity a dedicated estimator might still be more suitable..

mrocklin:
OK, it sounds like generalizing this doesn't make sense then.  Thanks for the explanation!

mrocklin:
Here is an example converting from a dask bag to a dask array of sparse arrays using dask.delayed:

```python
import dask.bag as db
import sparse
b = db.from_sequence(range(10), npartitions=3)

def f(partition):
    n = sum(partition)
    return sparse.random((n, 5), density=0.2)

f([0, 1, 2, 3])
# <COO: shape=(6, 5), dtype=float64, nnz=6>

delayed_values = b.map_partitions(f).to_delayed()
delayed_values[0].compute()
# <COO: shape=(6, 5), dtype=float64, nnz=6>

import numpy as np
import dask.array as da
arrays = [da.from_delayed(dv, shape=(np.nan, 5), dtype=float) for dv in delayed_values]
x = da.concatenate(arrays)
x
# dask.array<concatenate, shape=(nan, 5), dtype=float64, chunksize=(nan, 5)>

x.compute()
# <COO: shape=(45, 5), dtype=float64, nnz=45>
```

Note that we use `np.nan` in a shape when we don't know the number of elements along an axis of an array.

rth:
> Here is an example converting from a dask bag to a dask array of sparse arrays using dask.delayed:

Got it. Thanks a lot for this example, it's very helpful!

dask_ml/feature_extraction/text.py
TomAugspurger: Sorry for the delay in answering @rth!

I think the best way forward is something like

```python
bag2 = X.map_partitions(transformer)

objs = bag2.to_delayed()
arrs = [da.from_delayed(obj, (np.nan, self.n_features), self.dtype)
        for obj in objs]
result = da.concatenate(arrs, axis=0)
result
# dask.array<concatenate, shape=(nan, 1048576), dtype=float64, chunksize=(nan, 1048576)>
```

there are two potential issues with that approach

1. We don't know the length of each partition / block, so we end up with an array that has unknown chunks. This can cause issues with later estimators, but that's a larger issue than we should worry about for this PR.
2. `result.compute()` isn't actually valid :)

```pytb
<ipython-input-43-a4f886589eb2> in <module>()
----> 1 result.compute()
ValueError: setting an array element with a sequence.
```

That's because while you can store a `scipy.sparse` matrix in a dask.array, not everything works correctly. I have a WIP OneHotEncoder PR that runs into the same issue. I think for know, that's OK, it'll need to be solved in dask (or maybe by dask-ml by using `pydata/sparse`). But again, that's something we need to figure out library-wide.

docs/examples/plot_text_vectorization.py
```diff
@@ -0,0 +1,84 @@
...
+          .map(load_from_filename)
+          .flatten())
+
+vect = HashingVectorizer()
+
+X = vect.fit_transform(text)
+
+# there should be a a way to convert X to a sparse dask array
+# without moving & serializing the data from the respective processes
+X = scipy.sparse.vstack(X.compute())
```
TomAugspurger: This will hit the `X.compute()` issue. We can word around it with

```python
X.map_blocks(lambda block: block.toarray(), dtype=X.dtype).compute()
```

or we can use pydata/sparse to keep things sparse.

dask_ml/feature_extraction/text.py
mrocklin: I suspect that short term the solution for sparse arrays will be to place scipy.sparse matrices within dask arrays, but then avoiding using dask.array algorithms on those arrays.  Dask arrays should be fine as *containers* for sparse matrices, even if the algorithms don't work.

Note that this quasi-support is relatively new.  See https://github.com/dask/dask/pull/3738

TomAugspurger: > I suspect that short term the solution for sparse arrays will be to place scipy.sparse matrices within dask arrays

Agreed, short-term that's probably best (my phrasing above did not indicate that, my apologies).

TomAugspurger: I think in the example, it would be best to note this and show some workarounds

```python
import sparse

X.map_blocks(sparse.COO.from_scipy_sparse, dtype=X.dtype).compute()
```

@rth do you have time to make these changes? Sorry again for letting this linger.

dask_ml/feature_extraction/text.py
```diff
@@ -0,0 +1,146 @@
...
+
+        self._get_hasher().fit(None, y=None)
+        return self
+
+    def transform(self, X):
+        """Transform a sequence of documents to a document-term matrix.
+        Parameters
+        ----------
+        X : dask.Bag of raw text documents, length = n_samples
+            Samples. Each sample must be a text document (either bytes or
```
TomAugspurger: It seems like we only support `input='content'`, correct?

I suppose if `input='filename', then `X = db.read_text(X)`, should work.

TomAugspurger:
I think this will be ready to go.

One awkward part of our new examples workflow, we can't include examples relying on new features in the `dask-examples` repo, since they'll fail the tests. Even if we whitelisted them, we wouldn't want them to show up since they aren't runnable except on master.

So the proposal is to include new examples in Dask-ML, and then move them over to `dask-examples` as part of the release process.

docs/examples/plot_text_vectorization.py
mrocklin: See https://github.com/dask/dask/pull/3836

rth:
Thanks for continuing this @TomAugspurger , never got to it..

mrocklin:
Thanks for starting it @rth!

TomAugspurger:
Indeed, thanks @rth. I'd be very curious to hear how things work out if you get a chance to use this on a real-life dataset.

PR #125, opened 2018-01-26
Make scalers work on dask/pandas dataframes

jorisvandenbossche:


dask_ml/preprocessing/data.py
```diff
@@ -29,20 +29,32 @@ def fit(self, X, y=None):
         self._reset()
```
rth: Maybe also add input validation with check_array somewhere in fit (as it's done in `RobustScaler`)?

 Since currently with,
```py
df = pd.DataFrame([{'a': 3, 'b': 'one'}, {'a': 4, 'b': 'two'}])
```
`StandardScaler().fit(df)` doesn't produce any errors (unlike the one from scikit-learn) and `StandardScaler().fit_transform(df)` raises an exeption
```py
    [...]
    return self.fit(X, **fit_params).transform(X)
  File "/tmp/test.py", line 68, in transform
    X -= self.mean_
  [...]
ValueError: Unable to coerce to Series, length must be 2: given 1
``` 
that's not explicit about the fact that the dtype is unsupported...

jorisvandenbossche: Hmm, yes, and the pandas sum automatically drops the non-numeric column, so getting mismatched length. 
Adding some validation of numeric dtypes

mrocklin:

PR #127, opened 2018-01-30
[WIP]IID wrappers for scikit-learn

### LINKED ISSUE #126: Subpackage for easy out-of-core transformers
*If your data is IID*, it's quite easy to get all of scikit-learn.

```
def fit(self, X, y=None):
    result = super().fit(first_block_of_X, first_block_of_y)
    # copy over all the learned attributes of self

def transform(self, X, y=None):
    return X.map_partitions(super().transform)
```

With IID data, this will be a good approximation to fitting on the entire dataset, and then transforming.

We could implement this in a subpackage to make it clear that `dask_ml.iid.StandardScalar` is different from `dask_ml.StandardScalar`.

TomAugspurger:
Adds a new subpackage `dask_ml.iid` that wraps most estimators from scikit-learn so that

1. Training is done just on the first block / partition of a dask array or dataframe.
2. Transform and predict work correctly on dask arrays or dataframes

This covers the common case observed by @GaelVaroquaux that we're often OK with training on a sample.

TODO:

- Alternative name? I like `dask_ml.iid`, since it drives home the point that these wrappers aren't appropriate for all types of data. We'll need to do some education for users who aren't familiar with the term though.
- Alternative implementation? We could provide a function that rewrites a class after it's been created (similar to what `sklearn-xarray` does.) How does that affect pickling?
- Docs? Do we want to include every estimator in our API docs? What value does that bring, over saying 'just look at scikit-learn'?
- Other methods to implement? `score` probably, anything else?

Closes #126

### DIFF
--- a/dask_ml/base.py
+++ b/dask_ml/base.py
@@ -1,9 +1,14 @@
+import inspect
 import os
-import six
 from abc import ABCMeta
 
-import numpy as np
 import dask
+import dask.array as da
+import dask.dataframe as dd
+import dask.delayed
+import numpy as np
+import six
+import sklearn.base
 from dask.array import learn
 
 
@@ -77,6 +82,7 @@ class _BigPartialFitMixin(object):
 
 
 def _copy_partial_doc(cls):
+    """Insert a qualification into the base class' docstring."""
     for base in cls.mro():
         if base.__module__.startswith('sklearn'):
             break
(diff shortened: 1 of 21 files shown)


mrocklin:
> Training is done just on the first block / partition of a dask array or dataframe.

We might have some work to do to clearly convey this caveat to users.  My initial reaction is that this will be something of a surprise.

mrocklin:
Rather than the magic construction of many estimators we might also just make a class that lets people wrap up a scikit-learn estimator and publish that?  They might be comfortable doing a straightforward step like this and it might reduce the amount of magic that we provide/maintain.

TomAugspurger:
> Rather than the magic construction of many estimators we might also just make a class that lets people wrap up a scikit-learn estimator and publish that?

I read http://matthewrocklin.com/blog/work/2018/01/27/write-dumb-code immediately *after* writing this :)

I'll look into how that affects pickling / unpickling of estimator objects.

mrocklin:
Just random thoughts.  I think that you and @GaelVaroquaux have a better handle on what is best for this user community than I do.

TomAugspurger:
To be clear, I'm not tied to this implementation at all. The one thing I do like about it though is it's explicitly in a separate sub-package, which helps with

> We might have some work to do to clearly convey this caveat to users. My initial reaction is that this will be something of a surprise.

My hope is that the separate namespace and the warnings in the docs are sufficient.

We could also just define the classes normally and inherit from the mixin classes I've defined. We'll just need to update the files with each scikit-learn release which is not too onerous.

Another downside: we may now have three implementations of `PCA`

1. re-implemented from the ground up using dask arrays
2. `IncrementalPCA`, which wraps `sklearn.decomposition.PCA.partial_fit`
3 `dask_ml.iid.PCA`

which will potentially be confusing.

GaelVaroquaux:
> Rather than the magic construction of many estimators we might also
> just make a class that lets people wrap up a scikit-learn estimator and
> publish that?

Good idea

GaelVaroquaux:
Disclaimer: I did not have the time to look at the code of the PR.

In terms of design, in an iid package, we could have two functionality:

- Something for transformers that fits the transformer only on the first blocker. I'll call it the "FirstBlockFit" metaestimator for now.

- Something for supervised learning that does an ensemble on each block. I'll call it the "BlockEnsembler" for now.

I think that this functionality, together with a good documentation covers most the of cases.

To say more, I would need to read the PR/think more, but I don't have the time right now.

TomAugspurger:
I'm coming around on the idea of a wrapper / meta-estimator. That could take a scikit-learn estimator and a sampling strategy, like `sampling={'first', 'random'}`, names TBD. That will make it even clearer what's going on, and a bit more flexible.

> Something for supervised learning that does an ensemble on each block.

Yes, this is unimplemented so far. I'd like to see it as well. I'll try to pick it up next week.

TomAugspurger:
Closing in favor of #132 (though I think some of the testing machinery added here could be useful).

PR #140, opened 2018-02-20
DOC: Added PCA [ci skip]

TomAugspurger:


TomAugspurger:
The third commit is for compatibility with scikit-learn master. We failed to pass through the random state properly everywhere.

PR #144, opened 2018-02-27
PERF: Some scalability fixes

TomAugspurger:
Computing the upper-left block was taking a while since we didn't persist
the small block before doing pairwise distances.


TomAugspurger:
The failure is fixed on master, but I made this branch before that was merged, and Circle-CI doesn't test merge commits. It should pass on merge.

PR #156, opened 2018-04-03
Add style sheet

mrocklin:
This follows on https://github.com/dask/dask/pull/3360

### DIFF
--- a/docs/source/_static/style.css
+++ b/docs/source/_static/style.css
@@ -1,0 +1,71 @@
+@import url("theme.css");
+
+.wy-menu-vertical li ul li a,
+.wy-menu-vertical li.current a,
+.wy-menu-vertical a {
+  color: #ECB172;
+}
+
+.wy-menu-vertical li.toctree-l2.current>a:hover,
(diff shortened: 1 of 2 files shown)


mrocklin:
Alternatively we could also just link directly to Dask's style sheet

    app.add_stylesheet('https://dask.pydata.org/en/latest/_static/style.css')

jakirkham:
Do we need this at Dask too? Anywhere else it might be needed?

mrocklin:
This was copied over from Dask.  Did you mean somewhere else?

mrocklin:
> Do we need this at Dask too? Anywhere else it might be needed?

Presumably we might apply this to any of the various downstream libraries.  If we start doing this repeatedly (which might be wise), then I'll suggest that rather than copy around the CSS file as I've done here we instead link to a single version of Dask's stylesheet by placing this line in the source/conf.py file

```python
def setup(...):
    ...
    app.add_stylesheet('https://dask.pydata.org/en/latest/_static/style.css')
```

Issue #157, opened 2018-04-06
StandardScaler does not support a dataframe?


bnmnetp:
Given the following code:
```
df = pd.DataFrame({'A':[1,2,3,4,2,2,3,4], 'B':[4,4,4,3,3,3,7,7], 'C':[2,2,4,3,2,3,2,7]})
xd = dd.from_pandas(df, npartitions=2)
scl = StandardScaler()
scl.fit(xd)
```

I get the error
```
TypeError: 'Series' object does not support item assignment
```

Does the `dask_ml.preprocessing.StandardScaler` not support dataframes?  The documentation says it supports any of the Dask collections, so I assumed it would.  I also tried to convert it to a Dask array but that gives me a different error:  ` TypeError: 'float' object cannot be interpreted as an integer`

Maybe I'm just doing something wrong?

TomAugspurger:
Not currently, shouldn't be hard to support though. We should have `StandardScalar` define a `_check_array` that transforms dask.dataframes to dask.arrays I think (with `.values`).

bnmnetp:
@mrocklin I would love to help out, but my hacking time is very limited during the semester, and responding to issues on my own projects keeps me pretty busy.   I ran across this problem while trying to prepare an example for a class I'm teaching.  If its still lingering come summer I may be able to help out.

mrocklin:
Sounds good.  Thanks for the response.

cr458:
@mrocklin I could have a look at this if needed?

mrocklin:
Glad to hear it @cr458 !  This project would benefit from your help.

For future reference you don't need my permission to take on issues like this.  You're right that it's probably good to announce your intent to ensure that no one else has some unpublished work that might overlap, but in general anyone can work on anything.  Even if there was some more formal process I probably wouldn't be the one to ping, @TomAugspurger tends to lead development on dask-ml.

cr458:
@TomAugspurger, @mrocklin it seems the error described by @bnmnetp when passing a dask.array to the `StandardScaler.fit()` method arises from this line `attributes['n_samples_seen_'] = len(X)`.

Since the chunks are not defined for the dask.array, the array in his example has shape `(nan, 3)` and this raises the error `TypeError: 'float' object cannot be interpreted as an integer` due to the `np.nan` .

[...]

PR #163, opened 2018-04-20
Convert dask.dataframe to dask.array when using StandardScaler.fit()

cr458:
Closes to #157

### DIFF
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -28,7 +28,7 @@ class StandardScaler(skdata.StandardScaler):
     def fit(self, X, y=None):
         self._reset()
         attributes = OrderedDict()
-        if isinstance(X, dd.DataFrame):
+        if isinstance(X, (pd.DataFrame, dd.DataFrame)):
             X = X.values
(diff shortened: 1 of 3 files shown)


dask_ml/preprocessing/data.py
```diff
@@ -28,6 +28,8 @@ class StandardScaler(skdata.StandardScaler):
     def fit(self, X, y=None):
         self._reset()
         attributes = OrderedDict()
+        if isinstance(X, dd.DataFrame):
```
TomAugspurger: Could you check for dask or pandas DataFrame's here? Then the learned attributes will always be ndarrays.

cr458: Sounds good, have added a check for pandas DataFrames

TomAugspurger:
Added some release notes. This looks good, merging when the tests pass.

TomAugspurger:
Thanks @cr458!

cr458:
@TomAugspurger my pleasure. Keen to help out where I can as there is a lot of overlap between my research and the tools you are developing.

PR #172, opened 2018-05-21
Add CV / shuffle

TomAugspurger:
Array only right now.

Blockwise-only right now, to keep things performant.

Will followup with a helper `train_test_split` later tonight.


TomAugspurger:
Added docs, if anyone is able to take a look.

The basic version is: we can now do blockwise shuffling of dask arrays. Doing

```python
X_train, X_test, y_train, y_test = train_test_split(X, y)
```

will not have to shuffle any data between workers.

A version that *does* do a full shuffle, including between blocks, will hopefully be done tomorrow (separate PR).

TomAugspurger:
I'm going to merge this in a bit. Will continue to iterate on things, including some benchmarks, later. For now, I think the graph looks good (this is the graph for `train_test_split(X)`)

mydask

TomAugspurger:
> no A-like branches, each of which is an opportunity for communication in a distributed setting.

Just to confirm my understanding: with A-like branches, we could end up with the data on one machine (like `sub-#3`), and the indexes on another (like `concatenate-#1-, 0, 0`), which would require moving one?

TomAugspurger:
I think I'll still go ahead with merging this, since `ShuffleSplit().split(X)`, which generates the index arrays,  may still be useful.

In a followup we can optimize `train_test_split`, which returns the split arrays.

PR #184, opened 2018-05-30
MAINT: check if matrix is an ndarray in predict

stsievert:
This makes `model.score` more robust to types (and resolves an issue I had).


stsievert:
Updated. This only tests scoring and predicting with numpy arrays, not fitting with numpy arrays. I briefly looked into that, but the implementation isn't straightforward.

PR #185, opened 2018-05-30
Use default scheduler for training

TomAugspurger:


mrocklin:
Long term this should be fixed upstream.  Also, I think that we should move the upstream code in dask/array/learn.py to this repository.

TomAugspurger:
Agreed.

We're actually able to work around this easily by passing `get=client.get` to the `fit` call, so it may just be better to close the PR.

mrocklin:
I'm going to submit a PR moving an updated version of the
dask/array/learn.py code here

mrocklin:
This is now no longer necessary?

TomAugspurger:
Right.

PR #203, opened 2018-06-07
fix: outstanding occurrences of reblock replaced with rechunk

convexset:
I was just exploring the code and noticed the occurrence of reblock in

https://github.com/dask/dask/blob/master/dask/array/learn.py#L71
https://github.com/dask/dask/blob/master/dask/array/learn.py#L101

... I looked for the implementation and couldn't find any, and later found out that it was replaced with rechunk.

See: https://github.com/dask/dask/pull/3570

### DIFF
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -160,7 +160,7 @@ def fit(model, x, y, compute=True, **kwargs):
         assert x.chunks[0] == y.chunks[0]
     assert hasattr(model, 'partial_fit')
     if len(x.chunks[1]) > 1:
-        x = x.reblock(chunks=(x.chunks[0], sum(x.chunks[1])))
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks, this apparently isn't tested. Could you add a test that hits it? Something like

```python
X, y = make_classsification(chunks=20)
X = X.rechunk({1: 10})

assert X.numblocks[1] > 1

clf = SGDClassifier()
inc = Incremental(clf)
inc.fit(X, y)
```

convexset:
Hi, sorry, I'll get to this at some point. I've been using my spare time to do stuff in dask/dask/linalg.

convexset:
Done. =)

TomAugspurger:
Thanks @convexset!

PR #207, opened 2018-06-10
MAINT: make sure warm_start=True is specified

stsievert:
This PR implements a test to make sure that `warm_start=True` is specified if the estimator passed to `Incremental` supports it.

I found that even if `warm_start=False`, `Incremental` would reuse results to from previous calls to `fit`. This PR makes sure that relationship is specified explicitly.

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -284,6 +284,10 @@ class Incremental(ParallelPostFit):
     def __init__(self, estimator, scoring=None, **kwargs):
         estimator.set_params(**kwargs)
         super(Incremental, self).__init__(estimator=estimator, scoring=scoring)
+        if hasattr(estimator, 'warm_start') and not estimator.warm_start:
+            raise ValueError('Incremental requires warm_start=True so '
+                             'calls to est.fit will reuse results from '
+                             'previous calls')
 
     def fit(self, X, y=None, **fit_kwargs):
         result = fit(self.estimator, X, y, **fit_kwargs)
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -284,6 +284,7 @@ class Incremental(ParallelPostFit):
     def __init__(self, estimator, scoring=None, **kwargs):
         estimator.set_params(**kwargs)
         super(Incremental, self).__init__(estimator=estimator, scoring=scoring)
+
         if hasattr(estimator, 'warm_start') and not estimator.warm_start:
             raise ValueError('Incremental requires warm_start=True so '
                              'calls to est.fit will reuse results from '
--- a/tests/test_incremental.py
+++ b/tests/test_incremental.py
@@ -8,6 +8,7 @@ import sklearn.datasets
 from dask.array.utils import assert_eq
 from sklearn.base import clone
 from sklearn.linear_model import SGDClassifier
+from sklearn.utils.testing import all_estimators
 
 from dask_ml.wrappers import Incremental
 from dask_ml.utils import assert_estimator_equal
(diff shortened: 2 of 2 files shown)


mrocklin:
Thank you for identifying and resolving this @stsievert !  It looks like Python 2 tests are failing.  I suspect that this can be resolved by adding the following line to the top of the testing file

```python
from future import division
```

stsievert:
Resolved the Python 2 issue. I've also simplified the tests (by splitting into separate functions), and test this for all estimators (not only SGDClassifier).

TomAugspurger:
For SGDClassifier, anyway, IIUC, `partial_fit` uses the previous results, so there won't be any difference.

Do we always want to require `warm_start=True`?

stsievert:
> For SGDClassifier, anyway, IIUC, partial_fit uses the previous results

This is true, and also why I implemented it. I'm am not checking to make sure `partial_fit` reuses results, and am rather helping to clarify the API to make sure that the user expects `Incremental` to re-use results. `Incremental` implicitly assumes `warm_start=True`, and this PR makes sure it's explicitly specified so the user expects it.

TomAugspurger:
@stsievert do you think this is still necessary / a good idea?

IMO given that the entire premise of `Incremental` is incremental training via partial fit, requiring that the underlying estimator be instantiated with `warn_start=True` is unnecessary.

stsievert:
I'd still like to see this. I'm not worried about `warm_start=True` being unnecessary. The [sklearn introduction][1] says that

> Calling `fit()` more than once will overwrite what was learned by any previous `fit()`

Calling `Incremental.fit` will use previous results, which does not follow this convention. This would clarify user expectations, or maybe we should mention this more explicitly in the docs.

[1]:http://scikit-learn.org/stable/tutorial/basic/tutorial.html#refitting-and-updating-parameters

TomAugspurger:
Is the problem with `Incremental.fit` then? Should calling that "reset" the underlying estimator?

I was playing with some other meta-estimators from scikit-learn. It looks like we may be making a bad promise that the `Incremental.estimator` is actually the estimator trained.

For `sklearn.model_selection.GridSearchCV`, the `estimator` is not actually fit. It's cloned and then fit. Doing the same for `Incremental` would nicely solve this `warm_start` problem, and the problem of state leaking between `Incremental.fit` calls.

So my proposal is to

1. Remove the (implicit or explicit) promise that `Incremental.estimator` is actually fit
2. Change `Incremental.fit` to `clone` before calling fit
3. Maybe add an `estimator_` or `fitted_estimator_` attribute that has the actually trained estimator. This is analogous to `GridSearchCV.best_estimator_`

We'll also need to modify `Incremental.partial_fit` to be different from `Incremental.fit`, as it should *not* clone first.

Does that make any sense at all?

stsievert:
> 1.  Remove the (implicit or explicit) promise that Incremental.estimator is actually fit
> 2. Change Incremental.fit to clone before calling fit
> 3. Maybe add an estimator_ or fitted_estimator_ attribute that has the actually trained estimator. This is analogous to GridSearchCV.best_estimator_

This discussion is about `Incremental.fit`. Either way, `Incremental` will implement and optimize `partial_fit`. I see two choices:

1. have `Incremental` implement `fit`.
2. do not have `Incremental` implemented fit.

I don't think `Incremental.fit` should only do one pass over the data. That's a partial fit, and the data is not fitted until convergence like the name implies.

If `fit` is implemented, I think it should run until convergence because that's what the name implies. I think option 1 will be difficult because sklearn does not expose some property that indicates if the algorithm has converged. I believe the only thing we'd be able to do is make `max_iter` calls to `partial_fit`, unless we implement our own convergence functions.

I am leaning towards not implementing `fit`. Even besides the maintenance cost, I'm not sure the naming even implies we should implement `fit`. I'm okay leaving the API as is and deprecating `fit`.

TomAugspurger:
Given the recent updates to Incremental, is this no longer necessary?

Issue #197, opened 2018-06-06
SparseCoder


jakirkham:
Would be nice to wrap/implement [`SparseCoder`]( http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.SparseCoder.html ) for use with Dask-ML and Dask Arrays. This is useful for computing the sparse code that is associated with a precomputed dictionary.

jakirkham:
After peeling back a few layers, it seems the primary code lives in [this function]( https://github.com/scikit-learn/scikit-learn/blob/0.19.1/sklearn/decomposition/dict_learning.py#L27-L167 ). Basically it is just dispatching to different linear models in `scikit-learn`. So this should be trivial to build once those exist.

xref: https://github.com/dask/dask-ml/issues/101 (Lasso)
xref: https://github.com/dask/dask-ml/issues/104 (LassoLars)
xref: https://github.com/dask/dask-ml/issues/102 (Lars)

jakirkham:
Internally, `scikit-learn` appears to just be breaking the data into pieces, which it feeds into [`_sparse_encode`]( https://github.com/scikit-learn/scikit-learn/blob/0.19.1/sklearn/decomposition/dict_learning.py#L297-L308 ). Since the splitting is done just to make the jobs [roughly even]( https://github.com/scikit-learn/scikit-learn/blob/0.19.1/sklearn/decomposition/dict_learning.py#L295 ), it would certainly be reasonable to run the `SparseCoder` on existing chunks of a Dask Array that is being fit to as long as [the whole dictionary]( https://github.com/scikit-learn/scikit-learn/blob/0.19.1/sklearn/decomposition/dict_learning.py#L299 ) is provided. This would then use whatever parallelism Dask is setup to use.

[...]

PR #209, opened 2018-06-11
Call map_blocks with dtype in ParallelPostFit.fit

jakirkham:
Follow-up to comment ( https://github.com/dask/dask-ml/issues/197#issuecomment-396105384 )

Instead of letting `map_blocks` guess the `dtype` needed by `transform`, provide it the `dtype` of `X`. This may not always be right, but is more likely right than wrong. Further it gives the user a mechanism to fix the `dtype` returned by `map_blocks` when it's heuristics for detection fail, which is an improvement over the current behavior that does not supply such an option.

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -133,7 +133,7 @@ class ParallelPostFit(sklearn.base.BaseEstimator):
         transform = self._check_method('transform')
 
         if isinstance(X, da.Array):
-            return X.map_blocks(transform)
+            return X.map_blocks(transform, dtype=X.dtype)
         elif isinstance(X, dd._Frame):
             return _apply_partitionwise(X, transform)
         else:


jakirkham:
Oops...completely overlooked PR ( https://github.com/dask/dask-ml/pull/203 ). Closing in favor of that one.

jakirkham:
Agree this is a bit kludgy. Open to other suggestions. :)

TomAugspurger:
I'll need to think on it a bit. We may need to do something like http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html, where you initialize it with `kw_args` and `inv_kw_args`.

TomAugspurger:
Let's take the easy way out for now and allow a `dtype` parameter for `transform`.

```python
    def transform(self, X, dtype=None):
```

that will be passed through to `X.map_blocks(transform)`. Do you have an easy test case?

PR #211, opened 2018-06-11
TST: test the scoring kwarg for Incremental

stsievert:
I specified a `scoring` keyword arg to `Incremental` and ran into a bug: if I specified a `scoring` keyword arg, then an if-statement in `get_scorer` wasn't working or tested. This resolves that.

The tests primarily test to make sure no errors are thrown.

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -10,7 +10,7 @@ import sklearn.base
 
 from ._partial import fit
 from ._utils import copy_learned_attributes
-from .metrics import get_scorer
+from .metrics import get_scorer, check_scoring
 
 
 logger = logging.getLogger(__name__)
@@ -285,6 +285,7 @@ class Incremental(ParallelPostFit):
         super(Incremental, self).__init__(estimator=estimator, scoring=scoring)
 
     def fit(self, X, y=None, **fit_kwargs):
+        check_scoring(self.estimator, self.scoring)
         result = fit(self.estimator, X, y, **fit_kwargs)
 
         # Copy the learned attributes over to self
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -158,9 +158,8 @@ class ParallelPostFit(sklearn.base.BaseEstimator):
                 return self.estimator.score(X, y)
         """
         if self.scoring:
-            scoring = self.scoring
-            scorer = get_scorer(scoring)
-            return scorer(self.estimator, X, y)
+            scorer = get_scorer(self.scoring)
+            return scorer(self, X, y)
         else:
             return self.estimator.score(X, y)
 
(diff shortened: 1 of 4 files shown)


TomAugspurger:
Did you originally pass something like `scoring=dask_ml.metrics.scoring`. If so, I think that *should* raise, but we probably have a bad error message. That's a metric function, but the argument should be a scorer.

Our behavior would ideally be the same as scikit-learn here

```python
In [32]: from sklearn.linear_model import LogisticRegression

In [33]: import sklearn.metrics

In [34]: from sklearn.datasets import make_classification

In [35]: X, y = make_classification()

In [36]: gs = GridSearchCV(LogisticRegression(), {'penalty': ['L1', 'L2']}, scoring=sklearn.metrics.accuracy_score)

In [37]: gs.fit(X, y)
<ipython-input-37-ba603a2ebe90> in <module>()
----> 1 gs.fit(X, y)

~/sandbox/scikit-learn/sklearn/model_selection/_search.py in fit(self, X, y, groups, **fit_params)
    614
    615         scorers, self.multimetric_ = _check_multimetric_scoring(
--> 616             self.estimator, scoring=self.scoring)
    617
    618         if self.multimetric_:

~/sandbox/scikit-learn/sklearn/metrics/scorer.py in _check_multimetric_scoring(estimator, scoring)
    348     if callable(scoring) or scoring is None or isinstance(scoring,
    349                                                           six.string_types):
--> 350         scorers = {"score": check_scoring(estimator, scoring=scoring)}
    351         return scorers, False
    352     else:

~/sandbox/scikit-learn/sklearn/metrics/scorer.py in check_scoring(estimator, scoring, allow_none)
    290                              'require an estimator as its first parameter. '
    291                              'Please use `make_scorer` to convert a metric '
--> 292                              'to a scorer.' % scoring)
    293         return get_scorer(scoring)
    294     elif scoring is None:

ValueError: scoring value <function accuracy_score at 0x11258da60> looks like it is a metric function rather than a scorer. A scorer should require an estimator as its first parameter. Please use `make_scorer` to convert a metric to a scorer.
```

TomAugspurger:
scikit-learn is calling `sklearn.metrics.check_scorer` and (reasonably) hard-codes that nice exception message to scorers in `sklearn.metrics`, so we won't get it just by calling `check_scorer(dask_ml.metrics.accuracy_score)`.

stsievert:
Got it. This makes more sense now. I've updated the tests to reflect this.

dask_ml/wrappers.py
```diff
@@ -158,9 +158,8 @@ def score(self, X, y):
                 return self.estimator.score(X, y)
         """
         if self.scoring:
-            scoring = self.scoring
-            scorer = get_scorer(scoring)
-            return scorer(self.estimator, X, y)
+            scorer = get_scorer(self.scoring)
+            return scorer(self, X, y)
```
TomAugspurger: Oh, good catch. This should be self, not self.estimator.

TomAugspurger:
Thanks!

PR #216, opened 2018-06-20
MAINT: revert back to sklearn's train_test_split if not all dask arrays

### LINKED ISSUE #215: train_test_split only accepts dask arrays
Minimal working example:

``` python
from sklearn.datasets import fetch_20newsgroups
import numpy as np
import dask.array as da
import sklearn.model_selection as ms
from dask_ml.model_selection import train_test_split

data = fetch_20newsgroups(subset='train', categories=None)
X, y = data.data, data.target

assert isinstance(X, list)
assert isinstance(X[0], str)
assert isinstance(y, np.ndarray)
assert y.dtype == int

ms.train_test_split(X, y)
train_test_split(X, y)
```

Traceback:

```python-traceback
  File "bug.py", line 19, in <module>
    train_test_split(X, y)
  File "/Users/ssievert/anaconda3/lib/python3.6/site-packages/dask_ml/model_selection/_split.py", line 273, in train_test_split
    assert all(isinstance(arr, da.Array) for arr in arrays)
AssertionError
```

stsievert:
Fixes #215.

I ran into this while creating an example for https://github.com/dask/dask-searchcv/pull/72 that required the 20 newsgroups dataset. This is the easiest fix.

### DIFF
--- a/dask_ml/model_selection/_split.py
+++ b/dask_ml/model_selection/_split.py
@@ -12,6 +12,7 @@ from sklearn.model_selection._split import (
     _validate_shuffle_split_init,
     BaseCrossValidator,
 )
+import sklearn.model_selection as ms
 
 from dask_ml.utils import check_array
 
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks.

With this we'll handle everything but dask dataframes (https://github.com/dask/dask-ml/issues/219 for that, if you're interested.)

PR #217, opened 2018-06-20
Adds sample weights to accuracy_score

jrbourbeau:
This PR adds implements sample weighting for `accuracy_score`

### DIFF
--- a/dask_ml/metrics/classification.py
+++ b/dask_ml/metrics/classification.py
@@ -1,3 +1,7 @@
+
+import dask.array as da
+
+
 def accuracy_score(y_true, y_pred, normalize=True, sample_weight=None,
                    compute=True):
     """Accuracy classification score.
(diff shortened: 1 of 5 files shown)


dask_ml/metrics/classification.py
```diff
@@ -20,9 +24,8 @@ def accuracy_score(y_true, y_pred, normalize=True, sample_weight=None,
         If ``False``, return the number of correctly classified samples.
         Otherwise, return the fraction of correctly classified samples.
 
-    sample_weight : None
-        For compatibility with scikit-learn. This is currently not
-        supported.
+    sample_weight : 1d array-like, optional
+        Sample weights.
```
TomAugspurger: Could you add a

```
    .. versionadded:: 0.7.0
```

directive below `Sample weights`.

We should start doing this for all new features / keywords.

TomAugspurger:
Looks good overall.

Only thing might be how this behaves with dask < 0.18.0. If someone has dask 0.17, they'll get `AttributeError: module 'dask.array' has no attribute 'average'`. Maybe we could check the dask version and raise a nicer error message if it's too old?

jrbourbeau:
Good point, I'll check the dask version number and raise a `NotImplementedError` accordingly

jrbourbeau:
I thought checking the dask version might end up being useful elsewhere, so I added `DASK_VERSION` to `dask_ml._compat` (similar to the existing `SK_VERSION`).

TomAugspurger:
Sorry, thought this was merged.

jrbourbeau:
No probs, thanks @TomAugspurger!

PR #222, opened 2018-06-22
Updates contributing docs

jrbourbeau:
This PR adds a "Running tests" section to the contributing docs, along with a few other minor edits.

### DIFF
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -86,7 +86,7 @@ In addition to running tests, dask-ml verifies code style uniformity with the
 .. code-block:: none
 
     pip install flake8
-    flake8 dask_ml
+    flake8
 
 
 Conventions
--- a/docs/source/contributing.rst
+++ b/docs/source/contributing.rst
@@ -6,32 +6,39 @@ Thanks for helping to build ``dask-ml``!
 Cloning the Repository
 ~~~~~~~~~~~~~~~~~~~~~~
 
-Make a fork of the `repo <https://github.com/dask/dask-ml>`__ and clone that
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -23,7 +23,7 @@
 xgboost_requires = ['dask-xgboost', 'xgboost']
 complete_requires = tensorflow_requires + xgboost_requires
 
-extra_requires = {
+extras_require = {
```
stsievert: :+1:

docs/source/contributing.rst
```diff
@@ -50,9 +57,37 @@ And then
...
+    py.test tests/preprocessing
+
+
+In addition to running tests, dask-ml verifies code style uniformity with the
+``flake8`` tool:
+
+.. code-block:: none
+
+    pip install flake8
+    flake8 dask_ml
```
TomAugspurger: Could you change this to just `flake8`, so that the tests are linted too.

Side note: do either of you, or anyone else, have thoughts on using https://github.com/ambv/black to just auto-format everything for us? It'd be nice to not have to worry about style issues.

setup.py
TomAugspurger: whoops.

docs/source/contributing.rst
jrbourbeau: > Could you change this to just flake8, so that the tests are linted too.

Yeah, for sure. I'll update the circle builds too. 

> do either of you, or anyone else, have thoughts on using https://github.com/ambv/black to just auto-format everything for us?

I don't have any experience using black. Having code formatted automatically does seem appealing though.

TomAugspurger:
Thanks.

PR #223, opened 2018-06-22
MAINT: properly impelment check_scoring

stsievert:
This PR makes `check_scoring` return a callable as mentioned in the sklearn docs: http://scikit-learn.org/dev/modules/generated/sklearn.metrics.check_scoring.html. It returns the sklearn's `check_scoring` unless we've implemented it.

It also does two other maintenance items:

* Actually performs `assert` (gosh, did I really make that typo?)
* Gets rid of `FutureWarning`s from `SGDClassifier` by specifying `tol=1e-3`.


TomAugspurger:
Thanks!

PR #224, opened 2018-06-22
MAINT: accept ndarrays for input to fit and predict

stsievert:
This PR implements and tests

* creating dask arrays if the input args are np.ndarrays (with 1 chunk of the same shape) for both `fit` and `predict`
* having `predict` return the same type as the argument passed in

This generalizes to `Incremental` because it calls `fit`.

### DIFF
--- a/tests/test_partial.py
+++ b/tests/test_partial.py
@@ -6,6 +6,8 @@ from dask_ml._partial import fit, predict
 from dask_ml.datasets import make_classification
 from dask_ml.wrappers import Incremental
 
+import pytest
+
 
 x = np.array([[1, 0],
               [2, 0],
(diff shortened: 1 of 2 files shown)


tests/test_partial.py
```diff
@@ -28,16 +30,24 @@
...
         sol = sgd.predict(z)
-        result = predict(sgd, Z)
-        assert result.chunks == ((2, 2),)
-        assert result.compute().tolist() == sol.tolist()
+        result = predict(sgd, z_)
+
+        if isinstance(z_, da.Array):
+            assert result.chunks == ((2, 2),)
+            assert isinstance(result, da.Array)
+            assert result.compute().tolist() == sol.tolist()
```
TomAugspurger: FYI, we'd typically use `dask.array.utils.assert_eq` (I know you didn't write this test).

stsievert: Good to know. I thought that seemed clunky.

PR #221, opened 2018-06-21
ENH: Hyperband implementation

stsievert:
**What does this implement?**
This implements a state-of-the-art model selection algorithm, Hyperband. I give a concise summary in https://github.com/dask/dask-searchcv/pull/72#issue-192267118. Briefly, this algorithm only requires computational budget as input. Hyperband will find close to the best possible parameters with the given computational budget.* It performs the search via an adaptive scheme, and "kills" off poor performing models. The intuition for this is simple: it makes no sense to waste computation on poor performing models if the goal is to find the best performing model with as little computation as possible. More detail can be found in their paper: https://arxiv.org/abs/1603.06560

This will only work with models that implement `partial_fit`. I do not see this as blocking because optimization is iterative, and believe education and outreach will help. I aim to have this work for all classes that implement `score`, `partial_fit` and `{get, set}_params`.

<sup>* "will find" with high probability, "close" is within a log factor of the lower bound, "best" is highest expected cross-validation score.</sup>

**Reference Issues/PRs**
This addresses #161, and is an item on our beta roadmap (which has an issue: #210).

This is a refactor of https://github.com/dask/dask-searchcv/pull/72 and depends dask.distributed. The dependence on dask.distributed allows use of `as_completed`, which greatly simplifies the code and easily allows for an extension that makes this algorithm better suited for parallel architectures.

[1]:https://github.com/stsievert/dask-ml/blob/3208f6785cd7b9ac6803396542e634837fbbae32/dask_ml/model_selection/

**TODO**
- [x] implement and test Hyperband implementation is correct
- ~~implement and test support for scikit-learn models~~ done in #288
-  ~~implement logging~~ #356
- [x] implement and test ~~`DaskBaseSearchCV`~~ `AdaptiveSearchCV` inheritance
- [x] implement ~~two~~ one example~~s~~
- ~~feed random chunks to every partial_fit call~~ done in #288 
- [x] integrate with #356
- [x] supply arguments to create individual `SuccessiveHalving` brackets
- [x] implement reasonable patience parameter
- [x] documentation

**Future work**

* implement async variant
* implement Hyperband for black box functions

**Related work**
* https://github.com/scikit-learn/scikit-learn/pull/11266: `partial_fit` support in sklearn's `cross_validate`. This will yield more complete cross-validation (e.g., number of splits).
[...]

PR #225, opened 2018-06-23
MAINT: call model.fit directly if ndarrays passed to partial_fit

stsievert:
If `partial_fit` is called with NumPy arrays, why don't we call `partial_fit` directly with those arrays instead of making a new dask array?

I think this is the bug behind #221: `_partial_fit` would needlessly launch dask tasks.

### DIFF
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -167,7 +167,7 @@ def fit(model, x, y, compute=True, **kwargs):
     if isinstance(x, np.ndarray) and isinstance(y, np.ndarray):
         if hasattr(model, 'estimator'):
             estimator = _partial_fit(model.estimator, x, y, kwargs=kwargs)
-            model.estimator = estimator
+            copy_learned_attributes(estimator, model)
             return model
         return _partial_fit(model, x, y, kwargs=kwargs)
     if not (isinstance(x, da.Array) and
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -166,7 +166,9 @@ def fit(model, x, y, compute=True, **kwargs):
     assert x.ndim == 2
     if isinstance(x, np.ndarray) and isinstance(y, np.ndarray):
         if hasattr(model, 'estimator'):
-            return _partial_fit(model.estimator, x, y, kwargs=kwargs)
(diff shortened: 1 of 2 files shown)


stsievert:
I'm using `hasattr(model, 'estimator')` to get the underlying estimator. We should probably move this to a utils function; maybe two function in `wrappers.py` named `unwrap` and `wrap`?

dask_ml/_partial.py
```diff
@@ -164,10 +164,16 @@ def fit(model, x, y, compute=True, **kwargs):
     dask.array<x_11, shape=(400,), chunks=((100, 100, 100, 100),), dtype=int64>
     """
     assert x.ndim == 2
-    if isinstance(x, np.ndarray):
-        x = da.from_array(x, chunks=x.shape)
-    if isinstance(y, np.ndarray):
-        y = da.from_array(y, chunks=y.shape)
+    if isinstance(x, np.ndarray) and isinstance(y, np.ndarray):
```
TomAugspurger: It's possible that just `X` is passed, and `y` is None.

dask_ml/_partial.py
```diff
@@ -164,10 +164,16 @@ def fit(model, x, y, compute=True, **kwargs):
...
+        if hasattr(model, 'estimator'):
```
TomAugspurger: This seems fragile. Why not go through the `model`?

stsievert:
What tests do we want to pass? I see two cases, and these are reflected in the [types passed to `test_fit`][1]:

1. `fit` works regardless of model type and array type.
2. `fit` only works for dask-ml models and dask arrays.

The current implementation implements option 1 in #224.

I'd rather be explicit about types and go with option 2. Then `fit` will have a very specific job: build a graph using dask that fits on blocks of data, and will not be a function that works regardless of input type. This would involve reverting #224.

I think we should give a better error message if ndarrays are passed, and point to-be-implemented `unwrap` and `wrap` functions. I can implement these as part of #221.

Thoughts?

[1]:https://github.com/stsievert/dask-ml/blob/488177ac57f1239ab6c14aac9d25ff1c48bc2145/tests/test_partial.py#L34-L35

TomAugspurger:
2 sounds better to me as well.

stsievert:
#234 moves this to Incremental.partial_fit, which is what I was trying to do.

PR #226, opened 2018-06-25
Adds LabelEncoder

jrbourbeau:
This PR adds `LabelEncoder` as a drop-in replacement for `sklearn.preprocessing.LabelEncoder`.

### DIFF
--- a/dask_ml/preprocessing/label.py
+++ b/dask_ml/preprocessing/label.py
@@ -13,9 +13,13 @@ class LabelEncoder(sklabel.LabelEncoder):
 
     __doc__ = sklabel.LabelEncoder.__doc__
 
-    def fit(self, y):
+    def _check_array(self, y):
         if isinstance(y, dd.Series):
             y = da.asarray(y)
+        return y
+
+    def fit(self, y):
+        y = self._check_array(y)
 
         if isinstance(y, da.Array):
             classes_ = da.unique(y)
@@ -32,23 +36,21 @@ class LabelEncoder(sklabel.LabelEncoder):
 
     def transform(self, y):
         check_is_fitted(self, 'classes_')
-        if isinstance(y, dd.Series):
-            y = da.asarray(y)
+        y = self._check_array(y)
 
         if isinstance(y, da.Array):
-            return y.map_blocks(lambda b: np.searchsorted(self.classes_, b),
-                                dtype=self.classes_.dtype)
+            return da.map_blocks(np.searchsorted, self.classes_, y,
+                                 dtype=self.classes_.dtype)
         else:
             return np.searchsorted(self.classes_, y)
 
     def inverse_transform(self, y):
         check_is_fitted(self, 'classes_')
-        if isinstance(y, dd.Series):
-            y = da.asarray(y)
+        y = self._check_array(y)
 
         if isinstance(y, da.Array):
-            return y.map_blocks(lambda block: getitem(self.classes_, block),
-                                dtype=self.classes_.dtype)
+            return da.map_blocks(getitem, self.classes_, y,
+                                 dtype=self.classes_.dtype)
         else:
             y = np.asarray(y)
             return self.classes_[y]
--- a/dask_ml/preprocessing/label.py
+++ b/dask_ml/preprocessing/label.py
@@ -1,0 +1,52 @@
+from __future__ import division
+
+from operator import getitem
+
+import dask.array as da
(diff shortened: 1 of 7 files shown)


dask_ml/preprocessing/label.py
```diff
@@ -0,0 +1,54 @@
...
+from sklearn.preprocessing import label as sklabel
+from sklearn.utils.validation import check_is_fitted
+
+
+class LabelEncoder(sklabel.LabelEncoder):
+
+    __doc__ = sklabel.LabelEncoder.__doc__
+
+    def fit(self, y):
+        if isinstance(y, dd.Series):
```
TomAugspurger: Slight preference for moving this to a `LabelEnoder._check_array`. Then it can be re-used with `LabelEncodertransform`.

dask_ml/preprocessing/label.py
```diff
@@ -0,0 +1,54 @@
...
+        return self.fit(y).transform(y)
+
+    def transform(self, y):
+        check_is_fitted(self, 'classes_')
+        if isinstance(y, dd.Series):
+            y = da.asarray(y)
+
+        if isinstance(y, da.Array):
+            return y.map_blocks(lambda b: np.searchsorted(self.classes_, b),
+                                dtype=self.classes_.dtype)
```
mrocklin: Can I suggest placing `self.classes_` as a keyword argument and dropping the lambda?  Maybe something like the following:

```python
return y.map_blocks(lambda block, classes=None: np.searchsorted(classes, block), dtype=self.classes_.dtype, classes=self.classes_)
```

This avoids placing `self` into the `lambda`, which might cause issues when serializing and sending to other workers.

mrocklin: Or, something like the following might work as well and be a bit cleaner actually:

```python
da.map_blocks(np.searchsorted, self.classes_, y)
```

This is probably cheaper to serialize and also keeps the nice searchsorted function name for the diagnostics.

tests/preprocessing/test_label.py
```diff
@@ -0,0 +1,72 @@
...
+
+        assert_eq_ar(a.transform(array).compute(),
+                     b.transform(array.compute()))
+
+    @pytest.mark.parametrize('array', [y, s])
+    def test_inverse_transform(self, array):
+
+        a = dpp.LabelEncoder()
+        assert_eq_ar(a.inverse_transform(a.fit_transform(array)).compute(),
+                     array.compute())
```
mrocklin: Hopefully you don't need to call `.compute` here.  The `assert_eq` function should handle this for you.  It will also perform a number of other checks if it is provided dask objects directly.

dask_ml/preprocessing/label.py
jrbourbeau:
```python
return da.map_blocks(np.searchsorted, self.classes_, y, dtype=self.classes_.dtype)
```

Seems to work just fine and is cleaner in my opinion, thanks for the recommendation @mrocklin! Not sure why I used the lambda in the first place...

dask_ml/preprocessing/label.py
jrbourbeau: Can do @TomAugspurger, that seems reasonable to me

tests/preprocessing/test_label.py
jrbourbeau: Currently the `.tranform` method is implemented such that `LabelEncoder.transform(<dask-series>)` returns a dask array. It looks like neither the `assert_eq` in `dask.array.utils` nor `dask.dataframe.utils` can compare a dask array and a dask series (although there could be a utility function I'm unaware of that will do this). 

One way around this could be to remove the `.compute`s and wrap dask series with `da.asarray`. This way `assert_eq` will get two arrays instead of an array and a series. Something like

```python
if isinstance(array, da.Array):
    assert_eq_ar(a.inverse_transform(a.fit_transform(array)), array)
if isinstance(array, dd.Series):
    assert_eq_ar(a.inverse_transform(a.fit_transform(array)), da.asarray(array))
```

TomAugspurger: Does `dask_ml.utils._assert_eq` work for you?

TomAugspurger: Actually, how about `assert_eq_ar(a.inverse_transform(a.fit_transform(array)), da.asarray(array))`

TomAugspurger:
Thanks @jrbourbeau!

PR #235, opened 2018-06-27
Replaced channels with queues and other misc. cleanup.

laprej:
The current tf example code doesn't quite work as-is.  It also uses the outdated "channel" approach which dask now uses queues for.  This is a tiny change to bring things up to working order.


TomAugspurger:
Thanks!

These examples aren't currently run by our CI, but I've verified that things work locally with the latest version of tensorflow.

PR #237, opened 2018-06-27
CI: Use black for code formatting

TomAugspurger:
Just as an exercise to see if this would work.

I'm 50/50 on whether this is a good idea or not.


TomAugspurger:
If we do decide to go forward with this, we should wait till large outstanding PRs are in. This branch will be relatively easy to update.

TomAugspurger:
Also added coverage to the py36 build (turned off comments for now).

dask_ml/_partial.py
```diff
@@ -171,20 +170,28 @@ def fit(model, x, y, compute=True, **kwargs):
...
+        {
+            (name, i): (
+                _partial_fit,
+                (name, i - 1),
+                (x.name, i, 0),
+                (getattr(y, "name", ""), i),
+                kwargs,
+            )
+            for i in range(nblocks)
+        }
```
stsievert: I think I'm for black (it'd save me time), but this one-element-per-line is one formatting option I dislike, especially for many options.

stsievert: Black does have support for `# fmt: off` and `# fmt: on` flags to stop Black's formatting. I think we should use those if Black is used.

TomAugspurger: Hmm... Agreed that this is less pleasant.

However, I'm inclined to do all or nothing here. Locally, I would set up [a pre-commit hook](https://github.com/ambv/black#version-control-integration) and then never worry about it again :)

mrocklin:
+1 on all or nothing

TomAugspurger:
@stsievert will wait for your +/- 1 here.

Will also wait for large PRs (hypberband) to be merged first.

stsievert:
:+1:, though that's dependent on being able to use `# fmt` in future PRs

stsievert:
Should we add black's `.flake8` config? https://github.com/ambv/black/blob/master/.flake8 I got some errors about line length and binary operators after I formatted the Hyperband PR with black. They would have been resolved with their config.

> though that's dependent on being able to use # fmt in future PRs

Maybe not. Not having to think about formatting would be really nice.

TomAugspurger:
> Should we add black's .flake8 config? https://github.com/ambv/black/blob/master/.flake8 I got some errors about line length and binary operators after I formatted the Hyperband PR with black.

https://github.com/dask/dask-ml/pull/237/files#diff-380c6a8ebbbce17d55d50ef17d3cf906 has the changes to linting settings so that flake8 still passes.

> Not having to think about formatting would be really nice.

That's the hope :)

TomAugspurger:
Rebased on master. Apologies for the force push, but I think having a clean git commit history of 

1. configuration changes
2. auto-formatting
3. manual fixups

is useful here.

I'd like to merge this today if possible. I'll follow up with commits to outstanding PRs to use the new style.

TomAugspurger:
Here goes.

TomAugspurger:
Whoops, I just realized that I blew away the update to `contributing.rst`. Will do a followup now.

PR #241, opened 2018-06-28
Merge dask searchcv

TomAugspurger:
- https://github.com/dask/dask-ml/commit/53f60cb3420ab1d91b2a95f59c5b7a15a64732c4 has the "merge" commit
- https://github.com/dask/dask-ml/commit/7231d2a3e067110d59c1078c2bc27aba07083693 actually integrates the modules into Dask-ML (moving modules, updating imports, docs, CI, etc).

Note: we should **not** squash when merging.


TomAugspurger:
cc @jcrist

jcrist:
Looks good to me! 

Did you need to copy the license over? I'm not actually sure, since the license is the same.

TomAugspurger:
I wasn't sure either, and went with the safer option :)

jakirkham:
Should we also do this with dask-glm?

TomAugspurger:
Not sure yet.. Do you have thoughts?

TomAugspurger:
All green. Merging.

PR #236, opened 2018-06-27
Fit immediately in Incremental.partial_fit on concrete inputs

mrocklin:

PR #246, opened 2018-06-28
MAINT: ndarrays for score (and move ndarray logic from partial_fit to fit)

stsievert:
Two changes:

1. Modify `ParallelPostFit` to accepts score ndarrays if passed by using sklearn's `get_scorer`.
2. Move the logic in #236 to `Incremental.fit` (which works because `partial_fit` calls `fit`).

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -11,6 +11,7 @@ import sklearn.base
 from ._partial import fit
 from ._utils import copy_learned_attributes
 from .metrics import get_scorer, check_scoring
+from sklearn.metrics import get_scorer as sklearn_get_scorer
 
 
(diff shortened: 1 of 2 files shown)


dask_ml/wrappers.py
```diff
@@ -11,6 +11,7 @@
 from ._partial import fit
 from ._utils import copy_learned_attributes
 from .metrics import get_scorer, check_scoring
+from sklearn.metrics import get_scorer as sklearn_get_scorer
```
TomAugspurger: Slight preference for just using `import sklearn.metrics` and then `sklearn.metrics.get_scorer` (we'll soon have more space to work with :)

TomAugspurger:
Thanks @stsievert!

PR #253, opened 2018-06-29
DOC: Build with doctr

TomAugspurger:
Our RTD builds are consistently timing out.

This tries out doctr.


TomAugspurger:
Hmm I notice that we aren't (currently) using sphinx-gallery, so we could remove that and it's dependence on matplotlib (and qt) to cut down our build time.

TomAugspurger:
We'll I don't what the issue with the "training-on-large-datasets" notebook is, but it consistently hangs on travis. I thought it might be https://github.com/ipython/ipython/issues/5629, but the workarounds there didn't seem to help.

I'm going to merge this so that https://dask.github.io/dask-ml/ is updated, but I already see some things that need updating.

Issue #200, opened 2018-06-06
How to handle scoring for meta-estimators


TomAugspurger:
In #196, we came across an issue with the scoring API. For training an estimator on small NumPy arrays, a regular

```python
clf = sklearn.linear_model.SGDClassifier()
param_grid = {'alpha': [0.1, 10]}
grid_search = GridSearchCV(clf, param_grid, scoring=None)  # or scoring='accuracy'

with joblib.parallel_backend('dask'):
    grid_search.fit(X, y)
```

will induce calls to `grid_search.estimator.score(X_test, y_test)`. This is fine since `X` and `y` are small.

Now, we'd like to do something similar with `Incremental`, (the new wrapper to pass blocks of a dask array to the underlying `estimator.partial_fit`) which is itself a meta-estimator.

```python
clf = SGDClassifier()
inc = Incremental(clf)

grid_search = GridSearchCV(inc, param_grid, scoring=None)  # or scoring='accuracy'

with joblib.parallel_backend('dask'):
    grid_search.fit(X, y)  # X and y are *large* Dask arrays
```

As written, this will end up calling a scikit-learn scoring function with a CV split, which will potentially cause a memory error. The current workaround is to pass a `scoring` to `Incremental` or `GridSearchCV`

```python
from dask_ml.metrics import make_scorer
dask_scorer = get_scorer('accuracy')
inc = Incremental(clf, scoring=dask_scorer)
grid_search = GridSearchCV(inc, param_grid)
```

Then things will be fine, but that may be difficult to discover. A couple issues:

1. (less important) It's a bit unclear how `param_grid` should behave when `estimator` is a meta-estimator. I hacked up `Incremental.get/set_params` to pass things through. Maybe a syntax like `param_grid={'estimator.alpha': [0.1, 10]}` would make sense for scikit-learn? Has this come up before?
2. (more important): The string `'accuracy'` will get scikit-learn's accuracy scorer. This is unfortunate since `X_test` and `y_test` will be large dask Arrays, which will maybe cause a memory error when passed the the scikit-learn version. A few potential solutions
  a.) Do nothing, just write better documentation. Maybe make `scorer` a required argument to `Incremental`?
  b.) Rewrite the scikit-learn scorers to be compatible with both ndarrays and dask arrays (is that possible?)
  c.) A way to introspect an estimator to get the scorer used by `Estimator.score`. Something like

```python
from sklearn.metrics import get_estimator_scorer  # or just put in get_scorer
[...]

PR #258, opened 2018-07-02
API/REF: Clone before fitting in Incremental.fit

TomAugspurger:
This changes Incremental.fit to clone before fitting.

Alternative to https://github.com/dask/dask-ml/pull/207

### DIFF
--- a/tests/test_incremental.py
+++ b/tests/test_incremental.py
@@ -15,7 +15,8 @@ from dask_ml.metrics.scorer import check_scoring
 
 def test_incremental_basic(scheduler, xy_classification):
     X, y = xy_classification
-    with scheduler() as (s, [a, b]):
+
+    with scheduler() as (s, [_, _]):
         est1 = SGDClassifier(random_state=0, tol=1e-3)
         est2 = clone(est1)
 
@@ -26,11 +27,12 @@ def test_incremental_basic(scheduler, xy_classification):
 
         assert result is clf
 
-        assert isinstance(result.estimator.coef_, np.ndarray)
-        np.testing.assert_array_almost_equal(result.estimator.coef_,
+        assert isinstance(result.estimator_.coef_, np.ndarray)
+        np.testing.assert_array_almost_equal(result.estimator_.coef_,
                                              est2.coef_)
 
-        assert_estimator_equal(clf.estimator, est2, exclude=['loss_function_'])
+        assert_estimator_equal(clf.estimator_, est2,
+                               exclude=['loss_function_'])
 
         #  Predict
         result = clf.predict(X)
@@ -46,7 +48,8 @@ def test_incremental_basic(scheduler, xy_classification):
 
         clf = Incremental(SGDClassifier(random_state=0, tol=1e-3))
         clf.partial_fit(X, y, classes=[0, 1])
-        assert_estimator_equal(clf.estimator, est2, exclude=['loss_function_'])
+        assert_estimator_equal(clf.estimator_, est2,
+                               exclude=['loss_function_'])
 
 
 def test_in_gridsearch(scheduler, xy_classification):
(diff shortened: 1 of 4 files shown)


TomAugspurger:
cc @stsievert

tests/test_incremental.py
```diff
@@ -26,11 +27,12 @@ def test_incremental_basic(scheduler, xy_classification):
 
         assert result is clf
 
-        assert isinstance(result.estimator.coef_, np.ndarray)
-        np.testing.assert_array_almost_equal(result.estimator.coef_,
+        assert isinstance(result.estimator_.coef_, np.ndarray)
```
TomAugspurger: This is the API change.

tests/test_incremental.py
```diff
@@ -96,22 +99,20 @@ def test_scoring_string(scheduler, xy_classification, scoring):
         assert callable(check_scoring(clf, scoring=scoring))
         clf.fit(X, y, classes=np.unique(y))
         clf.score(X, y)
-        clf.estimator.score(X, y)
```
TomAugspurger: I don't think this was doing what we wanted earlier.

When Incremental is created with `scoring`, `clf.score(X, y)` will in general be different then `clf.estimator_.score(X, y)`. The latter always uses the "default" scoring for that estimator (e.g. accuracy score). I opted to just remove this line.

tests/test_incremental.py
```diff
@@ -96,22 +99,20 @@ def test_scoring_string(scheduler, xy_classification, scoring):
...
 
 
 def test_fit_ndarrays():
     X = np.ones((10, 5))
-    y = np.ones(10)
+    y = np.concatenate([np.zeros(5), np.ones(5)])
```
TomAugspurger: Needed a mix of zeros and ones. I'm not entirely sure what changed to necessitate that...

tests/test_incremental.py
```diff
@@ -122,7 +123,7 @@ def test_score_ndarrays():
     inc = Incremental(sgd, scoring='accuracy')
 
     inc.partial_fit(X, y, classes=[0, 1])
-    inc.fit(X, y)
+    inc.fit(X, y, classes=[0, 1])
```
TomAugspurger: need this, as `classes_` no longer leasks through the `partial_fit` to the `fit` (a good thing!)

stsievert:
I think if we implement `fit` support for `Incremental` we should loop over the data `max_iter` times beforehand.

For more complete thoughts on the API, see https://github.com/dask/dask-ml/pull/207#issuecomment-401834441

TomAugspurger:
Agreed that having a `max_iter` parameter would be useful for Incremental. However, exposes a difficulty alluded to in https://github.com/dask/dask-ml/issues/200 and how `Incremental.get/set_params` works. It would be ambiguos whether `max_iter` should go to `Incremental.max_iter` or `Incremental.estimator.max_iter`.

The solution is to probably to have `Incremental` to specify a name when creating the `Incremental`.

See https://stackoverflow.com/a/35534452/1889400 for how `VotingClassifier` works.

So either

```python
inc = Incremental(SGDClassifier(), name='sgdclassifier')
inc = Incremental(('sgdclassifier', SGDClassifier())
```

In the latter case, we would have a `make_incremental` helper that works like `make_pipeline`.

Then in a context where you need to get / set params, it'd be

```python
param_grid = {
    'sgdclassifier__alpha': [0.1, 10.0],
    'sgdclassifier__max_iter': [1000, 2000],
    'max_iter': [1, 5],
}

clf = GridSearchCV(inc, param_grid)
```

TomAugspurger:
Looks like https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/metaestimators.py has some utilities for get / set params in a meta-estimator.

TomAugspurger:
55fccf20e0c2204f5149db7161047ff31184ba85 has the changes.

Turns out BaseEstimator already does the right thing, we don't have to mess with `_BaseComposition`.

I'm encouraged by the lines of code we could delete from `Incremental`.

For now, I haven't made the name configurable. If you want to set the parameters for the underlying estimator, then you use `estimator__<parameter>`. We can make the `estimator` part configurable in the future if there's a demand for it.

TomAugspurger:
I'll implement `max_iter` now.

@stsievert, do you have a feeling for what a good default would be? I'm leaning towards `max_iter=1`, but am not sure.

mrocklin:
What does Scikit-Learn do in situations like these?  Do they have some sort
of convergence criterion that they check?  Perhaps we can do the same?

TomAugspurger:
It varies from model to model.

https://github.com/scikit-learn/scikit-learn/issues/1626 is the general API that we would want to follow here, when that is implemented.

TomAugspurger:
A basic version of `max_iter` should be working now.

TomAugspurger:
I'm finding this diff somewhat difficult to digest. I'm going to split off the `max_iter` to a second PR.

stsievert:
I'm on board with this now. I think `Incremental` should implement `fit` and have a  `max_iter` param.

> do you have a feeling for what a good default would be? I'm leaning towards max_iter=1, but am not sure.

I think I would default to `max_iter=5` – that's what sklearn's SGDClassifier `max_iter` defaults to, at least when no stopping criterion set in <v0.21. They set some stopping criteria in >= v0.21 and increase `max_iter`. For simple problems I'd say that 5 epochs is good enough.

This *should* depend on model complexity but I'm okay leaving that. e.g., fitting a linear model with twice as many features will require about `sqrt(2)` more samples to be seen by the optimization algorithm.

Issue #249, opened 2018-06-29
GridSearchCV does not pick up default Client


mrocklin:
```python
from dask.distributed import Client
client = Client()

import numpy as np
from sklearn.linear_model import SGDClassifier

model = SGDClassifier(    
    max_iter=1.0, 
)

X = np.random.random((1000, 10))
y = np.random.randint(0, 10, size=(1000,))

from dask_ml.model_selection import GridSearchCV
params = {'alpha': np.logspace(-4, 1, num=20)}
grid = GridSearchCV(model, params, return_train_score=True)
_ = grid.fit(X, y)

assert client.cluster.scheduler.transition_log
```

My guess is that this changed in the 0.18.0 Dask update.  The right function to use today is maybe 

https://github.com/dask/dask/blob/master/dask/base.py#L841-L853

PR #260, opened 2018-07-02
Use default scheduler sniffing logic from dask

mrocklin:
This removes the _normalize_scheduler function from dask_searchcv and
some of the scheduling names (threaded, sequential) and replaces them
with the Dask logic.

This both unifies logic across projects and also lets the distributed
scheduler take over when it is set as default.

Fixes https://github.com/dask/dask-ml/issues/249


TomAugspurger:
All green, thanks!

PR #266, opened 2018-07-03
ENH: max_iter for Incremental

### LINKED ISSUE #264: Multiple epochs for Incremental
Incremental should have a `max_iter` hyperparameter to control the number of passes (epochs) over the training data.

Implementing a `max_iter` is relatively straightforward. What we really want is some kind of "as many epochs as necessary until the underlying estimator has converged" parameter, which seems relatively difficult.

Split from https://github.com/dask/dask-ml/pull/258

TomAugspurger:
See https://github.com/dask/dask-ml/issues/264

I don't think this closes #264. We would want some kind of actual stopping criterion for that.

For now, this provides a convenient way to make multiple passes over the dataset.

I'm not at all sure what the default should be. In https://github.com/dask/dask-ml/pull/258#issuecomment-401954237 @stsievert suggested 5.

Having `max_iter=1` by default makes explaining / understanding `Incremental` easier. "Dask-ML just makes the `partial_fit` calls for you". But I'm not sold on that.

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -323,6 +323,11 @@ class Incremental(ParallelPostFit):
            a single NumPy array, which may exhaust the memory of your worker.
            You probably want to always specify `scoring`.
 
+    max_iter : int, default 1
+        The maximum number of passes over the training data (aka epochs).
+        It only impacts the behavior in the ``fit`` method, and not
+        `partial_fit`.
+
     Attributes
     ----------
     estimator_ : Estimator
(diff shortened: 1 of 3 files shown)


dask_ml/wrappers.py
```diff
@@ -323,6 +323,11 @@ class Incremental(ParallelPostFit):
            a single NumPy array, which may exhaust the memory of your worker.
            You probably want to always specify `scoring`.
 
+    max_iter : int, default 1
```
mrocklin: The use of `max` in the name is strange to me here.  It looks like we always choose this number.  Perhaps this is a convention though?

stsievert: Should the default be 1? That seems low. I'd choose a higher value, at least 10.

PR #270, opened 2018-07-03
[WIP] Wrap ``sklearn.model_selection`` routines with dask

### LINKED ISSUE #251: Add cross_validate helper
`sklearn.model_selection.cross_validate`  fits and scores several models over some CV splits of data.

Users can currently do this distributed on a cluster with

```python
with joblib.parallel_backend('dask', scatter=[X, y]):
    sklearn.model_selection.cross_validate(estimator, X, y)
```

Why not do that for them by defining a `dask_ml.model_selection.cross_validate` that does the `parallel_backend` and `scatter`ing for them?

gglanzani:
The commit adds ``cross_validate, cross_val_score``, and
``cross_val_predict`` methods that wrap the sklearn equivalent methods
by using ``joblib`` with dask as a backend.

It should close #251

@TomAugspurger I've marked it as WIP to see if this is the direction you would like the PR to go in order to close the issue. Let me know otherwise.

### DIFF
--- a/dask_ml/model_selection/_validate.py
+++ b/dask_ml/model_selection/_validate.py
@@ -1,0 +1,46 @@
+import functools
+
+from sklearn.model_selection import cross_validate as _cross_validate
+from sklearn.model_selection import cross_val_score as _cross_val_score
+from sklearn.model_selection import cross_val_predict as _cross_val_predict
+from dask.base import is_dask_collection
+
+from dask_ml import joblib
+
+
+def _joblibify(f):
+    """
+    Allows execution of functions using joblib('dask') as backend.
+
+    :param f: A function with signature ``f(something, X, y=None, **kwargs)``
+    :return: A function that runs on dask
+    """
+    @functools.wraps(f)
+    def wrap(*args, **kwargs):
+        X = args[1]
+        y = kwargs.get("y")
+
+        if not y and len(args) == 3:
+            y = args[2]
+
+        if not y:  # unsupervised
+            scatter = [X]
+        else:
+            scatter = [X, y]
+
+        if is_dask_collection(X) or is_dask_collection(y):
+            message = ('Dask collections are not supported '
+                       'by {} yet. '.format(f.__name__))
+            message += ('You can explicily compute them (if you data fit '
+                        'in memory) through ``X, y = dask.compute(X, y)``')
+            raise NotImplementedError(message)
+
+        with joblib.parallel_backend("dask", scatter=scatter):
+            return f(*args, **kwargs)
+
+    return wrap
+
+
+cross_validate = _joblibify(_cross_validate)
+cross_val_score = _joblibify(_cross_val_score)
+cross_val_predict = _joblibify(_cross_val_predict)
--- a/dask_ml/model_selection/__init__.py
+++ b/dask_ml/model_selection/__init__.py
@@ -8,6 +8,7 @@ from ._search import (
     compute_n_splits, check_cv
 )
 from ._split import ShuffleSplit, train_test_split
+from ._validate import cross_validate, cross_val_score, cross_val_predict
 
 
 __all__ = [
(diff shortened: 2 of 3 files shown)


gglanzani:
Always nice when the tests fail :)

It's currently failing because of this https://github.com/scikit-learn/scikit-learn/commit/106bb9e7f5e63b1b5b8862d1d834295ce6cd6c25

We're testing against sklearn-dev and @GaelVaroquaux renamed `sklearn.externals.joblib` to `sklearn.externals._joblib`.

Since the failure comes from `distributed`, I've opened a PR [there](https://github.com/dask/distributed/pull/2091).

GaelVaroquaux:
> It's currently failing because of this scikit-learn/scikit-learn@106bb9e

> We're testing against sklearn-dev and @GaelVaroquaux renamed
> sklearn.externals.joblib to sklearn.externals._joblib.

Indeed, and there is a half-done PR to address it in distributed:
https://github.com/dask/distributed/pull/2019

That said, we are discussing in scikit-learn whether we should revert
this API change, as it is breaking too many things.

gglanzani:
@GaelVaroquaux sorry, didn't see that one! I've closed PR 2091!

dask_ml/model_selection/_validate.py
```diff
@@ -0,0 +1,43 @@
...
+        if not y and len(args) == 3:
+            y = args[2]
+
+        if not y:  # unsupervised
+            scatter = [X]
+        else:
+            scatter = [X, y]
+
+        if is_dask_collection(X) or is_dask_collection(y):
+            raise NotImplementedError('This method is not implemented '
```
TomAugspurger: This should note that it isn't implemented for *dask collections* yet.

Perhaps you could indicate that users can pre-compute `X` and `y` if they fit in memory.

```
If your data fit in RAM on each worker, you can pre-compute the data with

X, y = dask.compute(X, y)

before calling {f.__name__}
```

TomAugspurger (commented):
This looks really good, thanks.

Could you also add these to helpers to `docs/source/modules/api.rst`, and add a basic test using it. Let me know if you need help writing a test.

We'll ignore the sklearn-dev failures for now, until there's a clear direction forward.

mrocklin:
I'm going to push back a bit (though please don't put to much weight here, I don't have much user experience here).  We might want to avoid explicitly wrapping sklearn operations if we can avoid it by just asking users to use sklearn with joblib.  This might reduce maintenance in the future, and keep their code sklearn friendly when they don't want dask involved.

Or, perhaps a different way of saying this:

What is the advantage of doing this vs just having the user wrap code with the joblib dask backend?

TomAugspurger:
That's a fair concern. These utilities would blur the line between "scikit-learn + distributed joblib for CPU bound problems" vs. Dask-ML for RAM-bound problems.

@gglanzani do you have any thoughts on that?

gglanzani:
@mrocklin I see your point. This machinery might indeed make *more* sense once these methods are implemented also on dask collections but as it now stands it has another advantage, namely if my code now reads:

```python
from sklearn.model_selection import cross_validate

def my_fun(*args, **kwargs):
    do_something()
    cross_validate(...)


def my_fun2(*args, **kwargs):
    do_something_else()
    cross_validate(...)
```

I don't need to change all the instances of `cross_validate`, but I need only to change the import to dask. So personally this has value for the end user. If this value is justified from a development perspective… I'm not qualified for this :)

@TomAugspurger Can you point me to a test which I could learn/copy from for this specific case?
I will also update the error message and update the docs.

mrocklin:
In your case with the new joblib you can call the following at the top of your script

    joblib.parallel_backend('dask')

And the rest of your code will work as before, keeping the sklearn import

gglanzani:
@mrocklin in that case I have to pass the ball to Tom who originally opened the issue 😄

TomAugspurger:
I'll think on it for a bit. Apologies for the (potentially) unused work @gglanzani.

gglanzani:
@mrocklin @TomAugspurger one extra advantage of the above approach is that it's still easy to use joblib with dask backend only on parts of the functions in the module.

But that's a minor thing.

Since we're talking about this, what would the best venue to talk about how to implement the functions for dask collections as well?

PR #275, opened 2018-07-03
MAINT: remove tokenize in partial_fit

### LINKED ISSUE #261: tokenize in Incremental.partial_fit is expensive
The calls to `dask.base.tokenize` in

https://github.com/dask/dask-ml/blob/d5801584d092d8f13f1b38aaf4da5dc3caa6a213/dask_ml/_partial.py#L108-L112

to `dask.base.tokenize` can take a sizable amount of time relative to the actual `partial_fit`. They could be replaced with simple notes that partial fit started / finished.

They could also probably be `debug` level.

stsievert:
Closes #261 

With this I've seen a speedup of about 2x for my example notebook in https://github.com/dask/dask-examples/pull/15, which is supported by the dask dashboard profiling graph.

### DIFF
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -105,11 +105,13 @@ class _BigPartialFitMixin(object):
 def _partial_fit(model, x, y, kwargs=None):
     kwargs = kwargs or dict()
     start = tic()
-    logger.info("Starting partial-fit %s", dask.base.tokenize(model, x, y))
+    if logger.getEffectiveLevel() == logging.DEBUG:
+        token = dask.base.tokenize(model, x, y)
+        logger.debug("Starting partial-fit %s", token)
     model.partial_fit(x, y, **kwargs)
     stop = tic()
-    logger.info("Finished partial-fit %s [%0.2f]",
-                dask.base.tokenize(model, x, y), stop - start)
+    if logger.getEffectiveLevel() == 'debug':
+        logger.debug("Finished partial-fit %s [%0.2f]", token, stop - start)
     return model
 
 
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -4,7 +4,6 @@ import logging
 import os
 import warnings
 from abc import ABCMeta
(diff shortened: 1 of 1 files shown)


stsievert:
This isn't tested, at least not in this code base. Should it be tested here?

dask_ml/_partial.py
```diff
@@ -105,11 +105,13 @@ def _get_predict_dtype(self, X):
 def _partial_fit(model, x, y, kwargs=None):
     kwargs = kwargs or dict()
     start = tic()
-    logger.info("Starting partial-fit %s", dask.base.tokenize(model, x, y))
+    if logger.getEffectiveLevel() == logging.DEBUG:
+        token = dask.base.tokenize(model, x, y)
+        logger.debug("Starting partial-fit %s", token)
```
mrocklin: We just shouldn't call tokenize here at all.  It's not strictly necessary.  I suspect that these logging lines were added during a debug session.  I think that we should remove them.

mrocklin: More generally, I think that we probably shouldn't be logging at this fine of granularity.  A user looking through logs would not be happy learning that there were thousands of lines every time they called fit on a dask-ml model

stsievert: Ah, good point. I've removed them.

TomAugspurger:
Just some linting issues now. The timer import can be removed.

mrocklin:
This looks good to me.  However looking at the code now it also raises a question:

Shouldn't we be cloning the model here before calling partial fit?  Otherwise we're mutating the input.

TomAugspurger:
Cloning doesn't copy over attributes like `coef_`, meaning we would start fresh on each call to `partial_fit`.

Are you worried that mutating the input will cause scheduling issues?

On master, we currently "own" this estimator, as we clone in `Incremental.fit` before passing it down here.

mrocklin:
What if we have to rerun this task because the worker that the result was on failed?

mrocklin:
My comment above though can be addressed separately.  +1 on this PR

PR #276, opened 2018-07-03
MAINT: randomly sample chunks for partial_fit calls

### LINKED ISSUE #274: Incremental.partial_fit does not randomly shuffle blocks on repeated calls
Or blocks are always fed into `estimator.partial_fit` in the same order. Here's an example of repeated `partial_fit` calls (interspersed by `score` calls):

screen shot 2018-07-03 at 11 29 11 am

I am sure this does not matter for the test data we create, but it might matter for other data (e.g., time series).

stsievert:
Closes #274.

A screenshot of different `partial_fit` calls:

| Before (almost) | After |
|:--:|:--:|
|screen shot 2018-07-03 at 11 06 04 am | screen shot 2018-07-03 at 1 07 36 pm |

I only show the before for some comparison, though it's not exact (and was in the debug process). It is "almost" because it uses a fixed seed every time, not a random seed. It does not run through the blocks in sequential order as mentioned in the previous docs.

### DIFF
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -121,8 +121,8 @@ def fit(model, x, y, compute=True, shuffle_blocks=True, random_state=None,
     Model must support the ``partial_fit`` interface for online or batch
     learning.
 
-    This method will be called on dask arrays in sequential order.  Ideally
-    your rows are independent and identically distributed.
+    Ideally your rows are independent and identically distributed. By default,
+    this function will step through chunks of the arrays in random order.
 
     Parameters
     ----------
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -13,7 +13,6 @@ import sklearn.utils
 
 import dask
 from dask.delayed import Delayed
-import dask.array as da
 
 from ._utils import copy_learned_attributes
 
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -5,6 +5,7 @@ import os
 import warnings
 from abc import ABCMeta
 from timeit import default_timer as tic
+import random
 
 import numpy as np
 import six
@@ -171,11 +172,13 @@ def fit(model, x, y, compute=True, **kwargs):
         x = x.rechunk(chunks=(x.chunks[0], sum(x.chunks[1])))
 
     nblocks = len(x.chunks[0])
+    order = list(range(nblocks))
+    random.shuffle(order)
 
-    name = 'fit-' + dask.base.tokenize(model, x, y, kwargs)
+    name = 'fit-' + dask.base.tokenize(model, x, y, kwargs, order)
     dsk = {(name, -1): model}
     dsk.update({(name, i): (_partial_fit, (name, i - 1),
-                                          (x.name, i, 0),
+                                          (x.name, order[i], 0),
                                           (getattr(y, 'name', ''), i), kwargs)
                 for i in range(nblocks)})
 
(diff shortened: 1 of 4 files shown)


stsievert:
I've had to reformat `test_incremental_basic` to train longer and assert that the results are close enough, not exactly the same.

dask_ml/_partial.py
```diff
@@ -171,12 +172,15 @@ def fit(model, x, y, compute=True, **kwargs):
         x = x.rechunk(chunks=(x.chunks[0], sum(x.chunks[1])))
 
     nblocks = len(x.chunks[0])
+    order = list(range(nblocks))
+    random.shuffle(order)
```
TomAugspurger: For reproducibility, Incremental should take a `random_state` and then use that for shuffling the data.

I also think the shuffling should be optionally enabled / disabled (not sure what the default should be) by a hyper parameter.

Finally, I'm not sure what the parameter name should be. Typically scikit-learn uses `shuffle=True/False`. We can either re-use that, or call it `shuffle_blocks`, and reserve `shuffle` for when we actually shuffle data between blocks.

stsievert: I think shuffling should be enabled by default. That's what sklearn default too, and it's what SGD theory points too.

I've used `shuffle_blocks`; that describes exactly what it does. We can use `shuffle` for full shuffling, and `shuffle_each_block` to shuffle in each block.

stsievert:
I have actually tested this now, and believe it's ready for merge.

mrocklin:
This all seems fine to me.  The sklearn-dev failures are interesting.  Do we know if it is due to this PR or something upstream?

stsievert:
I don't see how it could be this PR. All the failures are with `RobustScaler`, which doesn't call `fit` or `Incremental`.

mrocklin:
OK.  Fixing it separately here: https://github.com/dask/dask-ml/pull/283

stsievert:
There was a PR on `RobustScalar` that was merged earlier this morning: https://github.com/scikit-learn/scikit-learn/pull/11308.

mrocklin:
That would help explain things :)

tests/test_incremental.py
```diff
@@ -6,9 +6,9 @@
 from dask.array.utils import assert_eq
 from sklearn.base import clone
 from sklearn.linear_model import SGDClassifier
+import numpy.linalg as LA
```
mrocklin: The `LA` pronoun isn't very common.  I'd prefer the fully qualified name if possible.

```python
import numpy as np


np.linalg.svd(...)
```

dask_ml/wrappers.py
```diff
@@ -323,6 +323,12 @@ class Incremental(ParallelPostFit):
            a single NumPy array, which may exhaust the memory of your worker.
            You probably want to always specify `scoring`.
 
+    random_state : int or numpy.random.RandomState
```
TomAugspurger: Append `, optional`.

dask_ml/wrappers.py
```diff
@@ -323,6 +323,12 @@ class Incremental(ParallelPostFit):
...
+        Random object that determines how to shuffle blocks.
+
+    shuffle_blocks : bool
```
TomAugspurger: append `, default True`.

dask_ml/wrappers.py
```diff
@@ -323,6 +323,12 @@ class Incremental(ParallelPostFit):
...
+        Whether to randomly shuffle the blocks or now
```
TomAugspurger: typo: now -> not.

Could you elaborate on what this means? Specifically, clarify shuffling the block order from shuffling within blocks (from shuffling between blocks).

stsievert:
Thanks for the review @mrocklin and @TomAugspurger. I've addressed those comments.

TomAugspurger:
Thanks!

PR #280, opened 2018-07-03
Add sample_weight for regression metrics

jrbourbeau:
This PR adds support for using `sample_weight` with the existing regression metrics (i.e. `mean_squared_error`, `mean_absolute_error`, and `r2_score`)

### DIFF
--- a/dask_ml/metrics/regression.py
+++ b/dask_ml/metrics/regression.py
@@ -1,19 +1,29 @@
 import six
 
+import packaging.version
 import dask.array as da
 from dask.array.random import doc_wraps
 import sklearn.metrics
 
+from .._compat import DASK_VERSION
+
 
 def _check_sample_weight(sample_weight):
     if sample_weight is not None:
-        raise ValueError("'sample_weight' is not supported.")
+        no_average = DASK_VERSION <= packaging.version.parse("0.18.0")
+        if no_average:
+            raise NotImplementedError("'sample_weight' is only supported for "
+                                      "dask versions > 0.18.0.")
 
 
 def _check_reg_targets(y_true, y_pred, multioutput):
-    if multioutput != 'uniform_average':
-        raise NotImplementedError("'multioutput' must be 'uniform_average'")
-
+    allowed_multioutput_str = ('raw_values', 'uniform_average',
+                               'variance_weighted')
+    if isinstance(multioutput, six.string_types):
+        if multioutput not in allowed_multioutput_str:
+            raise ValueError("Allowed 'multioutput' string values are {}. "
+                             "You provided multioutput={!r}"
+                             "".format(allowed_multioutput_str, multioutput))
     if y_true.ndim == 1:
         y_true = y_true.reshape((-1, 1))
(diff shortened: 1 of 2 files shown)


dask_ml/metrics/regression.py
```diff
@@ -88,7 +108,20 @@ def r2_score(y_true, y_pred, sample_weight=None,
     output_scores[valid_score] = 1 - (numerator[valid_score] /
                                       denominator[valid_score])
     output_scores[nonzero_numerator & ~nonzero_denominator] = 0.
-    result = output_scores.mean(axis=0)
+
+    if isinstance(multioutput, six.string_types):
+        if multioutput == 'raw_values':
+            # return scores individually
+            return output_scores.compute() if compute else output_scores
```
mrocklin: Why the compute call here and not elsewhere?  

Please forgive my ignorance, but what are our current policies around automatically computing scores vs letting the user call compute themselves?

TomAugspurger: I'll need to verify, I believe the default `compute=True` came out of trying to get sklearn.model_selection.GridSearchCV to work well with Incremental.

Initially, I had it lazy, but something failed. Apologies for not having logged the output of what happens when scikit-learn receives a scalar dask array as a score, and for not digging further on it.

jrbourbeau: @TomAugspurger you're right about `sklearn.model_selection.GridSearchCV` not working well with scalar dask arrays as a score. The example snippet below is fine when `compute=True`, but fails when `compute=False`

```python
from functools import partial

from dask_ml.wrappers import Incremental
from dask_ml.metrics import accuracy_score

from sklearn.metrics import make_scorer
from sklearn.model_selection import GridSearchCV

# Get dataset
X, y = make_classification(chunks=25)

estimator = SGDClassifier(random_state=10)
acc = partial(accuracy_score, compute=False)
acc.__name__ = 'accuracy_score' # make_scorer looks for __name__
clf = Incremental(estimator, scoring=make_scorer(acc))

gs = GridSearchCV(clf, param_grid={'estimator__loss': ['hinge', 'log']})
gs.fit(X, y, classes=[0, 1])
```

When `compute=False`, calling `GridSearchCV.fit` raises 

```
ValueError: scoring must return a number, got dask.array<mul, shape=(), dtype=float64, chunksize=()> (<class 'dask.array.core.Array'>) instead. (scorer=score)
```

which comes from https://github.com/scikit-learn/scikit-learn/blob/30ebb98dff5a7773ac66d068c8103cd47c0427ec/sklearn/model_selection/_validation.py#L606-L609

TomAugspurger:
@jrbourbeau is this still a WIP?

TomAugspurger:

jrbourbeau:
Thanks @TomAugspurger. I might have some local changes (additional tests) I still want to push before merging. I'll check and let you know if this is the case

jrbourbeau:
It appears that dask/dask#3767 made some changes that are causing a test to fail here. I'll work on a fix

PR #283, opened 2018-07-05
Avoid calling compute prior to assert_eq in tests

mrocklin:
This changes some tests so that assert_eq accepts dask objects rather
than numpy ones.  This allows us to perform a variety of useful checks
like `x.dtype == x.compute().dtype`.


mrocklin:
Merging this in four hours if there are no comments.  This fixes a break in master's tests.

cc @TomAugspurger

mrocklin:
Thanks @TomAugspurger

Issue #193, opened 2018-06-04
Standardize logging


TomAugspurger:
1. Provide a contextmanager for "timed" log statements.
2. Consistent naming
3. Consistency in what is logged

jakirkham:
Would it be worth considering capturing and forwarding logging information from things like `scikit-learn` wrapped algorithms? Just thinking of our recent experience watching this information getting dropped. 😉

mrocklin:
I'm inclined to use the logging module and let downstream users decide what
logging streams they want to capture.

TomAugspurger:
The issue in that case was that scikit-learn just prints to stdout. There's
a stalled PR moving things to logging:
https://github.com/scikit-learn/scikit-learn/pull/6930.

In the meantime, users can do something like
https://www.electricmonk.nl/log/2011/08/14/redirect-stdout-and-stderr-to-a-logger-in-python/
to redirect the print statements, but we shouldn't do that as a library.

jakirkham:
Interesting, thanks for the reference. Gave that PR a nudge.

On a different point, it's probably worthwhile for Distributed to start capturing printed info on workers and start logging. Currently that info is just lost, which seems worse. Opened issue ( https://github.com/dask/distributed/issues/2033 ) on this point.

TomAugspurger:
As an example of the kind of changes I had in mind, see https://github.com/dask/dask-ml/pull/228/files#diff-86d9909fd7ff2d87b6a68dba02a52da5R217

and their use in https://github.com/dask/dask-ml/pull/228/files#diff-b551222a46da15072e0a51ca86cdae11R166

For timing things, I think a context manager like


```python
import contextlib
from timeit import default_timer

@contextlib.contextmanager
def _timed(name, *args, **kwargs):
    start = default_timer()
    logger.info("Starting %s", name)
    yield
    stop = defult_timer()
    logger.info("Finished %s in %s, name, stop - start)  # nicer formatting for time.
```

would be useful

gglanzani:
@TomAugspurger Were you thinking about something like [this](https://github.com/dask/dask-ml/pull/286)?

PR #286, opened 2018-07-07
Add utilities functions to time functions

gglanzani:
Reference #193

### DIFF
--- a/dask_ml/utils.py
+++ b/dask_ml/utils.py
@@ -1,3 +1,8 @@
+import contextlib
+import datetime
+import functools
+import logging
+from timeit import default_timer as tic
 from collections import Sequence
 from multiprocessing import cpu_count
 from numbers import Integral
@@ -12,6 +17,8 @@ from dask.dataframe.utils import assert_eq as assert_eq_df
 
 import sklearn.utils.extmath as skm
 import sklearn.utils.validation as sk_validation
+
+logger = logging.getLogger()
 
 
 def svd_flip(u, v):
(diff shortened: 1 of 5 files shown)


dask_ml/utils.py
```diff
@@ -241,6 +248,44 @@ def _format_bytes(n):
     return '%d B' % n
 
 
+@contextlib.contextmanager
+def _timer(name, *args, _logger=None, level='info', **kwargs):
+    """
+    Output execution time of a function to the given logger level
+
+    :param str name: How to name the timer (will be in the logs)
```
TomAugspurger: We use [numpydoc](http://numpydoc.readthedocs.io/en/latest/format.html) style docstrings. Could you update these to use a `Parameters` section (don't need ad `Returns`).

dask_ml/utils.py
```diff
@@ -241,6 +248,44 @@ def _format_bytes(n):
...
+    :return: None
+    """
+    start = time.perf_counter()
+    _logger =  _logger or logger
+    _logger.info("Starting %s", name)
+    yield
+    stop = time.perf_counter()
+    delta = datetime.timedelta(seconds=stop - start)
+    _logger_level = getattr(_logger, level)
+    _logger_level("Finished %s in %s", name, delta)  # nicer formatting for time.
```
TomAugspurger: How is this formatted for a timedelta?

gglanzani:
```
>>> import logging
>>> import datetime
>>> logging.warning("%s", datetime.timedelta(seconds=50))
WARNING:root:0:00:50
>>> logging.warning("%s", datetime.timedelta(seconds=500))
WARNING:root:0:08:20
>>> logging.warning("%s", datetime.timedelta(seconds=5000))
WARNING:root:1:23:20
```

gglanzani:
@TomAugspurger There are usages of `_timed` here and there. I've tried to remove the other usages as well (note that in some cases the time is given back, so there's no way to make the context manager return without some magic — but that would defeat the point).

TomAugspurger:

The usages of `timed` look good. I'll take a closer look tomorrow.

gglanzani:
@TomAugspurger I had to remove `_timed` from the methods.

The issue is that `functools.wraps` doesn't play nicely with methods.

There are three ways out:

- Not use the decorator for the method (chosen here);
- Not use `functools.wraps`, and lose the docstring;
- Use the `wrapt` package, which considerably slows the code down.

Let me know if you're ok with the trade-offs.

dask_ml/utils.py
```diff
@@ -246,6 +253,53 @@ def _format_bytes(n):
...
+    Parameters
+    ----------
+    name : str
+        How to name the timer (will be in the logs)
+    logger : logging.logger
+        The optional logger where to write
+    level : str
+        On which level to log the performance measurement
+    """
+    start = time.perf_counter()
```
TomAugspurger: Looks like `time.perf_counter` is new in python 3.3, and we support 2.7.

gglanzani: Yeah, I fixed that. I'm now checking (locally) why the `check_estimator` test is so slow.

TomAugspurger: I've noticed some strangeness with that test previously. If necessary, we can disable it, but I'm a bit concerned that it's showing this change has a performance impact (though I don't see why that would be).

gglanzani: @tomaugspurger I'm running it locally with the decorator turned off, but I don't see different results.

I'll test a bit more and report back.

gglanzani: @TomAugspurger I forgot an indentation in the Lloyd loop :) Fixed now

gglanzani:
@TomAugspurger I've rebased against master.

Local build is green, let's see what py27 holds for me.

dask_ml/utils.py
TomAugspurger: It's entirely possible that was my fault when I did the reformatting. If so then you have my apologies :)

TomAugspurger:
All green, thanks @gglanzani1

PR #288, opened 2018-07-08
Incremental model selection

mrocklin:
This implements a simple approach to do hyper-parameter optimization on
incrementally trained models.  It is heavily insprired by (and steals content
from) #221 but makes fewer opinions, and doesn't attempt to implement 
scikit-learn APIs.  I don't intend for this to be user-facing, or necessarily get 
merged, but I hope that it to enable conversation around the topic.

Currently it implements a function ``fit``, which takes in a model, training
and testing data and a parameter space.  It creates many models and trains them
on a block of data, scores them, and then keeps only the better performing
ones, training them on subsequent blocks of data.  
It continues to do this until only one is selected.

This function is not intended for public users, but is hopefully something we
can play with as we eventually build a full solution.  It is decently
configurable today and we will need to find sane defaults.  Notably, this
function does not take opinions on the following:

1.  How testing data should be separated from training data,
    currently they are both explicitly provided
2.  How many parameters to start with (provided as integer input)
3.  How to decrease this number over time (provided as function input)

Additionally it does not, but probably should provide configurations for the
following:

1.  How to select models from the current group
2.  How to decide on stopping criteria for any particular model
3.  How to decide how many models to select based on the scores received so far

I believe that this implementation is a decent start though and should be ok as a
conversation piece.  I think that it makes some decisions that we don't want to
think about yet (online determination of how many models to keep) while giving
us knobs for other parameters that I think we'll want to play with (model decay
functions, how to handle testing, etc..)

It is decently well tuned from a dask perspsective.  
It optimistically tries moving forward on models between rounds and then cancels
work if it's not necessary.   This depends on dask master.
Performance is also a bit better with https://github.com/dask/distributed/pull/2100 
merged in (but not necessary)

Notebook: https://gist.github.com/3a33bed7e7289550bfea8a48af702c82

It would be nice to settle on a sane default for this operation, which I understand is commonly needed.
I suspect that we should play a bit with some real problems using this to see what ends up being important before doing much more development work.


dask_ml/model_selection/_incremental.py
```diff
@@ -0,0 +1,288 @@
...
+            # This is a hack to keep the futures in the scheduler but not in memory
+            X_token = client.submit(len, X_future)
+            y_token = client.submit(len, y_future)
+            tokens[time_step] = (X_token, y_token)
+
+            return X_future, y_future
+
+        else:
+            x_key, y_key = seen[j]
+            return Future(x_key), Future(y_key)
```
mrocklin: I don't recommend looking too deeply at this function.  It's unclean and likely to go in the future.

dask_ml/model_selection/_incremental.py
```diff
@@ -0,0 +1,288 @@
...
+        history.append(meta)
+
+        # Evolve the model by a few time steps, then call score on the last one
+        model = models[ident]
+        for i in range(time_step, next_time_step):
+            X_future, y_future = get_futures(i + 1)
+            model = client.submit(_partial_fit, model, X_future, y_future,
+                                  fit_params, priority=-i + meta['score'])
+        score = client.submit(_score, model, X_test, y_test, scorer,
+                              priority=-time_step + meta['score'])
```
stsievert: Nice – the priority keyword is useful, and I wasn't aware of the implementation.

I should spend some time thinking about priority. I think we want to use a multiplicative ratio to try to remove scale: `priority=score / i`.

mrocklin:
I recommend that before we develop further on either solution that we play
a bit with concrete problems to see what is important.

Eventually my guess is that this implementation will grow to the point
where you can use it as a piece to build out the full algorithm that you
want without having to do any low-level dask work.  However, I'm going to
claim that we don't yet know what algorithm you'll want to build before
first having a few more serious examples to play with.

mrocklin:
After looking at the `Incremental.fit` problem I'm inclined to generalize this code a bit so that the caller provides a list of models, training and testing data, and a function that gets called after every batch of training data gets processed to determine how to proceed.  That function would accept a dictionary mapping model identity to scoring information, and it would then return a dictionary showing how much further to proceed on each model before checking in again.

```python
def update(score_info):
    """ Something like successive halving for hyperband """
    best = topk(n=len(score_info) / 2, score_info, lambda k: score_info[k]['scores'][-1])
    additional_steps = int(...)
    return {ident: scores_info[ident]['time-step'] + additional_steps for ident in best}

def update(score_info):
    """ Something like Incremental.fit """
    assert len(score_info) == 1
    [d] = score_info.values()
    if any(score > d['scores'][-10] + tolerance for score in d['scores'][-10:]):
        return {ident: d['time-step'] + 1}  # do one more time step
    else:
        return {ident: d['time-step']}  # no more work to do, signals termination

def update(score_info):
    """ Something like RandomSearchCV """
    out = {}
    for ident, d in score_info:
        if any(score > d['scores'][-10] + tolerance for score in d['scores'][-10:]):
            out[ident] = d['time-step'] + 1  # keep evolving this one
        else:
            out[ident] = d['time-step']  # don't evolve this one any further, it's done
```

The full fit function would have the something like the following inputs:

```python
def fit(model, parameter_list, X_train, y_train, X_test, y_test, update, fit_params, scorer):
    ...
```

This may not cover these cases though.  Critique welcome before I invest time into this.

mrocklin:
I simplified the example above significantly in an edit.  Better to look on github rather than depend on e-mail.

stsievert:
> I'm inclined to generalize this code a bit

This looks like a disciplined way of controlling when to call `partial_fit` on models. It looks like one float determines model quality. `update` controls whether to pass a particular model to `fit`.

I think this looks good to me. I would like to store state inside `update` (I assume `update` is called with `as_completed` as `fit` completes). For the output of `update`, I think a negative integer should cancel any `fit` calls that are currently running for that model.

I think we'd want to control the `fit` call signature. Could we have a list of metadata aside the scores? Something like

``` python
>>> scores
{'model-1': [0.3, 0.4, 0.5],
 'model-2': [0.5, 0.5, 0.5]}
>>> model_meta
{'model-1': {...},
 'model-2': {...}}
>>> all_meta  # for HyperbandCV
{'scorer': ...,
 'X_test': ...,
 'y_test': ...}
```

where `all_meta` and `model_meta` could be configured by the programmer of Hyperband/`Incremental.fit`/RandomizedSearchCV in some way.

Then the call signature for `fit` could be

``` 
fit(model, X, y, **model_meta[ident], **all_meta) -> model, meta
```

I'm not seeing why `update` should be an argument to `fit`, but https://github.com/scikit-learn/scikit-learn/pull/9599#issuecomment-399853508 seems related. I renamed `{X, y}_train` to `X, y`. I also think we should rename `update`.

mrocklin:
> It looks like one float determines model quality

I don't understand this.  Which float?  In principle there is no one thing that determines model quality here.  We're a bit lower level.

> I would like to store state inside update.

What state would you want to propagate.  I expect that the `scores` input will grow a number of other bits of metadata collected during the computation.

>   (I assume update is called with as_completed as fit completes)

In the context of successive halving update is called at the end of every epoch

> I think a negative integer should cancel any fit calls that are currently running for that model.

The convention I proposed above is that the user passes back the number of additional steps go to forward in each model before calling update again.  A number of `0` would result in no additional fits.  Removing the element from the dict would result in forgetting the model entirely.  I believe that your proposed use of `-1` fits one of these two.  Is that correct?

> I think we'd want to control the fit call signature. Could we have a list of metadata aside the scores? 

Yes, I agree that some collection of metadata is likely to be important.

> where all_meta and model_meta could be configured by the programmer of Hyperband/Incremental.fit/RandomizedSearchCV in some way.

So the user would also pass functions to include in the `partial_fit` or `score` calls?  Is there a particular place where they should be run?  Should the user have the freedom to provide the entire `partial_fit` or `score` call or is this too much?  If too much, then can you suggest an implementation of `_partial_fit` that would cover most needs that you can think of?  Where do we call meta_func?  Or should it be a context manager or something else?  What is the right balance of easy to implement and also sufficient?  (these are somewhat rhetorical questions, I wouldn't expect anyone to have a full answer to any of them without trying out a bunch of things).

```python
def _partial_fit(model_and_meta, X, y, fit_params, meta_func=None):
    ...

def _score(model_and_meta, X, y, scorer, meta_func=None):
    ...
```

> I'm not seeing why update should be an argument to fit, but scikit-learn/scikit-learn#9599 (comment) seems related. 

I'm probably misusing the term fit here.  That's just the name of my full function that does all of the training work.  I'm using the term `partial_fit` for training on a single batch.  I suspect that this is what you are calling fit above.

stsievert:
I got confused by naming/etc but am starting to understand (and disregard most of https://github.com/dask/dask-ml/pull/288#issuecomment-405416238; it's ill-informed). I think this adaptive training will require

* specifying how many times `_partial_fit` is called with a particular model
* receiving the score for each model after `_partial_fit` called so many times
* storage and modification of metadata. This metadata is per-model for Hyperband, but it might be global for all models in other use cases (e.g., bayesian searches).

I think the `_partial_fit` and `_score` functions are good in this implementation. I think `update` should have the call signature 

```
update(ident: str, train_metas: {str: dict}, model_metas: {str: dict}) 
    -> training_iters: {str: int}, model_metas: {str: dict}
```

* There are two pieces of metadata: one for training (`train_metas`), and one to be used and modified in `update` (`model_metas`)
* I would expect this to be called as soon as model `ident` has finished their `_score` call.
* The metadata information in `model_metas` might change

I think `train_metas` should have `scores` and `partial_fit_calls` keys, and also the same keys as `cv_results_` (e.g., `mean_fit_time`). This would be the object passed to and modified in `_partial_fit` and `_score`.

I think `model_metas` should hold on to historical information, and be initialized in `meta_func`. In the Hyperband case, this would mean holding onto the previous brackets and their scores in case any models need to be restarted. This would only live on the scheduler/master and not be passed to or modified in `_partial_fit` or `_score`.

I do think `update` should be renamed.


> The convention I proposed ... Removing the element from the dict would result in forgetting the model entirely

I think 'not present' should default to 'do nothing'. I'd only want to specify the state for some subset of the models, not all the models. For Hyperband with some subset of the models I'd specify which models should be trained more and which models should be killed.

I think there `{'model-0': -1}` should cancel the most recent `_partial_fit` call for `'model-0'`, but hold onto the results from the previous `_partial_fit` calls. I think we should use `-1` because it's kind of like going back one `_partial_fit` call.

mrocklin:
I'm not sure about some of the motivation behind your thoughts above.  

For the desire to call `update` after every score call I believe that you want this in order to support asynchronous hyperband, where you evolve tasks based on the scores that have arrived so far.  Instead of this approach I recommend that we just submit a training step on every model until a global `update` check, but submit them with a priority equal to their score.  This will mean that we still keep things saturated during stragglers, but won't let good models die if they happen to be unlucky about when they finish.  If async hyperband was your only motivation then I suggest that we keep things a bit simpler for now with the `update` function and run it only after all of the previous instructions finish.

Similarly I suspect that this was your motivation for having "no change" be the default in update.  But I'm not sure.  If you can include motivation behind your thoughts that would be useful

stsievert:
> If you can include motivation behind your thoughts that would be useful

Asynchronous Hyperband is part of my motivation but not all of it. This is an adaptive training function, one that decides to train models based on prior evaluations and stored metadata. Wikipedia has [a good example][1] of the model updating after seeing one evaluation, what I'm used to seeing in adaptive algorithms.

> run it [update] only after all of the previous instructions finish.

This won't perform well for synchronous Hyperband. There are a couple embarrassingly parallel loops in synchronous Hyperband, and there's no reason for loop to wait on another. It will work well for one successive halving call in synchronous Hyperband.

This is important: the initial call to `update` would return anywhere between 1 and 81 calls to `_partial_fit`. This doesn't really go away in later calls to `update`, which will specify anywhere between 3 and 27 `_partial_fit` calls.

This could be avoided if there's some way to specify the different Hyperband loops as independent, at least for synchronous Hyperband. This could work for asynchronous Hyperband too, as long as additional arguments are passed.

> run it [update] only after all of the previous instructions finish.

I think this should be supported as well. That's the simplest interface I would expect.

[1]:https://en.wikipedia.org/wiki/Online_machine_learning#Online_convex_optimisation

mrocklin:
> This is an adaptive training function, one that decides to train models based on prior evaluations and stored metadata. Wikipedia has a good example of the model updating after seeing one evaluation, what I'm used to seeing in adaptive algorithms

Understood.  And in principle I agree that we might gain from making decisions whenever we find new information.  To start out though with my current constraints (finding solutions to the above problems) I seem to be able to implement most things about as efficiently as desired.  If we choose to get fancier we can get fancier easily.  This seems to suffice for now though? 

> This won't perform well for synchronous Hyperband. There are a couple embarrassingly parallel loops in synchronous Hyperband, and there's no reason for loop to wait on another. It will work well for one successive halving call in synchronous Hyperband.

How would you interleave calls?  The paper specifies a nested loop.  That would be trivial to implement here.  We can definitely launch the various successive-halving computations concurrently if desired.  Or perhaps you're referring to some way to fuse and share computations between successive halving computations?

> The initial call to update would return anywhere between 1 and 81 calls to _partial_fit

Just checking, 81 is just an example number here, right?  One might ask for thousands I think?  I believe that 81 is just a number they use in a table to show an example of how things might evolve, and what they use in their LeNet benchmark.  I don't think it's a magic number, please let me know if I'm missing something.

> This could be avoided if there's some way to specify the different Hyperband loops as independent, at least for synchronous Hyperband. This could work for asynchronous Hyperband too, as long as additional arguments are passed.

It's definitely possible to run multiple of these at the same time, though they won't be sharing work.  I can imagine that that would be helpful if we have abundant parallelism, which seems possible.

To make this concrete, something like the following would work:

```python
@gen.coroutine
def hyperband(...):
    yield [successive_halving(...) for _ in ...]
```

The Dask client will then run this process many times from the same thread.

stsievert:
> 81 is just an example number here

Correct. Sorry for the confusion. `maxiter=81` and that can be customized.

> It's definitely possible to run multiple of these at the same time,

Then for synchronous Hyperband I am satisfied. My other concerns go out the window. 

There might be something more with tpot or Bayesian sampling, but I'll look at the code first.

>  If we choose to get fancier we can get fancier easily.

Could you push your code? I take it the notebook was run with modifications to `_fit`. I'd like to look that code over.

mrocklin:
I recommend the graph view of the dashboard when executing.

My current plan is to leave this as-is until some sort of case study arrives to benchmark it against.  I'm inclined to experiment a bit with successive_halving to see how different decay rates behave and then use that as motivation for hyperband after we have some practical experience.  Presumably UI bits would come after that.

TomAugspurger:

Will take a closer look later (tomorrow most likely).

stsievert:
This looks pretty good. A lot of my concerns are addressed by the fact that this  

* works with classes (e.g., `update=cls.fit`), so arbitrary state can be stored and arbitrary functions can be run.
* can be called in parallel (though I had some trouble getting https://github.com/dask/dask-ml/pull/288#issuecomment-405756267 to work). 

Making this work with asynchronous Hyperband will be some work but not an incredible amount.

mrocklin:
> Making this work with asynchronous Hyperband will be some work but not an incredible amount.

Yeah, I'm happy to set up concurrent execution of successive halving for hyperband (or walk someone else through it) after we have a bit more understaning of how randomsearch and successive halving perform.  I don't think that building hyperband is likely to be a priority until then.

dask_ml/model_selection/_incremental.py
```diff
@@ -0,0 +1,277 @@
...
+        history.append(meta)
+
+        # Evolve the model one more step
+        model = models[ident]
+        X_future, y_future = get_futures(meta["time_step"] + 1)
+        model = client.submit(_partial_fit, model, X_future, y_future, fit_params)
+        speculative[ident] = model
+
+        # Have we finished a full set of models?
+        if len(speculative) == number_to_complete:
```
stsievert: This assumes that `update` returns a bool, not an int indicating how many times to call `partial_fit`.

(a model is added to `speculative` as soon as it completes one `_partial_fit` call)

dask_ml/model_selection/_incremental.py
```diff
@@ -0,0 +1,277 @@
...
+        if future.cancelled():
+            continue
+        ident = meta["ident"]
+
+        info[ident].append(meta)
+        history.append(meta)
+
+        # Evolve the model one more step
+        model = models[ident]
+        X_future, y_future = get_futures(meta["time_step"] + 1)
```
stsievert: I think we should replace `meta['time_step'] + 1` with `np.random.choice(X.blocks)` instead.

I think data access should be random, both across models and across time.

stsievert: Never mind about random access. `get_futures` manages that.

dask_ml/model_selection/_incremental.py
mrocklin: > This assumes that update returns a bool

I don't think that that's true.  The attached notebook shows using update functions that return dictionaries mapping model ID's to non-negative integers.  The successive halving example is a good example of this.

stsievert: > (a model is added to speculative as soon as it completes one _partial_fit call)

Never mind. That's not the case. I didn't read the code carefully enough.

stsievert:
I thought I discovered some bugs, but turns out I didn't.

stsievert:
I've run into a bug where Dask can serialize and deserialize a model but the `fit` method tries to pickle the object (not Dask serialize) and it fails. Here's a notebook demonstrating this bug: https://gist.github.com/stsievert/dd65c1071593b8842267ac7a1f92a22e (this bug is on an outdated branch, https://github.com/mrocklin/dask-ml/pull/1, but it doesn't look like the commits since d8ab06aa35f4f2b73968e7726fe214cd791c6500 will resolve this issue).

After this issue is resolved, I will run into https://github.com/dask/distributed/issues/2183.

mrocklin:
Sorry for the long delay.  Short term I recommend seeing if you can pre-scatter the model.  My hope is that this will be a decent short term fix.

    model = client.scatter(model)

I believe that this is a case of https://github.com/dask/distributed/issues/2110

stsievert:
I ran into a bug on this branch, and implemented a quick and simple fix.

The algorithm in would try to retrieve a particular model from `models`, but it would throw a key error (which has to do with cancelling work).  Occasionally, the algorithm in https://github.com/dask/dask-ml/pull/221 will cancel work ahead of time and remove it from memory (`del models[desired_ident]`), even though it wanted to retrieve that key later.

stsievert:
Basically, I'd like a way to restore a model after it's been deleted. If the `del models[ident]` line isn't removed, something like

``` python
while True:  # line 168 of _incremental.py
    # ...
    instructions = update(info)
    bad = set(models) - set(instructions)

    # Delete the futures of bad models.  This cancels speculative tasks
    for ident in bad:
        del models[ident]
        del scores[ident]
        del info[ident]

    if not any(instructions.values()):
        for ident in instructions:
            models[ident] = restore_model(ident)
        break
```

mrocklin:
> Basically, I'd like a way to restore a model after it's been deleted. If the del models[ident] line isn't removed, something like

You would continue keeping ident in the dictionary that you provide in the update function, but with a value of `0`.  This will keep it around, but not evolve it further.

stsievert:
> You would continue keeping ident in the dictionary that you provide in the update function, but with a value of 0. This will keep it around, but not evolve it further.

D'oh. Of course. I've pushed those changes.

stsievert:
Two questions:

1. Should we test these functions if they're not user facing?
2. Can I merge mrocklin#2 into #221? It'd simplify git and save some time.

TomAugspurger:
If you merge dask-ml master, I believe the CI failure will be fixed.

I haven't isolated the cause of that segfault yet.

TomAugspurger:
@mrocklin I'll need to hear your input on all these linting checks :)

I've found that https://dask-ml.readthedocs.io/en/latest/contributing.html#pre-commit-hooks is a decent workflow.

stsievert:
In this, it'd be nice to support keyboard interrupts in this, and preserve the models/infos/etc on exit.

My use case is a user firing a keyboard interrupt and still recovering the best model's score/related info.

mrocklin:
Agreed.  That's a nice idea.  I'll take a look at that soonish.

mrocklin:
"soonish" meaning in the next few days

mrocklin:
@TomAugspurger do you have thoughts on the right way to make a RandomSearch version of this that is API-familiar to a Scikit-Learn user?  Is there even an established API for this kind of thing with `partial_fit`?

dask_ml/model_selection/_incremental.py
```diff
@@ -0,0 +1,280 @@
...
+        models.update(_models2)
+        scores.update(_scores2)
+        speculative = _specs2
+
+        new_scores = list(_scores2.values())
+
+    raise gen.Return((info, models, history))
+
+
+def fit(*args, **kwargs):
```
TomAugspurger: Would be best to make this the actual signature.

dask_ml/model_selection/_incremental.py
```diff
@@ -0,0 +1,280 @@
...
+    model : Estimator
+    params : dict
+        parameter grid to be given to ParameterSampler
+    X_train : dask Array
+    y_train : dask Array
+    X_test : Array
+        Numpy array or small dask array.  Should fit in memory.
+    y_test : Array
+        Numpy array or small dask array.  Should fit in memory.
+    start : int
```
TomAugspurger: removed I think.

Add additional_calls.

dask_ml/model_selection/_incremental.py
```diff
@@ -0,0 +1,280 @@
...
+    >>> model = SGDClassifier(tol=1e-3, penalty='elasticnet')
+    >>> params = {'alpha': np.logspace(-2, 1, num=1000),
+    ...           'l1_ratio': np.linspace(0, 1, num=1000),
+    ...           'average': [True, False]}
+
+    >>> X_test, y_test = X[:100000], y[:100000]
+    >>> X_train = X[100000:]
+    >>> y_train = y[100000:]
+
+    >>> info, model, history = yield fit(model, params,
```
TomAugspurger: Need to update.

mrocklin:
> class IncrementalSearch(BaseEstimator):

Would it make sense to factor this out from the Hyperband work going on now?

TomAugspurger:
Yes, I think so. But I'm not sure what order we should do things in.

- IncrementalSearch (is that the best name?) that provides a kind of baseline for this class of search models
- Hyperband is a relatively minimal extension to IncrementalSearch implementing its strategies for model selection, etc.

does that conform with others' expectations? I'm happy to take over the work from those that are busy with other things.

mrocklin:
I've updated the docstring considerably, including a runnable example.  Tests will fail until https://github.com/dask/distributed/pull/2232 is merged in.

mrocklin:
> Yes, I think so. But I'm not sure what order we should do things in.

> - IncrementalSearch (is that the best name?) that provides a kind of baseline for this class of search models
> - Hyperband is a relatively minimal extension to IncrementalSearch implementing its strategies for model selection, etc.

> does that conform with others' expectations? I'm happy to take over the work from those that are busy with other things.

Yes, I think that ideally something like Hyperband would look something like the following:

```python
class Hyperband(IncrementalSearch):
    def __init__(**hyperband_specific_kwargs);
        ...

    def additional_calls(self, scores):
        ...
```

If this existed I think that it would be easier to experiment with this sort of thing.  

I should write up documentation for this PR, but I might want to wait until such a super-class exists.

mrocklin:
I would be happy to leave the sklearn-specific conventions to someone else :)

TomAugspurger:
OK then, so how's this for a plan

1. Merge dask/distributed#2232
2. Merge this
3. I take a pass at an IncrementalSearch base class today
4. We rebase Hyperband on IncrementalSearch

And followup items for documentation, examples, etc. of IncrementalSeach

The goal being to maximize the effectiveness of @stsievert's time :)

mrocklin:
Sounds good to me

dask_ml/model_selection/_incremental.py
```diff
@@ -218,10 +219,23 @@ def get_futures(partial_fit_calls):
 
         new_scores = list(_scores2.values())
 
+    models = {k: client.submit(operator.getitem, v, 0) for k, v in models.items()}
```
mrocklin: This will break the current Hyperband implementation BTW

stsievert:
> class IncrementalSearch(BaseEstimator):

Perfect – I had this idea when playing with `stop_on_plateau`, but didn't bring it up. This should also inherit from `DaskBaseSearchCV`, right?

TomAugspurger:
That's my current plan. Should have something later this afternoon.

mrocklin:
Good to merge any time @TomAugspurger .  I'll leave this to you

dask_ml/model_selection/_incremental.py
```diff
@@ -0,0 +1,349 @@
...
+
+    # assume everything in fit_params is small and make it concrete
+    fit_params = yield client.compute(fit_params)
+
+    # Convert testing data into a single element on the cluster
+    # This assumes that it fits into memory on a single worker
+    if isinstance(X_test, da.Array):
+        X_test = client.compute(X_test)
+    else:
+        y_test = yield client.scatter(y_test)
```
TomAugspurger: Should this be `X_test = yield client.scatter(X_test)`?

mrocklin: Yes, it probably should.  My mistake.

PR #292, opened 2018-07-10
Use our scorers for default score methods in Incremental

### LINKED ISSUE #200: How to handle scoring for meta-estimators

TomAugspurger:
Closes #200

### DIFF
--- a/ci/environment-2.7.yml
+++ b/ci/environment-2.7.yml
@@ -18,6 +18,7 @@ dependencies:
   - psutil
   - pytest
   - pytest-cov
+  - pytest-mock
   - python=2.7
   - sortedcontainers
   - scikit-learn
--- a/ci/environment-3.6.yml
+++ b/ci/environment-3.6.yml
@@ -18,6 +18,7 @@ dependencies:
   - psutil
   - pytest
   - pytest-cov
+  - pytest-mock
   - python=3.6
   - sortedcontainers
   - scikit-learn
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -190,12 +190,24 @@ class ParallelPostFit(sklearn.base.BaseEstimator):
         score : float
                 return self.estimator.score(X, y)
         """
-        if self.scoring:
+        scoring = self.scoring
+
+        if not scoring:
+            if (type(self._postfit_estimator).score is
+                    sklearn.base.RegressorMixin.score):
+                scoring = 'r2'
+            elif (type(self._postfit_estimator).score is
+                    sklearn.base.ClassifierMixin.score):
+                scoring = 'accuracy'
+        else:
+            scoring = self.scoring
+
+        if scoring:
             if (not dask.is_dask_collection(X) and
                     not dask.is_dask_collection(y)):
-                scorer = sklearn.metrics.get_scorer(self.scoring)
+                scorer = sklearn.metrics.get_scorer(scoring)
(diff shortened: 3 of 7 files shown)


TomAugspurger:
Apparently,

```python
In [7]: type(SGDClassifier()).score is sklearn.base.ClassifierMixin.score
```

is True for Py3, and False for Py2. Changing the `is` to `==` seems to work on both, but I'm not sure `__eq__` means on functions...

TomAugspurger:
Tests pass now. Will merge tomorrow if there are no comments (a "wait till after SciPy" comment is perfectly reasonable).

Speaking of tests... does anyone have thoughts on this approach? I use mock to ensure that *our* `get_scorer` is called.

Issue #289, opened 2018-07-09
Doc builds failing


mrocklin:
https://readthedocs.org/projects/dask-ml/builds/7455126/

> Command killed due to excessive memory consumption

PR #308, opened 2018-07-19
DOC: Smaller Incremental examples

TomAugspurger:
Closes https://github.com/dask/dask-ml/issues/289

In local testing with an RTD docker image, I noticed that the `gs.fit(X, y, classes[0, 1])`  line took a long time. This will hopefully fix things.

### DIFF
--- a/docs/source/incremental.rst
+++ b/docs/source/incremental.rst
@@ -48,7 +48,7 @@ between machines.
    X, y = make_classification(chunks=25)
    X
 
-   estimator = SGDClassifier(random_state=10, max_iter=1000)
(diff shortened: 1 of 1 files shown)


TomAugspurger:
RTD build: https://readthedocs.org/projects/dask-ml/builds/7513213/

TomAugspurger:
Finished in 700 s (max of 900). So we're still OK for now, but that's a little close to the RTD limit. We still have the option of using travis / doctr, as in https://github.com/dask/dask-ml/pull/253, but that'll require some work to update references to the docs and forward the current ones to the new URL, so let's hold off on that till we need it.

PR #310, opened 2018-07-19
ENH: LabelEncoder supports pandas Categorical

TomAugspurger:
Enhances LabelEncoder to use CategoricalDtype for pandas and dask
series.

This improves the performance, and will be helpful for implementing OneHotEncoder efficiently for dask dataframes.

```python
import string
import numpy as np
import pandas as pd
import dask.dataframe as dd
import dask_ml.preprocessing

dtype = pd.api.types.CategoricalDtype(list(string.ascii_letters[:12]))
n = 1_000_000

data = dd.from_pandas(
    pd.Series(np.random.choice(dtype.categories, size=n), dtype=dtype),
    npartitions=n // 100_000
)
```

Setup

```python
le = dask_ml.preprocessing.LabelEncoder()  # True / False
codes = le.fit_transform(data)

print('fit')
%timeit le.fit(data)
print('transform')
%timeit le.transform(data)

print('transform-compute')
%timeit le.transform(data).compute()

print('inverse_transform')
%timeit le.inverse_transform(codes)

print('inverse_transform-compute')
%timeit le.inverse_transform(codes).compute()
```

Results

method                    | categorical | no categorical
------                    | ----------- | --------------
fit                       | 43.5 µs     | 625 ms
transform                 | 678 µs      | 33.1 ms
transform-compute         | 4.98 ms     | 128 ms
inverse_transform         | 3.66 ms     | 240 µs
inverse_transform-compute | 37 ms       | 160 ms

cc @jrbourbeau

### DIFF
--- a/dask_ml/preprocessing/label.py
+++ b/dask_ml/preprocessing/label.py
@@ -4,6 +4,7 @@ from operator import getitem
 
 import dask.array as da
 import dask.dataframe as dd
+import pandas as pd
 import numpy as np
 
 from sklearn.preprocessing import label as sklabel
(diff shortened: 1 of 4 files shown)


TomAugspurger:
cc also @jorisvandenbossche if you're interested in what a future scikit-learn `LabelEncoder` that could use a pandas-like CategoricalDtype might look like.

jorisvandenbossche:
Cool! Note that in the mean time, we don't use LabelEncoder anymore directly in OneHotEncoder: https://github.com/scikit-learn/scikit-learn/pull/10209 (both now use a common `_encode` function). 
Although I suppose this doesn't matter too much for dask-ml (I don't think it would be possible to subclass OneHotEncoder to re-use the code that called LabelEncoder ?)

TomAugspurger:
> Note that in the mean time, we don't use LabelEncoder anymore directly in OneHotEncoder: scikit-learn/scikit-learn#10209 (both now use a common _encode function).

Thanks, I was working off an older branch. This will still be a nice standalone addition though, and I think it's worthwhile diverging from (getting ahead of?) scikit-learn here as using the dtype information is more important for large datasets.

PR #313, opened 2018-07-24
OneHotEncoder

TomAugspurger:
This adds a OneHotEncoder for encoding categorical data.

Differences from scikit-learn:

1. Uses categorical information from pandas (like our labelencoder)
2. `OneHotEncoder(sparse=True).transform(dask_array)` returns a dask array, rather than a sparse matrix. Each block of the dask array is a sparse matrix.

I've documented these differences in `preprocessing.rst`.

Currently, this doesn't handle unknown (previously unseen) values in the `.transform`, though I've just had an idea how to do it...


dask_ml/preprocessing/_encoders.py
```diff
@@ -0,0 +1,278 @@
...
+                inner_ars = [
+                    # TODO: dtype
+                    da.from_delayed(obj, (n_rows, len(categories)), dtype=self.dtype)
+                    for j, (obj, n_rows) in enumerate(zip(objs, x.chunks[0]))
+                ]
+                arrs.append(da.concatenate(inner_ars))
+                # TODO: this check fails. See why.
+                # da.utils._check_dsk(arrs[-1].dask)
+
+            X = da.concatenate(arrs, axis=1)
```
mrocklin: What stops us from using `X.map_blocks(construct, categories=categories) here?

TomAugspurger: And removing all of the to / from delayed, instead doing it all in construct (but not using delayed)?

I think that should be doable...

mrocklin: Yes.  My hope is that it would be a bit simpler.  There is also some cost to switching between collection types.  I think that every such boundary triggers an optimization step, which also stops us from fusing successive operations.

jakirkham:
Did you look at `argwhere` in Dask Array?

TomAugspurger:
I might move quickly on this and merge it without giving others time to properly review. It's a large change, but I'd like to immediately follow things up with `ColumnTransformer`, and then actually try it out on the criteo dataset.

TomAugspurger:
The coverage check failed because the runner reporting the coverage (py36) skips those tests, as they require scikit-learn dev.

PR #314, opened 2018-07-25
CI: Test that sort order is enforced.

TomAugspurger:
This should fail right now.

https://circleci.com/gh/dask/dask-ml/1531


TomAugspurger:
That last one failed because I didn't include isort in the env. https://circleci.com/gh/dask/dask-ml/1534 should fail for the *right* reason now :)

.circleci/config.yml
```diff
@@ -16,17 +16,27 @@ jobs:
       - run:
           command: |
             source activate ${ENV_NAME}
-            pytest --cov=dask_ml tests
+            mkdir test-reports
+            pytest --cov=dask_ml --junitxml=test-reports/junit.xml tests
```
TomAugspurger: This hooks into circle-ci's output summary. Failing tests will be extracted and put at the top of the output.

TomAugspurger:
isort seems finicky about the environment that runs it. I'm getting different outputs based on which environment is running the command, but with the same version of isort in each environment.

I've tried to pin down all the modules that gave me troubles in `setup.cfg`, but others may have different troubles. I'd recommend merging this once it passes, and then just disabling this check if it gives people any trouble.

TomAugspurger:
Going to merge this, but to other contributors *please* speak up if all these linting rules become obnoxious.

I've found that using pre-commit and this `.pre-commit-config.yaml` configuration helps reduce annoyances.

```python
repos:
-   repo: https://github.com/ambv/black
    rev: stable
    hooks:
    - id: black
      language_version: python3.6

-   repo: https://github.com/pre-commit/mirrors-isort
    rev: "f35773e46d096de5c45365f1a47eeeef36fc83ed"
    hooks:
    - id: isort
```

PR #316, opened 2018-07-25
DOC: note about preprocessing stateless sklearn transforms

### LINKED ISSUE #312: [request

stsievert:
Closes #312 with a documentation note.

I'd rather have more documentation than having Dask-ML be a drop in replacement for Scikit-learn.


stsievert:
I guess this should be closed in favor of an implementation similar to #122. Thoughts?

PR #329, opened 2018-07-31
[WIP] Updated examples

### LINKED ISSUE #168: Documentation Examples Updates
Our RTD build is currently failing, so I'm taking this chance to update a few things.

1. We have two "examples" systems: notebooks & sphinx-gallery. I think we should standardize on one (probably sphinx-gallery).
2. Examples like http://dask-ml.readthedocs.io/en/latest/auto_examples/plot_logistic_regression.html#sphx-glr-auto-examples-plot-logistic-regression-py aren't super useful. I'd propose we move this kind of "benchmark / scaling" examples to a new GH repo `dask/dask-ml-benchmarks`. Ideally the examples in this repository would show something novel to dask-ml, that isn't present in other libraries (like scikit-learn).
3. CI: ideally every example is runnable on the RTD instance, which limits them to small examples. We can show small examples on subsets of data and leave comments for how to expand it.

TomAugspurger:
This updates our examples. They've been simplified to

1. remove all extremely long running notebooks (all these finish in under a minute on my machine)
2. Indicate how the example could be expanded with a "Scale up:" comment.

I've also made some editorial changes

1. removed the dask-glm notebook. I haven't figured out how to scale it down so that it runs in a reasonable amount of time
2. Split the "machine-learning" notebook from dask-examples into two: distributed scikit-learn (CPU bound) and training on large datasets (RAM bound)

I'd like copy these over to `dask/dask-examples` as part of the release process.

I suspect that running these mean that we'll have to move away from readthedocs, which we're already often maxing out.

#253 uses Travis to build the docs. These would be uploaded to https://dask.github.io/dask-ml/. We could also presumably talk to people so that we could upload to `dask-ml.pydata.org`, which would mirror `dask.pydata.org` nicely.

So, two questions

1. Are people OK with moving the docs away from readthedocs
2. Does anyone have experience with redirecting readthedocs to a new domain?

Closes #168


mrocklin:
Random thoughts:

-  I like removing the long running ones, especially if they're part of a docs build process
-  +1 on splitting the standard machine learning notebook into two
-  I'm slightly concerned about copying content between dask-examples and dask-ml.  What do we do when people submit a PR to one or the other changing things?  I wonder if we might keep content in one repository but refer to it in the automated process of another.  For example dask-ml might run tests against dask-examples or dask-examples might pull in part of dask-ml during the docker build process
-  For the pydata.org domain we should also ask numfocus.  I wonder if, instead, there is some way to nest all of the various dask subprojects under dask.pydata.org like dask.pydata.org/ml and dask.pydata.org/kubernetes .  I'm not sure how to do this though.
-  Thoughts on removing or heavily simplifying the tensorflow notebook (unrelated to this PR though)

TomAugspurger:
> I'm slightly concerned about copying content between dask-examples and dask-ml.

Likewise. I just pushed a possible solution to this. It's hacky, but seems to work. We just clone or update the dask-examples as part of the doc build setup, and copy the `dask-examples/machine-learning` folder over to `source/examples`. I've removed the incremental and xgboost examples from this repository. I'll do followup PRs with the rest.

One question, do we want *all* the examples to go in dask-examples? I think the answer is no. We probably don't want to ever include something like tensorflow in the dask-examples image since it's so large.

>  I wonder if, instead, there is some way to nest all of the various dask subprojects under dask.pydata.org

I'm not sure either. Looking briefly, there's a request for sphinx to handle multiple project roots, but nothing implemented yet.

> Thoughts on removing or heavily simplifying the tensorflow notebook (unrelated to this PR though)

Yes :) Hoping to do that next week though.

If we're ok with this approach I'll finish migrating notebooks over to dask-examples/machine-learning.

TomAugspurger:
https://github.com/dask/dask-examples/pull/27 for porting several notebooks to dask-examples.

After that we would just have tensorflow.ipynb in this repo. I'm OK with that for now, and I'll revisit it when I get a chance.

mrocklin:
> I'm not sure either. Looking briefly, there's a request for sphinx to
handle multiple project roots, but nothing implemented yet.

This may happen outside of sphinx.  For example, my blog
matthewrocklin.com/blog is served entirely separately from my normal
website, matthewrocklin.com.  I suspect that there are domain handling
things we could do here instead.

TomAugspurger:
That would be nice. So ideally

- `dask.pydata.org/ml` would serve https://dask.github.io/dask-ml/ (serve may not be the correct term)
- `dask.pydata.org/distributed` would serve http://dask-kubernetes.readthedocs.io/en/latest/
- `dask.pydata.org/kubernetes` would serve http://dask-kubernetes.readthedocs.io/en/latest/

etc.

Do you know who from numfocus I should reach out to about this?

mrocklin:
I usually contact Andy Terrel or Ralf Gommers for website things

Issue #269, opened 2018-07-03
Add additional CV methods


TomAugspurger:
http://scikit-learn.org/stable/modules/classes.html#splitter-classes

I would prioritize KFold and StratifiedKFold.

PR #340, opened 2018-08-22
[MRG] Adds KFold

thomasjpfan:
Addresses some of #269 by adding `KFold`


TomAugspurger:
Looks good at a glance, thanks. While playing with this, I discovered an issue with `check_array` not handling 1-D X properly, which I can fix independently. Will take a closer look later.

TomAugspurger:
Sorry for the delay. This all seems good.

PR #344, opened 2018-08-28
Numba k-means

TomAugspurger:
This changes our one C-extension to use numba.

In microbenchmarks, the numba version is about 2x slower than the Cython version. This seems likely to do with numba not having a way to annotate that wraparound indexing is not used: https://github.com/numba/numba/issues/3205.

This will simplify distribution, and I suspect that we'll want to take advantage of numba in more places shortly.


TomAugspurger:
Any objections from you @mrocklin?

To expand a bit further, I initially wrote this branch when our [wheel builder](https://github.com/TomAugspurger/dask-ml-builder) infrastructure broke down during the last release.

I revived it last night after having to futz with getting compilers installed on remote machines when trying to get dask-ml installed on a cluster.

I also spent some time last night investigating pynndescent, and am hopeful we'll be able to use it for a lightweight approximate nearest neighbors algorithm, which would mean a Numba depedency.

mrocklin:
No objections from me on the Numba dependency.

PR #348, opened 2018-08-29
DOC/PKG: Prepare for 0.9 release [ci skip]

TomAugspurger:
[ci skip]


TomAugspurger:
Seems like there's some incompatibility between NumPy 1.15.1 and our numba jitted code. I'm unfortunately unable to reproduce this locally.

I've temporarily pinned NumPy 1.15.0 in our CI, will investigate more later.

PR #353, opened 2018-09-04
Test new circleci config

TomAugspurger:
Just seeing if this works.


TomAugspurger:
(narrator: it doesn't)

I should blacklist gh-pages.

PR #356, opened 2018-09-06
Base class for estimators using _incremental.fit

TomAugspurger:
Here's a rough draft of what a base class calling `_incremental.fit` may look like. I've just been stealing pieces from https://github.com/dask/dask-ml/pull/221 and making it work for the random sampler with the `remove_worst` strategy.

Looking at #221 there were a few pieces that will be common to each subclass

* data checking
* CV splitting
* building cv_results (partly)
* selecting the best model (partly)

That leaves a few things that must be done by each class

* Building the dict of `{model_id: (param_list, additional_calls)}`
* transforming the output of `fit` on each of those into a common form

In the end, I'm not sure how much code reuse we'll actually get, but I think it's worth exploring a bit further. A next step would be to make a branch of #221 that builds on top of this, and see how well the hyperband structure maps on to this base class. From there we can decide whether the base class can be adjusted to fit hyperband's needs, or whether it should just be scrapped. I'll try to get to that today.

cc @stsievert

### DIFF
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -547,10 +547,21 @@ class BaseIncrementalSearchCV(DaskBaseSearchCV):
         cv_results = self._get_cv_results(results)
         best_estimator, best_index = self._get_best(results, cv_results)
 
+        # Clean up models we're hanging onto
+        ids = [
+            (loop_id, model_id)
+            for loop_id in results
+            for model_id in results[loop_id][1]
+        ]
+        for (loop_id, model_id) in ids:
+            del results[loop_id][1][model_id]
+
         self.scoring_ = scorer
         self.cv_results_ = cv_results
         self.best_estimator_ = best_estimator
         self.best_index_ = best_index
+        self.n_splits_ = 1
+        self.multimetric_ = False  # TODO: is this always true?
         return self
 
 
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -519,21 +519,14 @@ class BaseIncrementalSearch(DaskBaseSearchCV):
         """
         return results
 
-    def fit(self, X, y, **fit_params):
-        """Find the best parameters for a particular model.
-
-        Parameters
-        ----------
-        X, y : array-like
-        **fit_params
-            Additional partial fit keyword arguments for the estimator.
-        """
+    @gen.coroutine
+    def _fit(self, X, y, **fit_params):
         X, y = self._check_array(X, y)
 
         X_train, X_test, y_train, y_test = self._get_train_test_split(X, y)
         scorer = check_scoring(self.estimator, scoring=self.scoring)
 
-        results = fit(
+        results = yield fit(
             self.estimator,
             self._get_params(),
             X_train,
@@ -548,6 +541,7 @@ class BaseIncrementalSearch(DaskBaseSearchCV):
         results = self._process_results(results)
         history_results = self._get_history_results(results)
         best_estimator, best_index = self._get_best(results, history_results)
+        best_estimator = yield best_estimator
 
         # Clean up models we're hanging onto
         ids = list(results.models)
@@ -561,6 +555,17 @@ class BaseIncrementalSearch(DaskBaseSearchCV):
         self.n_splits_ = 1
         self.multimetric_ = False  # TODO: is this always true?
         return self
+
+    def fit(self, X, y, **fit_params):
+        """Find the best parameters for a particular model.
+
+        Parameters
+        ----------
+        X, y : array-like
+        **fit_params
+            Additional partial fit keyword arguments for the estimator.
+        """
+        return default_client().sync(self._fit, X, y, **fit_params)
 
 
 class RandomizedWorstIncrementalSearch(BaseIncrementalSearch):
@@ -618,3 +623,67 @@ class RandomizedWorstIncrementalSearch(BaseIncrementalSearch):
 
     def _get_params(self):
         return ParameterSampler(self.parameters, self.n_iter, self.random_state)
+
+
+class RandomIncrementalSearch(BaseIncrementalSearch):
+    def __init__(
+        self,
+        estimator,
+        param_distribution,
+        n_iter=10,
+        max_iter=100,
+        patience=10,
+        tol=0.001,
+        test_size=0.15,
+        random_state=None,
+        scoring=None,
+        iid=True,
+        refit=True,
+        error_score="raise",
+        return_train_score=_RETURN_TRAIN_SCORE_DEFAULT,
+        scheduler=None,
+        n_jobs=-1,
+        cache_cv=True,
+    ):
+        self.max_iter = max_iter
+        self.n_iter = n_iter
+        self.patience = patience
+        self.tol = tol
+        super(RandomIncrementalSearch, self).__init__(
+            estimator,
+            param_distribution,
+            test_size,
+            random_state,
+            scoring,
+            iid,
+            refit,
+            error_score,
+            return_train_score,
+            scheduler,
+            n_jobs,
+            cache_cv,
+        )
+
+    def _get_params(self):
+        return ParameterSampler(self.parameters, self.n_iter)
+
+    def _additional_calls(self, info):
+        out = {}
+        max_iter = self.max_iter
+        patience = self.patience
+        tol = self.tol
+
+        for ident, records in info.items():
+            if max_iter is not None and len(records) > max_iter:
+                out[ident] = 0
+
+            elif len(records) > patience:
+                old = records[-patience]["score"]
+                if all(d["score"] < old + tol for d in records[-patience:]):
+                    out[ident] = 0
+                else:
+                    out[ident] = 1
+
+            else:
+                out[ident] = 1
+        return out
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -45,6 +45,28 @@ def _partial_fit(model_and_meta, X, y, fit_params):
         * info : Dict[model_id, List[Dict]]
             Keys are integers identifying each model. Values are a
             List of Dictk
+        * models : Dict[model_id, Future[Estimator]]
+            A dictionary with the same keys as `info`. The values
+            are futures to the fitted models.
+        * history : List[Dict]
+            The history of model fitting for each model. Each element
+            of the list is a dictionary with the following elements:
+
+            'model_id', 'params', 'partial_fit_calls', 'partial_fit_time', 'score', 'score_time'])
+
+            * model_id : int
+                A superset of the keys for `info` and `models`.
+            * params : Dict[str, Any]
+                Parameters this model was trained with. 
+            * partial_fit_calls : int
+                The number of *consecutive* partial fit calls at this stage in
+                this models training history.
+            * partial_fit_time : float
+                Time (in seconds) spent on this partial fit
+            * score : float
+                Score on the test set for the model at this point in history
+            * score_time : float
+                Time (in seconds) spent on this scoring.
     """
     with log_errors():
         start = time()
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -2,6 +2,7 @@ from __future__ import division
 
 import itertools
 import operator
+from collections import namedtuple
 from copy import deepcopy
 from time import time
 
@@ -21,6 +22,8 @@ from ..utils import check_array
 from ._search import _RETURN_TRAIN_SCORE_DEFAULT, DaskBaseSearchCV
 from ._split import train_test_split
 
+Results = namedtuple("Results", ["info", "models", "history"])
+
 
 def _partial_fit(model_and_meta, X, y, fit_params):
     """
@@ -36,10 +39,12 @@ def _partial_fit(model_and_meta, X, y, fit_params):
 
     Returns
     -------
-    model : Estimator
-        The model that has been fit.
-    meta : dict
-        A new dictionary with updated information.
+    Results
+        A named tuple with three fields: info, models, history
+
+        * info : Dict[model_id, List[Dict]]
+            Keys are integers identifying each model. Values are a
+            List of Dictk
     """
     with log_errors():
         start = time()
@@ -229,7 +234,7 @@ def _fit(
 
     models = {k: client.submit(operator.getitem, v, 0) for k, v in models.items()}
     yield wait(models)
-    raise gen.Return((info, models, history))
+    raise gen.Return(Results(info, models, history))
 
 
 def fit(
@@ -363,14 +368,12 @@ def fit(
 # ----------------------------------------------------------------------------
 
 
-class BaseIncrementalSearchCV(DaskBaseSearchCV):
+class BaseIncrementalSearch(DaskBaseSearchCV):
     """Base class for estimators using the incremental `fit`.
 
     Subclasses must implement the following abstract method
 
-    * _get_ids_and_args : Build the dict of
-       ``{model_id : (param_list, additional_calls)}``
-        that are eventually passed to `fit`
+    * _additional_calls
     """
 
     def __init__(
@@ -382,7 +385,6 @@ class BaseIncrementalSearchCV(DaskBaseSearchCV):
         scoring=None,
         iid=True,
         refit=True,
-        cv=None,
         error_score="raise",
         return_train_score=_RETURN_TRAIN_SCORE_DEFAULT,
         scheduler=None,
@@ -393,12 +395,11 @@ class BaseIncrementalSearchCV(DaskBaseSearchCV):
         self.params = params
         self.test_size = test_size
         self.random_state = random_state
-        super(BaseIncrementalSearchCV, self).__init__(
+        super(BaseIncrementalSearch, self).__init__(
             estimator,
             scoring=scoring,
             iid=iid,
             refit=refit,
-            cv=cv,
             error_score=error_score,
             return_train_score=return_train_score,
             scheduler=scheduler,
@@ -420,10 +421,11 @@ class BaseIncrementalSearchCV(DaskBaseSearchCV):
         if isinstance(y, np.ndarray):
             y = da.from_array(y, y.shape)
         X = check_array(X, **kwargs)
+        kwargs["ensure_2d"] = False
         y = check_array(y, **kwargs)
         return X, y
 
-    def _get_cv(self, X, y, **kwargs):
+    def _get_train_test_split(self, X, y, **kwargs):
         """CV-Split the arrays X and y
 
         By default, :meth:`dask_ml.model_selection.train_test_split`
@@ -439,25 +441,16 @@ class BaseIncrementalSearchCV(DaskBaseSearchCV):
         )
         return X_train, X_test, y_train, y_test
 
-    def _get_ids_and_args(self):
-        # type: () -> Dict[int, Tuple(List[Dict], Union[Callable, object])]
-        """Abstract method for generating the argumnets passed to `fit`.
-
-        Returns
-        -------
-        ids_and_args : Dict
-            Should have integers for keys and the following arguments
-
-            * param_list : List[Dict]
-            * additional_fit_calls : Union[Callable, object]
-                Objects with a ``fit`` method have the bound
-                ``additional_fit_calls.fit`` method passed through to `fit`.
-                (Other) callables are simply passed through.
-        """
+    def _additional_calls(self):
         raise NotImplementedError
 
-    def _get_cv_results(self, results):
-        # type: (Dict) -> Dict
+    def _get_params(self):
+        # I don't think there's really a good default here.
+        # This assume random search. Could equally do grid search.
+        return ParameterSampler(self.params, self.n_iter)
+
+    def _get_history_results(self, results):
+        # type: (Results) -> Dict
         """Construct the CV results.
 
         Has the following keys:
@@ -473,62 +466,26 @@ class BaseIncrementalSearchCV(DaskBaseSearchCV):
         * partial_fit_calls
         * model_id
         """
-        info, model, history = zip(*results.values())
-        info = list(itertools.chain.from_iterable(x.values() for x in info))
-        model = list(itertools.chain.from_iterable(x.values() for x in model))
-        history = list(itertools.chain.from_iterable(history))
+        info, model, history = results
+        key = operator.itemgetter("model_id")
+        hist2 = sorted(history, key=key)
+        return hist2
 
-        scores = np.array([x["score"] for v in info for x in v])
-        ranks = np.argsort(-1 * scores) + 1
-        assert max(scores) == scores[ranks - 1][0]
-        model_ids = []
-        for loop_id in results:
-            loop_info = results[loop_id][0]
-            model_ids.extend(
-                [
-                    "{}-{}".format(loop_id, x["model_id"])
-                    for v in loop_info.values()
-                    for x in v
-                ]
-            )
-
-        cv_results = {
-            "params": [x["params"] for v in info for x in v],
-            "test_score": scores,
-            "mean_test_score": scores,  # for sklearn comptability
-            "rank_test_score": ranks,
-            "mean_partial_fit_time": np.array(
-                [np.mean(x["partial_fit_time"]) for v in info for x in v]
-            ),
-            "std_partial_fit_time": np.array(
-                [np.std(x["partial_fit_time"]) for v in info for x in v]
-            ),
-            "mean_score_time": np.array(
-                [np.mean(x["score_time"]) for v in info for x in v]
-            ),
-            "std_score_time": np.array(
-                [np.std(x["score_time"]) for v in info for x in v]
-            ),
-            "partial_fit_calls": [x["partial_fit_calls"] for v in info for x in v],
-            "model_id": model_ids,
-        }
-        return cv_results
-
-    def _get_best(self, results, cv_results):
+    def _get_best(self, results, history_results):
         # type: (Dict, Dict) -> Estimator
         """Select the best estimator from the set of estimators."""
-        rank = cv_results["rank_test_score"]
-        best_index = np.arange(len(rank))[rank - 1][0]
-        best_model_id = cv_results["model_id"][best_index]
-        assert cv_results["test_score"][best_index] == max(cv_results["test_score"])
+        best_model_id = first(results.info)
+        key = operator.itemgetter("model_id")
+        best_index = 0
+        # history_results is sorted by (model_id, partial_fit_calls)
+        # best is the model_id with the highest partial fit calls
+        for k, v in itertools.groupby(history_results, key=key):
+            v = list(v)
+            best_index += len(v)
+            if k == v:
+                break
 
-        loop_id, model_id = map(int, best_model_id.split("-"))
-        return results[loop_id][1][model_id].result(), best_index
-
-    def _update_results(self, results):
-        # type: (Dict) -> None
-        """Update `self` with attributes extracted from `results`."""
-        pass
+        return results.models[best_model_id], best_index
 
     def _process_results(self, results):
         """Called with the output of `fit` immediately after it finishes.
@@ -541,7 +498,7 @@ class BaseIncrementalSearchCV(DaskBaseSearchCV):
         return results
 
     def fit(self, X, y, **fit_params):
-        """Find the best parameters for a particular model
+        """Find the best parameters for a particular model.
 
         Parameters
         ----------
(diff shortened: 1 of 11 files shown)


dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +354,228 @@ def fit(
...
+
+class BaseIncrementalSearchCV(DaskBaseSearchCV):
+    """Base class for estimators using the incremental `fit`.
+
+    Subclasses must implement the following abstract methods
+
+    * _get_ids_and_args : Build the dict of
+       ``{model_id : (param_list, additional_calls)}``
+        that are eventually passed to `fit`
+    * _get_cv_results : Build the dict containing the CV results.
```
mrocklin: I would expect this to be shared among implementations.  I think that the `history` output of the `fit` function should have all of the information here.  Are there choices that are likely to change between implementations?

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +354,228 @@ def fit(
...
+
+
+class BaseIncrementalSearchCV(DaskBaseSearchCV):
+    """Base class for estimators using the incremental `fit`.
+
+    Subclasses must implement the following abstract methods
+
+    * _get_ids_and_args : Build the dict of
+       ``{model_id : (param_list, additional_calls)}``
+        that are eventually passed to `fit`
```
mrocklin: I would expect the `additional_calls` function to be a method of this class that developers would override.

I think that we probably don't need users to define `model_id`s.  The `fit` function will do this for us.  

I do think that we need a way to decide initial parameter lists.  This might be an overrideable method.

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +354,228 @@ def fit(
...
+    * _get_cv_results : Build the dict containing the CV results.
+    * _get_best_estimator : Select the best estimator from the set of
+      fitted estimators
```
mrocklin: I think that this should be easy to make a default for.  I would pass over all of the models in the resulting models dict and then pick the one with the best score coming out of `info`

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +354,228 @@ def fit(
...
+    * _update_results : Update self with attributes from the fitted
+      esimators.
```
mrocklin: I would also expect this to be somewhat standardized among different implementations.  What does this need to do?

dask_ml/model_selection/_incremental.py
TomAugspurger: Yes, I think so. Initially, the structure of my `IncrementalRandomizedSearchCV` differed from `Hyperband`, as my results were simply `results = fit(...)`, and not `results = {model_id: fit(...) for ...}`.

Now that both have the consistent `results` type of `{model_id: (info, models, history)}`, more of this should be sharable.

same applies to https://github.com/dask/dask-ml/pull/356#discussion_r215591027

dask_ml/model_selection/_incremental.py
TomAugspurger: I'm not really sure. For Hyperband, it's something like

```python

        self.history_ = sum([SHA._history for SHA in SHAs.values()], [])
        self.metadata_ = {
            "models": sum(m["models"] for m in meta),
            "partial_fit_calls": sum(m["partial_fit_calls"] for m in meta),
            "brackets": meta,
        }

```

This could probably have a default implementation that extracts the results that will likely be useful to all base classes (I'm just not sure what those items are at the moment).

mrocklin: I think that my objective here is for users not to have to write too much to make things work.  Some of these hooks should probably be optional.

dask_ml/model_selection/_incremental.py
stsievert: In my implementation of `_get_cv_results`, the `bracket` key (`model_id` here I think) was only required to make it so the `model_id`s were unique across brackets. This was important for downstream processing; I wanted unique model IDs for each bracket, not shared (and isn't the cleanest code).

This solution can be used with the "consistent `results` type of `{model_id: (info, models, history)}`". The result I came up at [_hyperband.py#L315-L318][1] is something like 

``` python
for loop_id, (info, models, hist) in results.items():   # results defined in comment above
    for h in hist:
        h["model_id"] = "loop{}-{}".format(loop_id, h["model_id"])
```

[1]:https://github.com/dask/dask-ml/blob/4e142d75bd52dd351c90a59e488ad9a62e38daa3/dask_ml/model_selection/_hyperband.py#L315-L318

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +354,228 @@ def fit(
...
+            )
+            for model_id, (param_list, additional_calls) in ids_and_args.items()
+        }
+        self._update_results(results)
+        self.cv_results_ = self._get_cv_results(results)
+        self.best_estimator_ = self._get_best_estimator(results)
+        return self
+
+
+class IncrementalRandomizedSearchCV(BaseIncrementalSearchCV):
```
stsievert: There are two parts to this class: the method of choosing parameters, and the incremental/adaptive method implemented (currently `remove_worst`). I'd rather the name reflect the incremental/adaptive component: maybe something like `KillWorstModelCV`?

[1]:https://pytorch.org/docs/stable/optim.html#torch.optim.lr_scheduler.ReduceLROnPlateau

mrocklin: I think that this is only up here as a strawman to see what things look like when extending the class.  I don't think that the `remove_worst` function is a particularly pragmatic approach.  It was just something simple that was small enough for me to put into a docstring.

dask_ml/model_selection/_incremental.py
TomAugspurger: @mrocklin does `_get_ids_and_args` make sense now? Initially the building of `param_lists` and `additional_calls` was two separate methods, but it seems like they may be dependent on eachother. So I figured having a single method to override the things passed into `fit` made the most sense.

This is the only mandatory method that has to be implemented by subclasses now.

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +356,302 @@ def fit(
...
+        Returns
+        -------
+        ids_and_args : Dict
+            Should have integers for keys and the following arguments
+
+            * param_list : List[Dict]
+            * additional_fit_calls : Union[Callable, object]
+                Objects with a ``fit`` method have the bound
+                ``additional_fit_calls.fit`` method passed through to `fit`.
+                (Other) callables are simply passed through.
```
mrocklin: I would design things so that the `IncrementalRandomizedSearchCV` class has `remove_worst` as a method, rather than as a return value of `_get_ids_an_args`.  I think that the `remove_worst` method is likely to be the main point of variance between different classes, and so is something that people will probably expect to specify at the class level.

(also, just to be clear, we shouldn't actually publish a class with the `remove_worst` function as a genuine solution.  It's not a good solution)

I also don't understand why this returns a dictionary.

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +356,302 @@ def fit(
...
+                y_train,
+                X_test,
+                y_test,
+                additional_calls=getattr(additional_calls, "fit", additional_calls),
+                fit_params=fit_params,
+                scorer=scorer,
+                random_state=self.random_state,
+            )
+            for model_id, (param_list, additional_calls) in ids_and_args.items()
+        }
```
mrocklin: The Hyperband implementation is weird in that they'll want to call fit several times rather than just once (which I think will be more common).  Additionally, Hyperband will want to call fit many times concurrently, so we'll need to use coroutines for this.  

Short term I suggest that we implement classes for fitting one model,  RandomSearch, and SuccessiveHalving using the functions found in [this notebook](https://gist.github.com/mrocklin/4c95bd26d15281d82e0bf2d27632e294).  Then we look at Hyperband as a special case that has to call fit a few times and see what extra methods might have to be pulled out.

I'm happy to take a look at this later today as well if that would be useful.

mrocklin: In short, I think that this approach is more complicated than necessary specifically because it is targetting Hyperband first and because Hyperband is a bit of a special case.  I think that we should target some of the simpler situations first, build conventions around them, and then see how we can add release valves for Hyperband to also work.

TomAugspurger: My reply seems to have been lost.

tldr: that seems sensible. I agree that Hyperband's multiple calls to fit complicated this PR, and I don't know how representative that will be of other strategies (I assume not).

TomAugspurger:
@mrocklin I think this is mostly done for now.

I may do a bit of restructuring to split out the CV-specific parts of DaskBaseSearchCV into a mixin. We currently inherit some things that don't make so much sense.

Followup items are to try out random search and successive halving from https://gist.github.com/mrocklin/4c95bd26d15281d82e0bf2d27632e294 on this. Do you want to do that or should I?

mrocklin:
> Do you want to do that or should I?

I can take a look.  It'll force me to better understand the system you've
put in place here.

TomAugspurger:
OK, just merged in master to hopefully fix the CI failure.

If you want to "check your work", here's a solution :)

<details>

```diff
diff --git a/dask_ml/model_selection/_incremental.py b/dask_ml/model_selection/_incremental.py
index c2983ae..b208ae9 100644
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -618,3 +618,68 @@ class RandomizedWorstIncrementalSearch(BaseIncrementalSearch):
 
     def _get_params(self):
         return ParameterSampler(self.parameters, self.n_iter, self.random_state)
+
+
+class RandomIncrementalSearch(BaseIncrementalSearch):
+    def __init__(
+        self,
+        estimator,
+        param_distribution,
+        n_iter=10,
+        max_iter=100,
+        patience=10,
+        tol=0.001,
+        test_size=0.15,
+        random_state=None,
+        scoring=None,
+        iid=True,
+        refit=True,
+        error_score="raise",
+        return_train_score=_RETURN_TRAIN_SCORE_DEFAULT,
+        scheduler=None,
+        n_jobs=-1,
+        cache_cv=True,
+    ):
+        self.max_iter = max_iter
+        self.n_iter = n_iter
+        self.patience = patience
+        self.tol = tol
+        super(RandomIncrementalSearch, self).__init__(
+            estimator,
+            param_distribution,
+            test_size,
+            random_state,
+            scoring,
+            iid,
+            refit,
+            error_score,
+            return_train_score,
+            scheduler,
+            n_jobs,
+            cache_cv
+        )
+
+    def _get_params(self):
+        return ParameterSampler(self.parameters, self.n_iter)
+
+    def _additional_calls(self, info):
+        out = {}
+        max_iter = self.max_iter
+        patience = self.patience
+        tol = self.tol
+
+        for ident, records in info.items():
+            if max_iter is not None and len(records) > max_iter:
+                out[ident] = 0
+
+            elif len(records) > patience:
+                old = records[-patience]["score"]
+                if all(d["score"] < old + tol for d in records[-patience:]):
+                    out[ident] = 0
+                else:
+                    out[ident] = 1
+
+            else:
+                out[ident] = 1
+        print(out)
+        return out
```

</details>

TomAugspurger:
I think some of these parameters should be removed:

* iid
* refit
* error_score
* return_train_score
* scheduler
* n_jobs
* cache_cv

But need to double check first.

mrocklin:
Care to push up your RandomSearch implementation?

mrocklin:
NM, I'm on it

mrocklin:
So we're not handling history correctly.  It's not clear to me what history should be here actually.  In SKLearn we have one entry per parameter/CV-split.  Now we have many entries.  

Currently this is manifesting because things like `best_score_` is broken (see recent commit).

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +381,309 @@ def fit(
...
+        * mean_partial_fit_time
+        * std_partial_fit_time
+        * mean_score_time
+        * std_score_time
+        * partial_fit_calls
+        * model_id
+        """
+        info, model, history = results
+        key = operator.itemgetter("model_id")
+        hist2 = sorted(history, key=key)
```
TomAugspurger: Can you try sorting by the tuple `(model_id, partial_fit_calls)` here? That may fix the history issue.

TomAugspurger: To check my understanding, is it true that for a given `model_id`, `partial_fit_calls` is always increasing as we move through history?

mrocklin: Yes

mrocklin:
`n_iter` here is closer to "number of initial parameters to try".  Should we change the parameter name?

TomAugspurger:
> Any thoughts on how to handle test_size? We probably can't assume that a flat 15% makes sense any longer.

I briefly considered something that would look at `X.nbytes` and the size of distributed RAM, but didn't pursue it. I'm not sure how robust it would be.

> Any thoughts on a better name than additional_calls?

`update_calls`? Not really sure.

> n_iter here is closer to "number of initial parameters to try". Should we change the parameter name?

`n_iter` is the name used by `RandomizedSearchCV`, but it has a slightly different meaning in this context so it may be safe to diverge. Perhaps `n_initial_parameters`.

stsievert:
> Any thoughts on how to handle test_size? We probably can't assume that a flat 15% makes sense any longer.

The only purpose of a test set is to provide a stable benchmark. I'd imagine for most uses, having a test dataset that fits in RAM will serve this purpose. I didn't put much thought into 15%, but chose it because it does a reasonable job at trading off amount of data to learn from with test stability (for some model complexity, for some definition of "reasonable").

I guess there's no requirement that the test dataset fit in RAM (though that's likely good enough). We could distribute the test set, score the model on each chunk and take a weighted average with the chunk size.

TomAugspurger:
> I guess there's no requirement that the test dataset fit in RAM (though that's likely good enough).

Hmm, interesting. Currently we explicitly call `compute` the test sets to convert them to ndarrays. We also wrap the scorer in `dask.delayed`. In practice, both of those means that the training set has to fit in memory, even if it doesn't actually need to in theory.

It doesn't look *too* difficult to modify `_fit` to support this directly. I think the main thing is ensuring that the scorer can support dask arrays, and that it returns a delayed score.

Though not sure how high-priority this should be. I share your suspicion that an in-memory dataset is likely to be good enough most of the time.

TomAugspurger:
FYI @mrocklin I pushed a commit fixing a few compatibility things and your `best_score_` issue.

mrocklin:
With the current implemenation the testing data has to fit in memory

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +383,384 @@ def fit(
...
+            cache_cv,
+        )
+
+    def _get_params(self):
+        return ParameterSampler(self.parameters, self.n_iter)
+
+    def _additional_calls(self, info):
+        def inverse(time):
+            """ Decrease target number of models inversely with time """
+            return int(self.n_iter / (1 + time) ** self.decay_rate)
```
TomAugspurger: Should this be `self.max_iter`? `self.n_iter` is used for the initial sampling. (we probably need more distinct names for these)

TomAugspurger:
Pushed some documentation updates.

On naming, it seems like the class name could capture a few components

* initial parameter selection strategy (Randomized vs. Grid)
* additional calls strategy
* the fact that we're doing hyperparameter optimization.

So I'd propose the naming rubric `<Randomized|Grid><Strategy><Search>`.

* `RandomizedPlateauSearch` (not sure about plateau)
* `RandomziedSuccessiveReductionSearch (that's a mouthful)

OTOH, do we think grid search is useful? We don't have anything that does that yet (though it's trivial to implement; just override `_get_params` to use ParameterGrid).

TomAugspurger:
Added those two to the public API, and tried to write a bit of prose about what we mean by incremental hyperparameter optimization. Some additional thoughts there would be appreciated.

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +394,548 @@ def fit(
...
+
+    max_iter : int, default 100
+        Maximum number of partial fit iterations per model. This is passed
+
+    patience : int, default 10
+        The number of previous models scores to consider when determining
+        whether the model has converged. The most recent score must be at
+        at most `tol` better than the all of the previous `patience` scores.
+        Increasing `patience` will tend to provide better models, at the cost
+        of increased computation.
```
stsievert: Maybe `patience` should be in terms of epochs/passes through data, not number of `partial_fit` calls (and a float instead of an int). It helps remove an implicit dependence on the chunk size. That's what other similar classes use – e.g., PyTorch's [`ReduceLROnPlateau`][1] defaults to `patience` being 10 epochs.

This would require having more metadata to record the number of examples the estimator has seen, which will involve changes to `partial_fit`.

[1]:https://pytorch.org/docs/stable/optim.html#torch.optim.lr_scheduler.ReduceLROnPlateau

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +394,548 @@ def fit(
...
+                    out[ident] = 0
+                else:
+                    out[ident] = 1
+
+            else:
+                out[ident] = 1
+        return out
+
+
+class SuccessiveReductionSearch(BaseIncrementalSearch):
```
stsievert: Could we make this class private? It's required for #221 but I'm not inclined to expose it.

stsievert:
> So I'd propose the naming rubric `<Randomized|Grid><Strategy><Search>`.

Maybe `<Strategy>Search`? The initial selection method isn't the focus of this class, though it is important. I'd be inclined to choose random parameter sampling by default unless `params` is a list, and maybe provide a `parameter_selection` keyword to choose between grid and random search.

dask_ml/model_selection/_incremental.py
TomAugspurger: We're in a private submodule, and it's not imported in `model_selection/__init__.py`, so at the moment it's private.

TomAugspurger:
> Maybe <Strategy>Search?

Sounds reasonable.

> maybe provide a parameter_selection keyword to choose between grid and random search.

That seems more natural to me.

dask_ml/model_selection/_incremental.py
mrocklin: > I'm not inclined to expose it

Why not?

stsievert: Almost all of this class can be revealed through Hyperband if some "skip these brackets" interface is provided. This would allow the user to say "focus on the most exploratory brackets" and let Hyperband select the parameters.

Additionally, this class

* is implemented to allow Hyperband to work, and wouldn't be present otherwise
* has 3 hyper-params, and it's not clear how to trade off between them (`n_iter`, `max_iter` and `decay_rate`).

dask_ml/model_selection/_incremental.py
mrocklin: Using `self.n_iter` is correct here, this is a decaying exponential starting at the number of initial parameters.

dask_ml/model_selection/_incremental.py
mrocklin: Sometimes we train on too much data and would like to stop early.

I agree that losing a dependence on chunk size would be useful.  I don't think that we can go with full dataset size though.  Number of records?  Amount of time?  Amount of bytes?

dask_ml/model_selection/_incremental.py
mrocklin: > is implemented to allow Hyperband to work, and wouldn't be present otherwise

That's not necessarily true.  I like this class because it achieves the general objective of "train a lot at the beginning but focus on well-performing models" while also being simple and easy to understand.  I think that we should point people towards Hyperband in docs, but I don't see a reason to hide this implementation.

mrocklin: I think that you're generally saying "but Hyperband performs better" and I'm saying "that's not the only thing being considered".

mrocklin:
OK, so some pending topics that I see:

1.  Rename SuccessiveReduction -> ExponentialDecay or something similar?
2.  Rename additional_calls?
3.  Add max_iter and/or patience/tol to successive ?  I'll need to try things out to see how this does. 
 @stsievert somewhere you have a notebook in which you apply your hyperband solution onto a real-world problem that uses CPUs only.  Would you mind providing a link to that here?

TomAugspurger:
One more thing to consider: exactly what is shared between DaskBaseSearchCV and this class. I think there's not too much, so it might be cleaner to not inherit from DaskBaseSearchCV. I can take care of that on this branch or as a followup.

TomAugspurger:
On the names, I think I prefer `ExponentialDecay` since it's more indicative of how things are successively reduced.

I haven't come up with anything better than `additional_calls`. At first it didn't make sense to me, but after having played with it on this branch it seems like the most sensible name.

stsievert:
> Add max_iter and/or patience/tol to successive? ... Would you mind providing a link to that here?

Here's a link to the notebook that searches for `patience`: https://gist.github.com/stsievert/33bdc47b52ffc085dddd75e9b719cc07 I'm running this right now on GCP, and it should be done by the end of today.

> Rename additional_calls?

Maybe `get_additional_calls`?

mrocklin:
Do you have any qualitative results from things you've observed?  I'm
surprised that it takes a full day to get a sense for what's going on
here.  I was planning to just run something for 20m or so to get a sense of
what was going on.

stsievert:
> Do you have any qualitative results from things you've observed? 

Having a short patience decreases the best score Hyperband finds. It looks like having `patience` be 20 epochs (aka passes over training data) preserves the best score Hyperband finds.

It takes so long because running Hyperband once takes about an hour (at least for this example on GCP). I'm running it over different `patience`s and doing a couple repeats too.

dask_ml/model_selection/_incremental.py
stsievert: Agreed. I think we should provide the parameters of this class in `HyperbandCV.metadata`, which will allow the user to use this class if they desire (and then we won't have to implement "skip this bracket").

mrocklin:
OK, so ExponentialDecaySearch now has all of the functionality of RandomizedSearch if we set `decay_rate=0`.

mrocklin:
So how do people feel about the following:

```python
ExponentialDecay(model, params, n_initial_parameters=100)  # 100 initial parameters
ExponentialDecay(model, params, n_initial_parameters='grid')  # Use full grid
```

TomAugspurger:
That sounds reasonable to me.

TomAugspurger:
https://github.com/dask/dask-ml/pull/356/commits/658109ecb7ffea7bf25aa7d3c9ab74d08b84d17e makes `BaseIncrementalSearch` not inherit from `DaskBaseSearchCV`, since many of the methods / parameters didn't make sense here.

I opted to simply copy the implementation over for the ones we care about. It'd be DRYer to move those to a shared Mixin, but this seems simpler. No preference here.

Finally, https://github.com/dask/dask-ml/pull/356/commits/af25b79774f1bae04bbd64c1f988f6d6c2b3e645 adds docs for using an incremental hyper-parameter optimizer with `ParallelPostFit`. That'll be, I think, the most common use case, so that `.score`, `.predict`, etc. happen on the cluster for dask arrays.

mrocklin:
Should we automatically provide ParallelPostFit functionality for
predict/score on this class?

TomAugspurger:
I had a partially written comment thinking through that, but deleted it
after having not come to a conclusion.

The argument for auto-wrapping with something like ParellelPostFit is that
it's kind of the whole purpose for
this incremental hyperparameter optimization.

The downside is that it's magic that's liable to break things. Initially
wrappers.Incremental had some of this
magic for get_params and set_params, but we removed as it caused larger
issues (e.g. both the metaestimator
and the underlying estimator having the same keyword).

So for now, I'd recommend avoiding the temptation to be clever and just
documenting things.

stsievert:
It looks like this is ready for integration with Hyperband. I'm planning on rebasing off this branch and building up Hyperband tomorrow.

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +394,558 @@ def fit(
...
+    def score(self, X, y=None):
+        if self.scorer_ is None:
+            raise ValueError(
+                "No score function explicitly defined, "
+                "and the estimator doesn't provide one %s" % self.best_estimator_
+            )
+        return self.scorer_(self.best_estimator_, X, y)
+
+
+class RandomizedIncrementalSearch(BaseIncrementalSearch):
```
stsievert: Wasn't this going to be renamed to `StopOnPlateauSearch` (with other comment `StopOnPlateauSearchCV`)?

mrocklin:
> Yeah, this feels pretty close to being ready. @mrocklin do you have any objections to merging it as is?

No objections to merging as is, but I'd like to take another look at names and the API before releasing if possible.  I probably won't get to this before next week.

dask_ml/model_selection/_incremental.py
mrocklin: I don't think that the term `StopOnPlateau` is easily recognizable by non-experts.

mrocklin: I'm not super in-favor of `RandomizedIncrementalSearch` either though.  Given that this is now a subset of functionality with `ExponentialDecaySearch` (also not a great name) I think that we should consider merging them, choosing a more generic name, and providing a couple of keywords to choose between some common choices (exponential decay, patience, grid/random).

mrocklin:
How do we feel about the following?

Exponential decay with patience on randomly selected parameters

    IncrementalSearchCV(model, params, n_initial_params=1000, decay_rate=1.0, patience=20, tol=0.1)

100 partial_fit calls with no decay on a full grid

    IncrementalSearchCV(model, params, n_initial_params='grid', decay_rate=0.0, max_iter=100)

cc @ogrisel your feedback on API and functionality here would be welcome

TomAugspurger:
That seems reasonable, with the caveat that we shouldn't have the CV suffix if we're just doing a single train-test split. Now that `RandomizedIncrementalSearch` is a special-case with no decay, we should get rid of it.

mrocklin:
OK, done.  

@TomAugspurger I'm curious about your thoughts on the `Incremental` meta-estimator.  Are there still cases where this is useful (for example in pipelines) or should we start to deprecate it in favor of this mechanism?

TomAugspurger:
I think `wrappers.Incremental` is still useful when you aren't doing hyperparameter optimization, and it's a bit easier to explain I think. I'll update the docs to try to clarify things.

mrocklin:
For the default size of the test data how about using the size of a single chunk.  For example if the input data has ten chunks then we use 10%.  That provides some guarantee that the result will fit into memory.  This would be the default value if `test_size=None`

TomAugspurger:
Yes that's a very good idea.

dask_ml/model_selection/_incremental.py
```diff
@@ -347,3 +389,415 @@ def fit(
...
+
+    @if_delegate_has_method(delegate=("best_estimator_", "estimator"))
+    def transform(self, X):
+        self._check_is_fitted("transform")
+        return self.best_estimator_.transform(X)
+
+    @if_delegate_has_method(delegate=("best_estimator_", "estimator"))
+    def inverse_transform(self, Xt):
+        self._check_is_fitted("inverse_transform")
+        return self.best_estimator_.transform(Xt)
```
mrocklin: It looks like the methods above are untested.

TomAugspurger: I can put some together quick.

TomAugspurger:
Edge case when there's a single parameter

```python

    model = SGDClassifier(tol=1e-3)

    params = {"alpha": np.logspace(-2, 10, 3)}

    search = IncrementalSearch(model, params, n_initial_parameters="grid")
    search.fit(X, y, classes=[0, 1])
 ```

```pytb
StopIteration                             Traceback (most recent call last)
~/sandbox/dask-ml/dask_ml/model_selection/_incremental.py in _fit(self, X, y, **fit_params)
    533         history_results = self._get_history_results(results)
--> 534         best_estimator, best_index = self._get_best(results, history_results)
    535         best_estimator = yield best_estimator

~/sandbox/dask-ml/dask_ml/model_selection/_incremental.py in _get_best(self, results, history_results)
    486         """Select the best estimator from the set of estimators."""
--> 487         best_model_id = first(results.info)
    488         key = operator.itemgetter("model_id")

~/Envs/dask-dev/lib/python3.6/site-packages/toolz/itertoolz.py in first(seq)
    367     """
--> 368     return next(iter(seq))
    369

StopIteration:

The above exception was the direct cause of the following exception:

RuntimeError                              Traceback (most recent call last)
<ipython-input-7-048117aa2d7a> in <module>()
----> 1 search.fit(X, y, classes=[0, 1])

~/sandbox/dask-ml/dask_ml/model_selection/_incremental.py in fit(self, X, y, **fit_params)
    559             Additional partial fit keyword arguments for the estimator.
    560         """
--> 561         return default_client().sync(self._fit, X, y, **fit_params)
    562
    563     @if_delegate_has_method(delegate=("best_estimator_", "estimator"))

~/Envs/dask-dev/lib/python3.6/site-packages/distributed/client.py in sync(self, func, *args, **kwargs)
    645             return future
    646         else:
--> 647             return sync(self.loop, func, *args, **kwargs)
    648
    649     def __repr__(self):

~/Envs/dask-dev/lib/python3.6/site-packages/distributed/utils.py in sync(loop, func, *args, **kwargs)
    275             e.wait(10)
    276     if error[0]:
--> 277         six.reraise(*error[0])
    278     else:
    279         return result[0]

~/Envs/dask-dev/lib/python3.6/site-packages/six.py in reraise(tp, value, tb)
    691             if value.__traceback__ is not tb:
    692                 raise value.with_traceback(tb)
--> 693             raise value
    694         finally:
    695             value = None

~/Envs/dask-dev/lib/python3.6/site-packages/distributed/utils.py in f()
    260             if timeout is not None:
    261                 future = gen.with_timeout(timedelta(seconds=timeout), future)
--> 262             result[0] = yield future
    263         except Exception as exc:
    264             error[0] = sys.exc_info()

~/Envs/dask-dev/lib/python3.6/site-packages/tornado/gen.py in run(self)
   1131
   1132                     try:
-> 1133                         value = future.result()
   1134                     except Exception:
   1135                         self.had_exception = True

~/Envs/dask-dev/lib/python3.6/site-packages/tornado/gen.py in run(self)
   1145                             exc_info = None
   1146                     else:
-> 1147                         yielded = self.gen.send(value)
   1148
   1149                     if stack_context._state.contexts is not orig_stack_contexts:

RuntimeError: generator raised StopIteration
```

mrocklin:
On it

TomAugspurger:
Added small tests for predict_proba, predict_log_proba, transform, and decision_function.

mrocklin:
Sometimes the shape of the result in test_transform differs.  Is this expected?

TomAugspurger:
Sorry, fixed.

mrocklin:
+1

PR #359, opened 2018-09-07
Use dask_sphinx_theme for styling in docs

mrocklin:


TomAugspurger:
https://travis-ci.org/dask/dask-ml/builds/425863821 is the build that'll be deployed.

TomAugspurger:
Live a ml.dask.org. Things seem to be in good shape.

TomAugspurger:
Huh, the dask logo is missing on http://ml.dask.org/modules/api.html

Haven't investigated why yet.

mrocklin:
It's using a relative rather than absolute path

PR #360, opened 2018-09-10
Doc redirect

TomAugspurger:


TomAugspurger:
I made changes on the RTD website to have the `/rtd` branch be the latest.

I had to cut down on what is actually done so that the RTD build actually finishes. I'll followup with a PR restoring those changes.

TomAugspurger:
Actually, since this is against the `/rtd` branch the docs at ml.dask.org should be fine.

TomAugspurger:
For future reference, I should have used https://docs.readthedocs.io/en/latest/guides/adding-custom-css.html

```
def setup(app):
    app.add_javascript('js/custom.js')
```

rather than trying to override layout.html.

Issue #347, opened 2018-08-29
Add PolynomialTransformer


TomAugspurger:
http://scikit-learn.org/dev/modules/generated/sklearn.preprocessing.PolynomialFeatures.html#sklearn.preprocessing.PolynomialFeatures

This *should* be relatively straightforward, but there may be unexpected difficulties along the way.

datajanko:
If you don't mind, I will work on that

TomAugspurger:
That'd be great!

datajanko:
So I am making some progress here and fitting works so far.

Now I just realized, that typical dask-ml transformer that work on numpy arrays, also put numpy arrays out. However, for the polynomial features, this might not be desired. If an array, fits into ram, due to the added columns it might be to big. So my suggestion would be to always return a dask-array or dask data frame object with the chunk size, related to the size of the input object.

Does this make sense? Different suggestions?

TomAugspurger:
I think I'd prefer returning NumPy array output for NumPy array input.

Generally, Dask-ML only works on arrays that are partitioned vertically. At
some point, you'll need to have an entire block's worth of columns in
memory.

For the user whose original ndarray fits in memory, but the the transformed
array doesn't, I would recommend they
convert the ndarray to a Dask Array with some number of blocks before
passing it to PolynomialTransformer.

Does that make sense?

datajanko:
Thanks for the quick response Tom. It makes sense, and putting some burden on the user is completely fine for me.

datajanko:
I don't know if it's better to discuss here or in the WIP pull requests.

Until now I learned a lot about desk's internals which is nice. 

However, for the next steps, I'd need some feed back on the following issues:

- What should I do with chunks of unknown size/shape. This can happen if you have a dask data frame and call values.
- Concerning DataFrames: I'd tackle it by converting the data frame to a dask-array/numpy array, do the transforms there and just construct a new frame with columns from the get_feature_names method. Comments on that?
- Sparse matrices/frames. Not sure what is necessary here and what should be implemented.

Any further comment (on anything) is welcome.

PR #367, opened 2018-09-16
[MRG] Poly trans: Issue #347

datajanko:
At the current stage the pull request implements PolynomialFeatures in dask-ml for dask arrays with known shape and chunk size.

Implements Issue #347

### DIFF
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -919,8 +919,16 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
 
 
 class PolynomialFeatures(skdata.PolynomialFeatures):
+    """    preserve_dataframe : boolean
+            If True, preserve pandas and dask dataframes after transforming.
+            Using False (default) returns numpy or dask arrays and mimics
+            sklearn's default behaviour
 
-    __doc__ = skdata.PolynomialFeatures.__doc__
+        Examples
+    """
+
+    splitted_orig_doc = skdata.PolynomialFeatures.__doc__.split("    Examples\n")
+    __doc__ = "".join([splitted_orig_doc[0], __doc__, splitted_orig_doc[1]])
 
     def __init__(self, preserve_dataframe=False, *args, **kwargs):
         super(PolynomialFeatures, self).__init__(*args, **kwargs)
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -916,6 +916,82 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
 
         return X
 
+
 class PolynomialFeatures(skdata.PolynomialFeatures):
 
     __doc__ = skdata.PolynomialFeatures.__doc__
+
+    def fit(self, X, y=None):
+        """
+        Compute number of output features.
+        Parameters
+        ----------
+        X : array-like, shape (n_samples, n_features)
+            The data.
+        Returns
+        -------
+        self : instance
+
+        This implementation was copied and modified from Scikit-Learn.
+
+        See License information here:
+        https://github.com/scikit-learn/scikit-learn/blob/master/README.rst
+        """
+        if isinstance(X, dd.DataFrame):
+            # only shape[1] is important, compute necessary for check_array
+            n_samples, n_features = check_array(X.compute(), accept_sparse=True).shape
+        else:
+            n_samples, n_features = check_array(X, accept_sparse=True).shape
+        combinations = self._combinations(n_features, self.degree,
+                                          self.interaction_only,
+                                          self.include_bias)
+        self.n_input_features_ = n_features
+        self.n_output_features_ = sum(1 for _ in combinations)
+        return self
+
+    def transform(self, X):
+        """Transform data to polynomial features
+        Parameters
+        ----------
+        X : array-like or sparse matrix, shape [n_samples, n_features]
+            The data to transform, row by row.
+            Sparse input should preferably be in CSC format.
+        Returns
+        -------
+        XP : np.ndarray or CSC sparse matrix, shape [n_samples, NP]
+            The matrix of features, where NP is the number of polynomial
+            features generated from the combination of inputs.
+
+        This implementation was copied and modified from Scikit-Learn.
+
+        See License information here:
+        https://github.com/scikit-learn/scikit-learn/blob/master/README.rst
+        """
+        check_is_fitted(self, ['n_input_features_', 'n_output_features_'])
+
+        X = check_array(X, dtype=FLOAT_DTYPES, accept_sparse='csc')
+        n_samples, n_features = X.shape
+
+        if n_features != self.n_input_features_:
+            raise ValueError("X shape does not match training shape")
+
+        combinations = self._combinations(n_features, self.degree,
+                                          self.interaction_only,
+                                          self.include_bias)
+        if sparse.isspmatrix(X):
+            columns = []
+            for comb in combinations:
+                if comb:
+                    out_col = 1
+                    for col_idx in comb:
+                        out_col = X[:, col_idx].multiply(out_col)
+                    columns.append(out_col)
+                else:
+                    columns.append(sparse.csc_matrix(np.ones((X.shape[0], 1))))
+            XP = sparse.hstack(columns, dtype=X.dtype).tocsc()
+        else:
+            XP = np.empty((n_samples, self.n_output_features_), dtype=X.dtype)
+            for i, comb in enumerate(combinations):
+                XP[:, i] = X[:, comb].prod(1)
+
+        return XP
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -930,12 +930,17 @@ class PolynomialFeatures(skdata.PolynomialFeatures):
     splitted_orig_doc = skdata.PolynomialFeatures.__doc__.split("    Examples\n")
     __doc__ = "".join([splitted_orig_doc[0], __doc__, splitted_orig_doc[1]])
 
-    def __init__(self, preserve_dataframe=False, *args, **kwargs):
-        super(PolynomialFeatures, self).__init__(*args, **kwargs)
+    def __init__(
+        self,
+        degree=2,
+        interaction_only=False,
+        include_bias=True,
+        preserve_dataframe=False,
+    ):
+        super(PolynomialFeatures, self).__init__(degree, interaction_only, include_bias)
         self.preserve_dataframe = preserve_dataframe
 
     def fit(self, X, y=None):
-
         self._transformer = skdata.PolynomialFeatures()
         X_sample = X
         if isinstance(X, dd.DataFrame):
@@ -949,7 +954,6 @@ class PolynomialFeatures(skdata.PolynomialFeatures):
         return self
 
     def transform(self, X, y=None):
-
         if isinstance(X, da.Array):
             n_cols = len(self._transformer.get_feature_names())
             X = check_array(X, accept_multiple_blocks=False, accept_unknown_chunks=True)
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -939,7 +939,10 @@ class PolynomialFeatures(skdata.PolynomialFeatures):
     def transform(self, X, y=None):
 
         if isinstance(X, da.Array):
-            XP = X.map_blocks(self._transformer.transform, dtype=X.dtype)
+            n_cols = len(self._transformer.get_feature_names())
+            # might raise Error if columns are separated by chunks
+            chunks = (X.chunks[0], n_cols)
+            XP = X.map_blocks(self._transformer.transform, dtype=X.dtype, chunks=chunks)
         elif isinstance(X, pd.DataFrame):
             data = X.pipe(self._transformer.transform)
             columns = self._transformer.get_feature_names(X.columns)
@@ -947,7 +950,7 @@ class PolynomialFeatures(skdata.PolynomialFeatures):
         elif isinstance(X, dd.DataFrame):
             data = X.map_partitions(self._transformer.transform)
             columns = self._transformer.get_feature_names(X.columns)
-            XP = dd.from_array(data, "auto", columns)
+            XP = dd.from_dask_array(data, columns)
         else:
             # typically X is instance of np.ndarray
             XP = self._transformer.transform(X)
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -915,3 +915,7 @@ class OrdinalEncoder(BaseEstimator, TransformerMixin):
             X[col] = series
 
         return X
+
+class PolynomialFeatures(skdata.PolynomialFeatures):
+
+    __doc__ = skdata.PolynomialFeatures.__doc__
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -937,14 +937,15 @@ class PolynomialFeatures(skdata.PolynomialFeatures):
         See License information here:
         https://github.com/scikit-learn/scikit-learn/blob/master/README.rst
         """
-        if isinstance(X, dd.DataFrame):
-            # only shape[1] is important, compute necessary for check_array
-            n_samples, n_features = check_array(X.compute(), accept_sparse=True).shape
-        else:
-            n_samples, n_features = check_array(X, accept_sparse=True).shape
-        combinations = self._combinations(n_features, self.degree,
-                                          self.interaction_only,
-                                          self.include_bias)
+        n_samples, n_features = check_array(
+            X,
+            accept_unknown_chunks=True,
+            accept_dask_dataframe=True,
+            accept_sparse=True,
+        ).shape
+        combinations = self._combinations(
+            n_features, self.degree, self.interaction_only, self.include_bias
+        )
         self.n_input_features_ = n_features
         self.n_output_features_ = sum(1 for _ in combinations)
         return self
@@ -967,28 +968,47 @@ class PolynomialFeatures(skdata.PolynomialFeatures):
         See License information here:
         https://github.com/scikit-learn/scikit-learn/blob/master/README.rst
         """
-        check_is_fitted(self, ['n_input_features_', 'n_output_features_'])
+        check_is_fitted(self, ["n_input_features_", "n_output_features_"])
 
-        X = check_array(X, dtype=FLOAT_DTYPES, accept_sparse='csc')
+        X = check_array(
+            X,
+            accept_dask_dataframe=True,
+            accept_unknown_chunks=True,
+            accept_dask_array=True,
+        )
         n_samples, n_features = X.shape
 
         if n_features != self.n_input_features_:
             raise ValueError("X shape does not match training shape")
 
-        combinations = self._combinations(n_features, self.degree,
-                                          self.interaction_only,
-                                          self.include_bias)
-        if sparse.isspmatrix(X):
-            columns = []
-            for comb in combinations:
-                if comb:
-                    out_col = 1
-                    for col_idx in comb:
-                        out_col = X[:, col_idx].multiply(out_col)
-                    columns.append(out_col)
+        combinations = self._combinations(
+            n_features, self.degree, self.interaction_only, self.include_bias
+        )
+        # if sparse.isspmatrix(X):
+        #    columns = []
+        #    for comb in combinations:
+        #        if comb:
+        #            out_col = 1
+        #            for col_idx in comb:
+        #                out_col = X[:, col_idx].multiply(out_col)
+        #            columns.append(out_col)
+        #        else:
+        #            columns.append(sparse.csc_matrix(np.ones((X.shape[0], 1))))
+        #    XP = sparse.hstack(columns, dtype=X.dtype).tocsc()
+        if isinstance(X, da.Array):
+            # if n_samples is np.NAN:
+            #    n_samples = X.compute().shape[0]
+            chunk_x, _ = X.chunksize
+            column_vectors = []
+            for i, comb in enumerate(combinations):
+                # if bias allowed, first comb is (). np's slices differs from dask's
+                if not comb:
+                    column_vectors.append(
+                        da.ones((n_samples, 1), chunks=chunk_x, dtype=int)
+                    )
                 else:
-                    columns.append(sparse.csc_matrix(np.ones((X.shape[0], 1))))
-            XP = sparse.hstack(columns, dtype=X.dtype).tocsc()
+                    column_vectors.append(X[:, list(comb)].prod(1).reshape(-1, 1))
+            XP = da.concatenate(column_vectors, axis=1)
         else:
             XP = np.empty((n_samples, self.n_output_features_), dtype=X.dtype)
             for i, comb in enumerate(combinations):
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -945,10 +945,7 @@ class PolynomialFeatures(skdata.PolynomialFeatures):
             columns = self._transformer.get_feature_names(X.columns)
             XP = pd.DataFrame(data=data, columns=columns)
         elif isinstance(X, dd.DataFrame):
-            meta = pd.DataFrame(
-                columns=self._transformer.get_feature_names(X.columns), dtype=float
-            )
-            data = X.map_partitions(self._transformer.transform, meta=meta)
+            data = X.map_partitions(self._transformer.transform)
             columns = self._transformer.get_feature_names(X.columns)
             XP = dd.from_array(data, "auto", columns)
         else:
--- a/dask_ml/preprocessing/data.py
+++ b/dask_ml/preprocessing/data.py
@@ -921,6 +921,9 @@ class PolynomialFeatures(skdata.PolynomialFeatures):
 
     __doc__ = skdata.PolynomialFeatures.__doc__
 
+    def __init_(self, degree=2, interaction_only=False, include_bias=True):
+        super(PolynomialFeatures, self).__init__(degree, interaction_only, include_bias)
+
     def fit(self, X, y=None):
         """
         Compute number of output features.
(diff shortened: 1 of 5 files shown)


datajanko:
No, I have to apologize for not having invested more time on the dask fundamentals.
I was a bit distracted by realizing that the sklearn transformer is not stateless, which made me think that the "mapping" approach is not feasible.

Anyway, thanks for you valuable comments. I'm really happy to learn stuff.

datajanko:
Short Update: Following the lines of your suggestion works for dask arrays.

dask_ml/preprocessing/data.py
```diff
@@ -915,3 +915,62 @@ def inverse_transform(self, X):
...
 
         return X
+
+
+class PolynomialFeatures(skdata.PolynomialFeatures):
+
+    __doc__ = skdata.PolynomialFeatures.__doc__
+
+    def __init_(self, degree=2, interaction_only=False, include_bias=True):
+        super(PolynomialFeatures, self).__init__(degree, interaction_only, include_bias)
```
TomAugspurger: This `__init__` can just be removed if we aren't doing anything.

dask_ml/preprocessing/data.py
```diff
@@ -915,3 +915,62 @@ def inverse_transform(self, X):
...
+
+    def fit(self, X, y=None):
+        """
```
TomAugspurger: We can just inherit the docstring from scikit-learn I think.

dask_ml/preprocessing/data.py
```diff
@@ -915,3 +915,62 @@ def inverse_transform(self, X):
...
+        if isinstance(X, dd.DataFrame):
+            X_sample = X._meta
+        if isinstance(X, da.Array):
+            X_sample = np.ones((1, X.shape[1]), dtype=X.dtype)
+
+        self._transformer.fit(X_sample)
+        return self
+
+    def transform(self, X, y=None):
+        """Transform data to polynomial features
```
TomAugspurger: Same: just inherit.

dask_ml/preprocessing/data.py
```diff
@@ -915,3 +915,62 @@ def inverse_transform(self, X):
...
+        https://github.com/scikit-learn/scikit-learn/blob/master/README.rst
+        """
+        self._transformer = skdata.PolynomialFeatures()
+        X_sample = X
+        if isinstance(X, dd.DataFrame):
+            X_sample = X._meta
+        if isinstance(X, da.Array):
+            X_sample = np.ones((1, X.shape[1]), dtype=X.dtype)
+
+        self._transformer.fit(X_sample)
```
TomAugspurger: After we fit, we'll want to copy over the attribute learned attributes. I think that `dask_ml._utils.copy_learned_attributes(self._transformer, self)` will work here.

tests/preprocessing/test_data.py
```diff
@@ -471,3 +471,45 @@ def test_inverse_transform(self):
...
+
+
+class TestPolynomialFeatures:
+    def test_basic(self):
+        a = dpp.PolynomialFeatures()
+        b = spp.PolynomialFeatures()
+
+        a.fit(X)
+        b.fit(X.compute())
+        assert_estimator_equal(a._transformer, b)
```
TomAugspurger: This should be `assert_estimator_equal(a, b)`. It'll fail until you do the `copy_learned_attributes`.

dask_ml/preprocessing/data.py
```diff
@@ -915,3 +915,62 @@ def inverse_transform(self, X):
             X[col] = series
 
         return X
+
+
+class PolynomialFeatures(skdata.PolynomialFeatures):
+
+    __doc__ = skdata.PolynomialFeatures.__doc__
```
TomAugspurger: Since we inherit, this should be unnecessary (I may be wrong though).

datajanko: I just copied the approach from the additional transformers. However, 
```class C(skdata.PolynomialFeatures):
    pass
print(C.__doc__)
None```
shows that the doc string is not inherited for sklearn based transformers (that inherit from BaseEstimator). I also checked that for standard python classes the doc string is not passed to "children"

**Comment by datajanko** [comment]
Hmm, it seems like my only one of my local commits appears here. Anyway.

Code seems to work: The transformed arrays coincide (even after starting with data frames).

TODOs

- Documentation
- Code Cleanup
- Further tests, e.g. checking data frames (pandas and dask) will give the same result. Test fit more extensively, and maybe some ideas I'll have

Hope to commit tomorrow around the same time

**Comment by datajanko** [comment]
Unfortunately not the last update. I'm still having a failed test. I assume it happens due to incomplete meta data. An example:
```
from dask import array as da
from dask_ml.preprocessing import PolynomialFeatures
import dask.dataframe as ddf

X, y = make_classification(chunks=50)
df = X.to_dask_dataframe().rename(columns=str)

p = PolynomialFeatures()
p.fit_transform(df)
```

Throws an AttributeError:
```
AttributeError: 'DataFrame' object has no attribute '_repr_data'
```
However, this is probably only due to jupyter as
```
p.fit_transform(df).compute()
```
returns a dataframe but with a duplicate index. 
I have the suspicion, that this could be solved by passing to `map_partitions` the meta information about the column type. Could this be the way? Which types would one choose?

File: dask_ml/preprocessing/data.py [hunk 7]
```diff
@@ -915,3 +916,43 @@ def inverse_transform(self, X):
...
+            XP = X.map_blocks(self._transformer.transform, dtype=X.dtype)
+        elif isinstance(X, pd.DataFrame):
+            data = X.pipe(self._transformer.transform)
+            columns = self._transformer.get_feature_names(X.columns)
+            XP = pd.DataFrame(data=data, columns=columns)
+        elif isinstance(X, dd.DataFrame):
+            meta = pd.DataFrame(
+                columns=self._transformer.get_feature_names(X.columns), dtype=float
+            )
+            data = X.map_partitions(self._transformer.transform, meta=meta)
```
TomAugspurger: Hmm, slight issue here I think.

`_transformer.transform` is going to (AFAICT) always return an ndarray. So we have a choice

1. mimic scikit-learn, and return a dask array here (so just `XP = X.map_partitions(self._transformer.transform)`.
2. Add a keyword like `preserve_dask_dataframe=True`, to control whether we should convert the dask array back to a dask dataframe

Right now, I think that just returning a dask array is fine here.

datajanko:
Unfortunately, I was too optimistic, probably some artifacts in jupyter.
So as the test indicates, I'm creating duplicate index entries.
An example to reproduce:

```
from dask import array as da
from dask_ml.preprocessing import PolynomialFeatures
import dask.dataframe as ddf
X, y = make_classification(chunks=50)
df = X.to_dask_dataframe().rename(columns=str)
p = PolynomialFeatures()
```

Now

```
p.fit_transform(df)
```
throws - in jupyter - an AttributeError
```
AttributeError: 'DataFrame' object has no attribute '_repr_data'
```

However
```
p.fit_transform(df).compute()
```
provides an output, and the arrays seem to be correct. However, the index isn't.
I was training to add a meta keyword argument to map_partitions, but this didn't help either.

Any ideas?

Besides: I'd assume there should be some documentation added right? Any wishes on what to put where?

TomAugspurger:
The AttributeError indicates that something is wrong with the metadata. Haven't looked to see if it's a dask bug or not.

However, given my earlier comment we may not need to worry about it. If you don't pass `meta=meta` to the `X.map_partitions`, then we'll just return an ndarray and everything will be fine.

TomAugspurger:
For documentation you can add it to the list of "scikit-learn clones" in `doc/source/preprocessing.rst` and to `doc/source/modules/api.rst`, again under preprocessing.

dask_ml/preprocessing/data.py
datajanko: I'll have to think about this. In the other approach I have I didn't use meta. There map_partitions returned a dask dataframe. I'll have a deeper look into this. Maybe it's just too late ;)

TomAugspurger: Maybe make sure you're on a recent version of dask. Returning a dask array from `DataFrame.map_partitions` is somewhat new.

datajanko:
It seems that `map_partition`has some issues if the function creates more columns than originally in the dataframe! Even providing meta information was not helping. I'll post some details later. Though `map_partition` somehow shows some values, it seems one always would have to call `reset_index()` afterwards, if one can avoid errors. This seems a bit odd and unsafe.

It seems, that falling back to `map_block` for dask arrays is a solution.

datajanko:
Okay, so I investigated a bit and two (maybe three options) unfold.
1. Use map_partitions followed by reset_index
2. Use map_blocks but compute the number of rows

I'm not sure what is computational more expensive. 
<details>
<summary>Set Up </summary>

```
import dask
from sklearn.preprocessing.data import PolynomialFeatures

from dask import array as da
import dask.dataframe as ddf

X, y = make_classification(chunks=3, n_samples=6, n_features=2)
df = X.to_dask_dataframe().rename(columns=str)
X_from_df = df.to_dask_array()

spf = PolynomialFeatures()
spf.fit(X.compute())
```

</details>

Essentially, the issue is that the concatenation of `to_dask_array` and `to_dask_dataframe`
is not the identity function, and thus the divisions can not be inferred for data frames
<details>

```
print(X)
print(df.shape, df.divisions)

dask.array<normal, shape=(6, 2), dtype=float64, chunksize=(3, 2)>
(Delayed('int-97a26258-e448-4f93-830e-72985a66cde2'), 2) (0, 3, 5)

print('X:', X)
print("'id'(X):", X_from_df)
print("df:", df.shape, df.divisions) 
print("id(df):", df.to_dask_array().to_dask_dataframe().shape,  df.to_dask_array().to_dask_dataframe().divisions)

X: dask.array<normal, shape=(6, 2), dtype=float64, chunksize=(3, 2)>
'id'(X): dask.array<array, shape=(nan, 2), dtype=float64, chunksize=(nan, 2)>
df: (Delayed('int-756aa071-eaf0-4883-9bcf-383c630872a5'), 2) (0, 3, 5)
id(df): (Delayed('int-d7b1c932-daf6-44c6-943c-5de1ddd49297'), 2) (None, None, None)
```

</details>

For a dask array, we may provide the correct chunk sizes in the map_blocks function, and we obtain the correct output shapes. However, we also need to pass a dtype!

<details>

```
n_cols = len(spf.get_feature_names())
row_chunk_size = X.chunksize[0]
print(n_cols, row_chunk_size)

6 3

# wrong dimensions, need to set dtype
print(X.map_blocks(spf.transform, dtype='float'))
# only accurate solution: necessary to now the row_count!
print(X.map_blocks(spf.transform, dtype='float', chunks=(row_chunk_size, n_cols)))
# nan rows
print(X_from_df.map_blocks(spf.transform, dtype='float'))
# random guess also false
print(X_from_df.map_blocks(spf.transform, dtype='float', chunks=(10, n_cols)))
# shape: only one chunk_size predicts the correct shape!
print(X_from_df.map_blocks(spf.transform, dtype='float').compute().shape)

dask.array<transform, shape=(6, 2), dtype=float64, chunksize=(3, 2)>
dask.array<transform, shape=(6, 6), dtype=float64, chunksize=(3, 6)>
dask.array<transform, shape=(nan, 2), dtype=float64, chunksize=(nan, 2)>
dask.array<transform, shape=(20, 6), dtype=float64, chunksize=(10, 6)>
(6, 6)
```

</details>

Finally, the suggestion and issues for data frames:

```
#duplicate index -> reset_index will solve this
df_t = df.map_partitions(spf.transform).to_dask_dataframe(columns=spf.get_feature_names())
# Avoiding reset_index
X_t = X_from_df.map_blocks(spf.transform, dtype='float', chunks=(row_chunk_size, n_cols)).to_dask_dataframe(columns=spf.get_feature_names())
# Could use the dask graph and set the divisions manually, might be hard to do

print(df_t.index)
print('---------')
print(X_t.index)

Dask Index Structure:
npartitions=2
    int64
      ...
      ...
dtype: int64
Dask Name: from-dask, 12 tasks
---------
Dask Index Structure:
npartitions=2
0    int64
3      ...
5      ...
dtype: int64
Dask Name: from-dask, 14 tasks
```

We see, that there is no index information for the map_partitions approach (a compute shows duplicate index values). Moreover a third approach could be to use the dataframe-constructor, but we would need to calculate the divisions.

So: Which approach do you prefer? What is more expensive? Resetting the index or calculation the number of rows?

datajanko:
Any comments on the last commit?

dask_ml/preprocessing/data.py
```diff
@@ -915,3 +916,43 @@ def inverse_transform(self, X):
...
+        # pandas dataframe treated by sklearn and returns np.array
+        self._transformer.fit(X_sample)
+        copy_learned_attributes(self._transformer, self)
+        return self
+
+    def transform(self, X, y=None):
+
+        if isinstance(X, da.Array):
+            n_cols = len(self._transformer.get_feature_names())
+            # might raise Error if columns are separated by chunks
```
TomAugspurger: Can you explain this error to me? Is the issue something like

```python
X = da.random.uniform(size=(100, 10), chunks=(25, 5))`
```

i.e. we have two blocks on the features?

Typically, we use dask_ml.utils.check_array to handle this. There's a `accept_multiple_blocks` keyword.

I think that we *want* to raise an exception in this case, since the transform would need to be different for the different blocks.

dask_ml/preprocessing/data.py
```diff
@@ -915,3 +916,43 @@ def inverse_transform(self, X):
...
+            chunks = (X.chunks[0], n_cols)
+            XP = X.map_blocks(self._transformer.transform, dtype=X.dtype, chunks=chunks)
+        elif isinstance(X, pd.DataFrame):
+            data = X.pipe(self._transformer.transform)
+            columns = self._transformer.get_feature_names(X.columns)
+            XP = pd.DataFrame(data=data, columns=columns)
```
TomAugspurger: Slight preference to match scikit-learn exactly here, since this is a clone.

If you want, you could add a keyword `preserve_dataframe` to control whether we should convert the array to a dataframe.

dask_ml/preprocessing/data.py
```diff
@@ -915,3 +916,43 @@ def inverse_transform(self, X):
...
+        elif isinstance(X, dd.DataFrame):
+            data = X.map_partitions(self._transformer.transform)
+            columns = self._transformer.get_feature_names(X.columns)
+            XP = dd.from_dask_array(data, columns)
```
TomAugspurger: same comment on `preserve_dataframe`.

dask_ml/preprocessing/data.py
datajanko: Exactly. Sorry for not directly raising an error here. Thanks for the hint on the accept_multiple_blocks.

datajanko: Okay, added `check_array`. Btw, should I also use check_array for the dask dataframe?

dask_ml/preprocessing/data.py
datajanko: Since, this will force me to alter the doc-strings, I'll not be able to tackle the issue before the weekend. I think, I'll have enough time on Sunday

datajanko:
So what do to with the `sklearn_dev` error. The I think, this should be mergeable right now.

datajanko:
Another question: As of the contributing pipelines, dask transformers should have a columns keyword. Should this be added here as well? Maybe one could add a follow up issue.

TomAugspurger:
> As of the contributing pipelines, dask transformers should have a columns keyword. Should this be added here as well?

No, I think that is unnecessary now that ColumnTransformer is in scikit-learn.

TomAugspurger:
Fixed a merge conflict and pushed. Hopefully the CI is passing now.

tests/preprocessing/test_data.py
```diff
@@ -471,3 +471,109 @@ def test_inverse_transform(self):
...
+        assert_estimator_equal(a.fit(df), b.fit(df.compute()))
+        assert_estimator_equal(a.fit(df), b.fit(df.compute().values))
+
+    def test_estimator_comparison(self):
+        a = dpp.PolynomialFeatures()
+        b = spp.PolynomialFeatures()
+
+        a.fit_transform(X).compute()
+        b.fit_transform(X.compute())
+        assert_estimator_equal(a, b)
```
TomAugspurger: I think that this `assert_estimator_equal` can be moved down to `test_array_transform` and this test can be removed.

tests/preprocessing/test_data.py
```diff
@@ -471,3 +471,109 @@ def test_inverse_transform(self):
...
+
+    def test_array_transform(self):
+        a = dpp.PolynomialFeatures()
+        b = spp.PolynomialFeatures()
+
+        res_a = a.fit_transform(X).compute()
```
TomAugspurger: Can you remove this `.compute` and replace it with `assert is_dask_collection(res_a)`? `assert_eq_ar` will take care of computing it when we go to compare.

tests/preprocessing/test_data.py
```diff
@@ -471,3 +471,109 @@ def test_inverse_transform(self):
...
+        res_b = b.fit_transform(X.compute())
+        assert_eq_ar(res_a1, res_b)
+        assert_eq_ar(res_a2, res_b)
+
+    def test_df_transform(self):
+        a = dpp.PolynomialFeatures(preserve_dataframe=True)
+        b = spp.PolynomialFeatures()
+
+        # pandas dataframe
+        res_pdf = a.fit_transform(df.compute()).values
```
TomAugspurger: What's the behavior for a dask dataframe? Can you test that (if it isn't tested elsewhere)?

tests/preprocessing/test_data.py
```diff
@@ -471,3 +471,109 @@ def test_inverse_transform(self):
...
+        trans_dask_df = a.fit_transform(df)
+
+        # pandas dataframe
+        res_pandas_df = a.fit_transform(df.compute())
+        # spp returns a numpy array
+        res_b = b.fit_transform(df.compute())
+        assert_eq_ar(trans_dask_df.compute().values, res_b)
+        assert_eq_df(trans_dask_df.compute().reset_index(drop=True), res_pandas_df)
+
+    def test_df_dont_preserve_df(self):
```
TomAugspurger: This test feels redundant. I believe that everything here is covered elsewhere.

tests/preprocessing/test_data.py
TomAugspurger: Ah I see `test_daskdf_tranform` below. Any reason not to include that here? We could pametrize the test by `daskify`

```python
if not daskify:
    df = df.compute()

...

if daskify:
    assert dask.is_dask_collection(res_pdf)

assert_eq_ar(res_pdf, res_b)
```

tests/preprocessing/test_data.py
datajanko: It feels redundant, but actually it's not. So far I haven't considered a df as an input with an array as an output. An earlier implementation was not having that, as I was always returning a dataframe if the input was also a dataframe. I'll just add those tests in the `test_df_transform`if I'm managing to duskily that. But maybe it's too cautious.

tests/preprocessing/test_data.py
datajanko: originally, I didn't combine it because of the dataframe comparison in `test_daskdf_transform`. But I think using `daskify` will solve the issue.

datajanko:
I just got a `RuntimeError`
```RuntimeError: scikit-learn estimators should always specify their parameters in the signature of their __init__ (no varargs). <class 'dask_ml.preprocessing.data.PolynomialFeatures'> with constructor (self, preserve_dataframe=False, *args, **kwargs) doesn't  follow this convention.```

Moreover, something like
```
X, y = make_classification(chunks=50)
df = X.to_dask_dataframe().rename(columns=str)
a = dpp.PolynomialFeatures(preserve_dataframe=True)
res_df = a.fit_transform(frame)
```
seems to work, but inspecting
```
df
```
gives
<detail>
```
AttributeError                            Traceback (most recent call last)
~/anaconda3/envs/dask-ml-dev/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

~/anaconda3/envs/dask-ml-dev/lib/python3.6/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    398                         if cls is not object \
    399                                 and callable(cls.__dict__.get('__repr__')):
--> 400                             return _repr_pprint(obj, self, cycle)
    401 
    402             return _default_pprint(obj, self, cycle)

~/anaconda3/envs/dask-ml-dev/lib/python3.6/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    693     """A pprint that just redirects to the normal repr function."""
    694     # Find newlines and replace them with p.break_()
--> 695     output = repr(obj)
    696     for idx,output_line in enumerate(output.splitlines()):
    697         if idx:

~/anaconda3/envs/dask-ml-dev/lib/python3.6/site-packages/dask/dataframe/core.py in __repr__(self)
    392 
    393     def __repr__(self):
--> 394         data = self._repr_data.to_string(max_rows=5, show_dimensions=False)
    395         return """Dask {klass} Structure:
    396 {data}

~/anaconda3/envs/dask-ml-dev/lib/python3.6/site-packages/dask/dataframe/core.py in __getattr__(self, key)
   2518             return new_dd_object(merge(self.dask, dsk), name,
   2519                                  meta, self.divisions)
-> 2520         raise AttributeError("'DataFrame' object has no attribute %r" % key)
   2521 
   2522     def __dir__(self):

AttributeError: 'DataFrame' object has no attribute '_repr_data'

AttributeError                            Traceback (most recent call last)
~/anaconda3/envs/dask-ml-dev/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/anaconda3/envs/dask-ml-dev/lib/python3.6/site-packages/dask/dataframe/core.py in _repr_html_(self)
   3131 
   3132     def _repr_html_(self):
-> 3133         data = self._repr_data.to_html(max_rows=5,
   3134                                        show_dimensions=False, notebook=True)
   3135         return self._HTML_FMT.format(data=data, name=key_split(self._name),

~/anaconda3/envs/dask-ml-dev/lib/python3.6/site-packages/dask/dataframe/core.py in __getattr__(self, key)
   2518             return new_dd_object(merge(self.dask, dsk), name,
   2519                                  meta, self.divisions)
-> 2520         raise AttributeError("'DataFrame' object has no attribute %r" % key)
   2521
   2522     def __dir__(self):

AttributeError: 'DataFrame' object has no attribute '_repr_data'
```
</detail>
I though I resolved that error

**Comment by TomAugspurger** [comment]
> I just got a RuntimeError
> RuntimeError: scikit-learn estimators should always specify their parameters in the signature of their __init__ (no varargs). <class 'dask_ml.preprocessing.data.PolynomialFeatures'> with constructor (self, preserve_dataframe=False, *args, **kwargs) doesn't follow this convention.


Sorry I missed that. That's because you're using `*args, **kwargs` in `PolynomialFeatures.__init__`. We need to list out the keyword arguments explicitly.

**Comment by datajanko** [comment]
> > I just got a RuntimeError
> > RuntimeError: scikit-learn estimators should always specify their parameters in the signature of their **init** (no varargs). <class 'dask_ml.preprocessing.data.PolynomialFeatures'> with constructor (self, preserve_dataframe=False, *args, **kwargs) doesn't follow this convention.
> 
> Sorry I missed that. That's because you're using `*args, **kwargs` in `PolynomialFeatures.__init__`. We need to list out the keyword arguments explicitly.

Sorry for mentioning that I'm aware of that. Next two days are a but busy. I hope I'll find some time on Thursday or Friday to solve the last remaining issues and modify the tests accordingly.

Moreover, I'll add a test that ensures that a dataframe can be inspected.

**Comment by datajanko** [comment]
I found the "bug" in dask. `dd.from_dask_array` does not like multiple column names. The test dataframe contains a `1`which is the same column description as that of the bias.
So a hot fix is to ensure that the columns of the dataframe don't contain the `1` and maybe raise a warning. Other suggestions?

I'll check if dask already has an issue concerning that

**Comment by TomAugspurger** [comment]
> So a hot fix is to ensure that the columns of the dataframe don't contain the 1 and maybe raise a warning. Other suggestions?

that seems fine.

**Comment by mrocklin** [comment]
The upstream issue seems pretty solvable as well if anyone wants to take a
shot.

**Comment by datajanko** [comment]
So I'd suggest that I try to solve the issue in dask, as @mrocklin suggested. If I'm not able to fix this quickly, I'd provide a solution with the hot fix. Would this be okay, or should we rather close this issue first with the hot fix and create a new task to remove the fix once dask has the "bug" fixed?

**Comment by datajanko** [comment]
Any hints on the coverage bug? Should I try to increase the coverage?

**Comment by mrocklin** [comment]
I think it's not actually coverage, I think that it's a flake issue.  I recommend running `flake8` on your codebase and seeing if anything comes up.

cc @TomAugspurger feedback on the black process ^^  It's not entirely clear to users what went wrong when things go wrong, or how to fix it.

**Comment by datajanko** [comment]
Hmm i ran black and pushen the changes. Will Check tomorrow morning again and push the Update. Thanks for your comments

**Comment by mrocklin** [comment]
```
[black]
would reformat /root/project/dask_ml/preprocessing/__init__.py
All done! 💥 💔 💥
1 file would be reformatted, 92 files would be left unchanged.
```

datajanko:
Damn, i rebased and did not run black again. Maybe it's too late today. Will push the changes tomorrow. Thanks again

datajanko:
@TomAugspurger and @mrocklin I think this could be merged right now. Any further comments? With my latest addition in dask, the representation bug is also gone. So assuming one hast the nightly dask, this will work without issues. Is there any suggestion how/if to make users aware of that?

TomAugspurger:
Looks good, thanks @datajanko.

Dask-ML tends to require fairly recent versions of dask. We'll likely bump the required version of dask shortly after the next release. But since it's just a bug in the repr I don't think we're in a rush.

PR #370, opened 2018-09-19
API: Incremental search improvements

stsievert:
This adds "CV" to `BaseIncrementalSearch` and it's children. `BaseIncrementalSearchCV` does do some (very basic) cross validation.

This PR

* ~~API: rename `IncrementalSearch` to `TimeDecaySearchCV`~~
* API: renames `*Search` to `*SearchCV`
* API: renames `history_results_` to `history_`
* API: adds `cv_results_` to `BaseAdaptiveSearchCV`.
* MAINT: selects best model by highest score at last partial fit call
* ~~MAINT: adds private `AdaptiveStopOnPlateauSearchCV` to all easier implementation of adaptive algorithm that want to stop on plateau (e.g., #221).~~
* DOC: clarify definition "best" (which also discussed to be highest final score) in docstring
* DOC: Add "Attributes" section to AdaptiveSearchCV in docstring
* ~~DOC: clarify use of "adaptive" and "incremental" in docs~~

~~I also tried to come up with a better name for `IncrementalSearchCV`: it seems like a very general name (but I didn't think about it too hard).~~


TomAugspurger:
My main motivation for not using the CV suffix was that it doesn't support the `cv` parameter found in `RandomizedSearchCV` and `GridSearchCV`.

stsievert:
> My main motivation for not using the CV suffix was that it doesn't support the cv parameter found in RandomizedSearchCV and GridSearchCV.

Hm... My motivation for changing to `*SearchCV` was because it does a search with a validation set. It does not score models from the training set.

I've made some other changes too, summarized in https://github.com/dask/dask-ml/pull/370#issue-216493783.

mrocklin:
I think that this is probably the sort of thing we should discuss before anyone invests significant work in the renaming.  My guess is that we might cycle through a dozen names before finding the right one.  

I agree that incremental isn't great, but I also think that decay is probably also not ideal.

TomAugspurger:
On the base of the name, Adaptive is certainly descriptive, but is it too general?

About CV, I'm I don't have a strong opinion either way. The thing that pushed me to remove it was `GridSearchCV(..., cv=1)` raising an error.

```pytb
In [26]: KFold(1)
<ipython-input-26-f1793b6f886b> in <module>()
----> 1 KFold(1)

~/Envs/pandas-dev/lib/python3.6/site-packages/scikit-learn/sklearn/model_selection/_split.py in __init__(self, n_splits, shuffle, random_state)
    417     def __init__(self, n_splits=3, shuffle=False,
    418                  random_state=None):
--> 419         super(KFold, self).__init__(n_splits, shuffle, random_state)
    420
    421     def _iter_test_indices(self, X, y=None, groups=None):

~/Envs/pandas-dev/lib/python3.6/site-packages/scikit-learn/sklearn/model_selection/_split.py in __init__(self, n_splits, shuffle, random_state)
    282                 "k-fold cross-validation requires at least one"
    283                 " train/test split by setting n_splits=2 or more,"
--> 284                 " got n_splits={0}.".format(n_splits))
    285
    286         if not isinstance(shuffle, bool):

ValueError: k-fold cross-validation requires at least one train/test split by setting n_splits=2 or more, got n_splits=1.
```

TomAugspurger:
Sorry, missed this notification earlier.

`history_` -> `history_results_` seems good. The idea was to mirror `cv_results_` but I suppose scikit-learn needed to differentiate from the `cv` parameter. We don't have that concern with history.

> selected the best model by highest score at last partial fit call

I think that was already being done.

> This required some changes to always get the most recent score (with the highest partial fit call) from history_. 

idle question: do we know that partial_fit_score is monotonically increasing for a given model through history? I don't think we can assume that later = higher (not saying that you do. Haven't looked at the code yet).

TomAugspurger:
My reasons for not implementing cv_results originally were

1. simplicity
2. us not supporting the `cv` parameter.

So going back to me not wanting `CV` suffix because we don't support the `cv` parameter. Do you have thoughts on @stsievert? Is `Search` expressive enough on its own that we're doing hyperparmeter optimization

dask_ml/model_selection/_incremental.py
```diff
@@ -476,26 +476,55 @@ def _get_history_results(self, results):
...
-        key = operator.itemgetter("model_id")
-        best_index = -1
-        # history_results is sorted by (model_id, partial_fit_calls)
-        # best is the model_id with the highest partial fit calls
-        for k, v in itertools.groupby(history_results, key=key):
-            v = list(v)
-            best_index += len(v)
-            if k == best_model_id:
-                break
+        for h in hist2:
```
TomAugspurger: Should this just be done by `_fit`?

dask_ml/model_selection/_incremental.py
```diff
@@ -476,26 +476,55 @@ def _get_history_results(self, results):
...
+            if h["model_id"] not in info:
+                info[h["model_id"]] = []
+            info[h["model_id"]] += [h]
+        return hist2, info
+
+    def LD_to_DL(self, LD):
+        """ list of dicts to dict of lists. Assumes same keys in all dicts.
+        https://stackoverflow.com/questions/5558418/list-of-dicts-to-from-dict-of-lists
+        """
+        return {k: [dic[k] for dic in LD] for k in LD[0]}
```
TomAugspurger: style nit: easier to inline below, especially if we're using it only once.

dask_ml/model_selection/_incremental.py
```diff
@@ -476,26 +476,55 @@ def _get_history_results(self, results):
...
+                "test_score": h["score"],
+                "rank_test_score": sum(scores >= h["score"]),
+                "model_id": h["model_id"],
+                "params": h["params"],
+                "partial_fit_calls": h["partial_fit_calls"],
+            }
+        cv_results = list(cv_results.values())  # list of dicts
+        cv_results = self.LD_to_DL(cv_results)
+        cv_results = {k: np.array(v) for k, v in cv_results.items()}
+        return cv_results
```
TomAugspurger: What does each record here indicate? For e.g. `GridSearchCV`, looking at the first element from each list (so `pd.DataFrame(cv_results).iloc[0]`) is the all the stats for a single hyper parameter combination.

Ours would have the same meaning, correct?

stsievert: Correct. `cv_results_` holds certain variables (`params`, `test_score`, `mean_partial_fit_time`, etc) for a particular parameter combo. All time-dependent variables (`test_score`, etc) are for the last time the model is evaluated.

TomAugspurger: Good deal. In that case, it would be good to have this available to users.

stsievert:
> So going back to me not wanting CV suffix because we don't support the cv parameter. Do you have thoughts on stsievert? Is `Search` expressive enough on its own that we're doing hyperparmeter optimization

I do not think `Search` is expressive enough; to me `*Search` indicates "train and test on the same data" and `*SearchCV` indicates "train and test on different data". This class does perform (basic) cross validation because it creates separate datasets to test and train on. This also underpins the naming of `cv_results_`.

The `cv` parameter specifies the amount of cross validation for a particular cross validation scheme, `KFold` (and does not specify whether or whether not to use cross validation). This class could also the vary the amount of cross validation with the something like

``` python
search = AdaptiveSearchCV()
datasets = [train_test_split(X, y) for _ in range(cv)]
scores = [search.fit(X1, y1).score(X2, y2)
          for X1, X2, y1, y2 in datasets]
```

though I don't think this should be implemented right now (one train/test set is used in practice).

> idle question: do we know that partial_fit_score is monotonically increasing for a given model through history? I don't think we can assume that later = higher (not saying that you do. Haven't looked at the code yet).

Correct – monotonically increasing scores is a very bad assumption (though that's almost the case when `patience=1` and `tol=0`).

I think we should choose `best_estimator_` based off the maximum final score, not the maximum score after any `partial_fit_call`. Two cases where reporting the maximum score on any `partial_fit` breaks down:

* maybe an estimator receives a high score on a particular `partial_fit` call because the same data is fed to `partial_fit` and `score`.
* maybe a particular estimator overfits the data. At one point, this model could have had the best score for any model. However, it could converges to the lowest possible (finite) score.

The protection against `score` decreasing too much is to stop training on plateau, which is specified by the `patience` and `tol` parameters (though allowing some decrease of `score` may be required to reach a better solution).

Some of this will be fixed by https://github.com/dask/dask-ml/pull/370#discussion_r219328288, but not all of it.

>> selected the best model by highest score at last partial fit call
>
> I think that was already being done.

`best_model_id` was selected to be the model returned by `_fit` without looking at any `score`s https://github.com/dask/dask-ml/pull/370/files#diff-0a54ded74c8013caf0588e9a5c879e99L487.

dask_ml/model_selection/_incremental.py
stsievert: I've made the change which is makes `info` have the same data as `history`, but in a different format. This breaks a test:

https://github.com/dask/dask-ml/blob/b5b6954cbd8a4da89aa4da9612e970a0680290da/tests/model_selection/test_incremental.py#L181

The relevant bits of my diff are

``` diff
- assert len(info) == len(models) == 1
+ assert len(info) > len(models) == 1
+ assert set(models.keys()).issubset(set(info.keys()))
+
+ calls = {k: [h["partial_fit_calls"] for h in hist] for k, hist in info.items()}
+ for k, call in calls.items():
+     assert (np.diff(call) >= 1).all()

```

This change provides some ease-of-use and another test. Do we want to make this change?

TomAugspurger: Can you post what `info` looks like now? Breaking tests is fine for now, if we think it's a better data format.

TomAugspurger:
> though I don't think this should be implemented right now (one train/test set is used in practice).

Agreed, though it's good to know that this could be done in the future.

> best_model_id was selected to be the model returned by _fit without looking at any scores https://github.com/dask/dask-ml/pull/370/files#diff-0a54ded74c8013caf0588e9a5c879e99L487.

IIUC, at that point `info` had just a single element, though I may be mistaken. But this may be moot given your changes to `info`.

dask_ml/model_selection/_incremental.py
```diff
@@ -476,26 +481,49 @@ def _get_history_results(self, results):
...
+                "model_id": h["model_id"],
+                "params": h["params"],
+                "partial_fit_calls": h["partial_fit_calls"],
+            }
+        cv_results = list(cv_results.values())  # list of dicts
+        cv_results = {k: [res[k] for res in cv_results] for k in cv_results[0]}
+        cv_results = {k: np.array(v) for k, v in cv_results.items()}
+        return cv_results
+
+    def _get_best(self, results, cv_results, history):
```
TomAugspurger: We should document our definition of "best" here, since there are at least two reasonable definitions:

a. the model with the highest final `fit_score`
b. the model with the highest `fit_score` at any point in `history`

TomAugspurger: Also document parameters, return types since this is a bit more complex now.

dask_ml/model_selection/_incremental.py
```diff
@@ -476,26 +481,49 @@ def _get_history_results(self, results):
...
+                "test_score": h["score"],
+                "rank_test_score": sum(scores >= h["score"]),
+                "model_id": h["model_id"],
+                "params": h["params"],
+                "partial_fit_calls": h["partial_fit_calls"],
+            }
+        cv_results = list(cv_results.values())  # list of dicts
+        cv_results = {k: [res[k] for res in cv_results] for k in cv_results[0]}
+        cv_results = {k: np.array(v) for k, v in cv_results.items()}
+        return cv_results
```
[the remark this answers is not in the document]
stsievert: Oh, good catch. Thanks.

stsievert:
@TomAugspurger and I talked about this PR, and generated a couple more TODOs:

- [ ] API: `DecaySearchCV` => `TimeDecaySearchCV` (related: [`time_inverse_decay`][1])
- [ ] API: set `param__*` keys in `cv_results_`
- [x] API: keep the name `AdaptiveSearchCV`
- [ ] DOC: clarify definition "best" (which also discussed to be highest final score)
- [ ] DOC: document how to choose `patience` and `tol` (probably by example)
- [ ] DOC: Add "Attributes" section to `AdaptiveSearchCV` 


[1]:https://www.tensorflow.org/api_docs/python/tf/train/inverse_time_decay

TomAugspurger:
Many :) Some considered were IncrementalSearch (the same), DecaySearch, ExponentialSearch, ExponentialDecaySearch, TimeInverseDecaySearch, probably others.

We want to convey the most important aspect of this strategy, the decaying number of models trained as we moved through time. TimeDecaySearch seemed to strike a reasonable balance between accuracy and length.

As for CV, the train-test split we do internally is a form of cross validation, so it's best to include the suffix.

mrocklin:
Note that this is also currently what backs RandomSearch, which doesn't significantly decrease the number of models over time (except when they seem to plateau)

dask_ml/model_selection/_incremental.py
```diff
@@ -652,13 +665,17 @@ class IncrementalSearch(BaseIncrementalSearch):
         If ``patience`` is used the maximum number of ``partial_fit`` calls
         between ``score`` calls.
 
-    tol : float, default 0.001
+    tol : float, default -0.01
```
TomAugspurger: Why flip the sign here? IIRC, these are typically expressed as positive numbers.

dask_ml/model_selection/_incremental.py
```diff
@@ -652,13 +665,17 @@ class IncrementalSearch(BaseIncrementalSearch):
...
         The required level of improvement to consider stopping training on
         that model. The most recent score must be at at most ``tol`` better
         than the all of the previous ``patience`` scores for that model.
         Increasing ``tol`` will tend to reduce training time, at the cost
         of worse models.
 
+        Setting this value to be negative allows for some noise tolerance in
```
TomAugspurger: I think this doesn't sit with the previous paragraph (maybe the previous paragraph is wrong).

It sounds like `tol` is influencing the question "should we stop?". Something like  `if current_score - previous_score > tol: stop`. Setting this to negative would mean we'll consider stopping on a model that performed worse than the previous one, while with positive the final model must always be better than the previous `patience` (until we hit `max_iter`).

I also can't tell what we mean by "more accurate estimator", or why that would be. I can imagine it being because we're not overfitting as much, but I can also imagine other explanations.

Looking at the code, we say `if all(current_score < old + tol for old in patience)`. So a negative tol makes that condition more likely to be true, i.e. we continue training even when our score has dipped. So I think the first paragraph explaining `tol` is incorrect?

docs/source/hyper-parameter-search.rst
```diff
@@ -179,28 +179,29 @@ tasks in the graph and only perform the fit step once for any
...
 
 .. autosummary::
-   dask_ml.model_selection.IncrementalSearch
+   dask_ml.model_selection.DecaySearchCV
 
-Broadly speaking, incremental optimization starts with a batch of models (underlying
-estimators and hyperparameter combinationms) and repeatedly calls the underlying estimator's
-``partial_fit`` method with batches of data.
+
+These model selection algorithms are *adaptive*: they decide to keep training
```
TomAugspurger: I think we should split these two ideas a bit more cleanly in the docs:

1. Adaptive: Using previous fit information to decide which models to train next
2. Incremental: Training on large datasets by incrementally passing batches to `partial_fit`.

As written, this paragraph implies that both of these are necessary, when you could imagine Incremental without Adaptive, and Adaptive without Incremental.

I'm going back a bit on what we discussed yesterday. Our implementation *requires* Incremental (it only works with underlying estimators that implement `partial_fit`). So to me that's the most important of the two, and if we're only going to highlight one (e.g. the section title) it should be incremental.

docs/source/hyper-parameter-search.rst
```diff
@@ -179,28 +179,29 @@ tasks in the graph and only perform the fit step once for any
...
+estimators based off all previously collected information. As such, these
+estimators must implement ``partial_fit`` to train on batches of data.
 
 .. note::
 
    These estimators require the optional ``distributed`` library.
 
 Here's an example training on a "large" dataset (a Dask array) with the
-``IncrementalSearch``
+``DecaySearchCV``
```
TomAugspurger: These need to be updated for the new name.

dask_ml/model_selection/_incremental.py
```diff
@@ -692,6 +709,72 @@ class IncrementalSearch(BaseIncrementalSearch):
...
+        * ``mean_partial_fit_time``
+        * ``mean_score_time``
+        * ``std_partial_fit_time``
+        * ``std_score_time``
+        * ``test_score``
+        * ``rank_test_score``
+        * ``model_id``
+        * ``partial_fit_calls``
+        * ``params``
+        * ``param_{key}`` for every ``key`` in the parameters for that model.
```
TomAugspurger: Rather than "parameters for that model", should it be "hyperparameters in `params`"?

dask_ml/model_selection/_incremental.py
```diff
@@ -692,6 +709,72 @@ class IncrementalSearch(BaseIncrementalSearch):
...
+        * ``model_id``
+        * ``params``
+        * ``partial_fit_calls``
+
+        The key ``model_id`` corresponds to the ``model_id`` in ``cv_results_``.
+        This list of dicts can be imported into Pandas.
+
+    best_score_ : float
+        Best score achieved on the hold out data, where "best" means "highest
+        score after a models final partial fit call".
```
TomAugspurger: missing apostrophe in "model's".

TomAugspurger: Is the explanation "highest final score on the validation set" clearer?

dask_ml/model_selection/_incremental.py
stsievert: Or "where `key` is every key in `params`"?

stsievert: Scikit-learn gives an example `cv_results_` in their docs, and don't address this generally.

dask_ml/model_selection/_incremental.py
stsievert: I avoided that because I wanted to define "final". I've changed it to

>  where "best" means "the highest score on the hold out set after a model's last partial fit call."

docs/source/hyper-parameter-search.rst
stsievert: Updated.

docs/source/hyper-parameter-search.rst
stsievert: I agree with you; I'm also rethinking naming. I agree that `Incremental` is more important to highlight than `Adaptive`; all adaptive algorithms must be incremental, but not all incremental algorithms are adaptive.

TODO:
- [ ] rename `BaseAdaptiveSearchCV` => `BaseIncrementalSearchCV`

dask_ml/model_selection/_incremental.py
stsievert: I set it negative to stop training when the scores starts to decrease. I'd expect to stop when performance starts to degrade, not when performance doesn't increase enough.

I guess my expectations are wrong; PyTorch's [`ReduceLROnPlateau`][1] performs it's action when `all(old > best + threshold for old in plateau)` for some positive `threshold`. That is, it stops increasing when models haven't improved enough in `patience` epochs.

TODO:

- [ ] default `tol=1e-4` (PyTorch's `ReduceLROnPlateau` `threshold`).

[1]:https://pytorch.org/docs/stable/optim.html#torch.optim.lr_scheduler.ReduceLROnPlateau

dask_ml/model_selection/_incremental.py
```diff
@@ -654,13 +655,15 @@ def _get_params(self):
             return ParameterSampler(self.parameters, self.n_initial_parameters)
 
     def _additional_calls(self, info):
-        best = self._adapt(info)
-        bad = set(info) - set(best)
+        self._all_models.update(set(info.keys()))
+        instructions = self._adapt(info)
+        bad = set(self._all_models) - set(instructions)
```
stsievert: I'm getting a bug with this sometimes. This commit attempts to resolve it. On occasion, I get a `KeyError` in `_incremental.py` with `speculative.pop(ident)` at https://github.com/dask/dask-ml/blob/37eca7d335509c2a4aa9332aa454f57092318487/dask_ml/model_selection/_incremental.py#L230  

which I take to mean some models are being killed early. This happens `bad == set()` at https://github.com/dask/dask-ml/blob/37eca7d335509c2a4aa9332aa454f57092318487/dask_ml/model_selection/_incremental.py#L212

dask_ml/model_selection/_incremental.py
```diff
@@ -654,13 +655,15 @@ def _get_params(self):
...
         # Keep models in case they quickly reach a high plateau (yes, unlikey)
-        best.update({model_id: 0 for model_id in bad})
+        instructions.update({model_id: 0 for model_id in bad})
+
         out = {}
-        for k, steps in best.items():
+        for k, steps in instructions.items():
             records = info[k]
             if self.max_iter and len(records) >= self.max_iter:
                 out[k] = 0
```
stsievert: TODO: resolve the selection of the beginning of the plateau. Right now, the code assumes every records corresponds to 1 partial fit call. This may not be true if adapt is general. Case in point: `old = records[-self.patience]["score"]`.

dask_ml/model_selection/_incremental.py
```diff
@@ -256,6 +256,11 @@ def get_futures(partial_fit_calls):
 
     models = {k: client.submit(operator.getitem, v, 0) for k, v in models.items()}
     yield wait(models)
+    info = {}
```
TomAugspurger: Slight preference for

```
info = defaultdict(list)
for h in history:
    info[h["model_id"]].append(h)

info = dict(info)
```

dask_ml/model_selection/_incremental.py
```diff
@@ -599,20 +614,94 @@ def score(self, X, y=None):
         return self.scorer_(self.best_estimator_, X, y)
 
 
-class IncrementalSearch(BaseIncrementalSearch):
+class AdaptiveStopOnPlateauSearchCV(BaseIncrementalSearchCV):
```
TomAugspurger: Inconsistent with the name in the `__init__.py`.

docs/source/hyper-parameter-search.rst
```diff
@@ -15,7 +15,7 @@ The underlying estimator will need to be able to train on each cross-validation
 See :ref:`hyperparameter.drop-in` for more.
 
 If your data is large and the underlying estimator implements ``partial_fit``, you can
-Dask-ML's :ref:`*incremental* hyperparameter optimizers <hyperparameter.incremental>`.
```
TomAugspurger: I'm starting to doubt the "two kinds of hyperparameter optimization estimators" I laid out on line 6.

After our conversation the other day, it sounds like there are (at least) 3.

1. Static search: All combinations to be tried are specified up front. The full dataset (or a CV split of the full dataset) is used on each training call.
2. Adaptive search: Regardless of exactly which data is used (CV split of full dataset, or a batch), the salient feature of adaptive search is that models are prioritized according to past performance. 
3. Incremental search: Regardless of how models are chosen for training (static or adaptive), the salient feature of Incremental search is that models are trained on *batches* of data passed to `_parital_fit`. The underlying estimator *must* implement `partial_fit`.

Our current implementation is a mix of 2 and 3. Does that framing accord with other's thoughts? Can we think of a clear way to explain that? The docs below do a decent job I think. This line, though, I think emphasizes adaptive too much given the framing above (you have a large dataset).

dask_ml/model_selection/_incremental.py
TomAugspurger: @mrocklin any thoughts here? I've seen this occasionally to, but haven't been able to debug it successfully.

dask_ml/model_selection/_incremental.py
mrocklin: -1 on this name.  I think that it is far too long and technical.  I think that it would scare non-techncial users.

dask_ml/model_selection/_incremental.py
mrocklin: No thoughts off-hand.  Is anyone able to produce a minimal example by any chance?

dask_ml/model_selection/_incremental.py
stsievert: `AdaptiveStopOnPlateauSearchCV` is private (or at least that's what I intend). It is intended for use by other developers who have an adaptive algorithm they'd like to implement and also have stop on plateau (e.g., #221).

Here's the structure I intended:

* `BaseIncrementalSearchCV`: private.
    * Use: base class for all incremental searches
* `AdaptiveStopOnPlateauSearchCV`: private, inherits from `BaseIncrementalSearchCV`
    * Use: allow other devs easily create adaptive algorithms that stop on plateau (by overriding `_adapt`)
* `InverseTimeDecaySearchCV`: public, inherits from `AdaptiveStopOnPlateauSearchCV`

mrocklin: I think that something like the implementation of `IncrementalSearch` in master should be public.  I think that `InverseTimeDecaySearchCV` is also probably an overly-technical name.

I recommend that we open an issue where we can discuss name changes to what is in master today and future restructurings.  In general I recommend doing this before investing much work.

TomAugspurger: Indeed. The changes to `info` and `cv_results_`, etc. are useful in their own right. Those shouldn't be held up by a naming discussion.

docs/source/hyper-parameter-search.rst
stsievert: Maybe a clearer way of explaining it:

> There are two dimensions to searches:
>
> 1. Does the search use `partial_fit` or `fit`?
> 2. Is the search passive or adaptive? That is, does it use previous evaluations to select which models to train further?

Your categories fall pretty nicely into this (maybe renaming "static" to "passive").

dask_ml/model_selection/__init__.py
```diff
@@ -19,8 +19,8 @@
 
 
 try:
```
TomAugspurger: Needs to be reverted too (or changed to `IncrementalSearchCV`)

dask_ml/model_selection/_incremental.py
```diff
@@ -599,20 +613,94 @@ def score(self, X, y=None):
...
```
TomAugspurger: Same, this needs to be changed back, or just changed to IncrementalSearchCV.

stsievert: This diff line isn't completely accurate at first glance (`IncrementalSearch` was not renamed to `AdaptiveStopOnPlateauSearchCV`). I added a private class `AdaptiveStopOnPlateauSearchCV` to allow other adaptive algorithms to stop on plateau. I only intend this class to be used Dask-ML developers, not by Dask-ML users.

mrocklin: I think that @TomAugspurger 's comment was aimed to try to scope down this PR to just the `info` and `cv_results_` changes so that we can them in.  Any change to the class hierarchy here will probably require more discussion before a good plan is reached.

mrocklin: If you have time to separate the changes @stsievert I suspect that that would be appreciated.

stsievert: Ah. Got it. `AdaptiveStopOnPlateauSearchCV` is removed.

dask_ml/model_selection/_incremental.py
TomAugspurger: Could you remind me what we decided here? I think that I was in favor of `tol` being positive by default, since that aligns with the default in scikit-learn.

TomAugspurger: Regardless of the default, it'd be good to note that `tol<0` is well defined (we're allowed to stop on a model that does not too much worse that the one before it).

edit: Ah, good I see that you have that below.

dask_ml/model_selection/_incremental.py
```diff
@@ -715,16 +796,33 @@ class IncrementalSearch(BaseIncrementalSearch):
...
-    ...                            n_initial_parameters=1000,
-    ...                            patience=20, max_iter=100)
+    >>> search = IncrementalSearchCV(model, params, random_state=0,
+    ...                              n_initial_parameters=1000,
+    ...                              patience=20, max_iter=100)
+
+    ``tol`` and ``patience`` are for stopping training on a "plateau", or when
+    the hold-out score flattens and/or falls. For example, setting ``tol=0``
+    and ``patience=2`` will dictate that scores are increasing (except for
+    potentially the last score).
```
TomAugspurger: Should the parenthetical be "except for potentially when `max_iter` is reached"? That's a bit more descriptive.

dask_ml/model_selection/_incremental.py
```diff
@@ -747,7 +845,8 @@ def __init__(
         self.tol = tol
         self.scores_per_fit = scores_per_fit
         self.max_iter = max_iter
-        super(IncrementalSearch, self).__init__(
+        self._all_models = set()
```
TomAugspurger: Could you explain why this is needed?

dask_ml/model_selection/_incremental.py
stsievert: Whoops, I need to make the default `tol` positive. PyTorch's `ReduceLROnPlateau` also defaults to a postive `tol` (IIRC, I linked to the source elsewhere).

TODO: make tol positive again

dask_ml/model_selection/_incremental.py
stsievert: To keep models in memory, so the model with the best score (which may not be the last model used by `_fit`; e.g., it may have plateaued early). `_all_models` holds the IDs of all models and returns them from `_additional_calls`, so `_fit` doesn't remove those models.

The relevant lines are

https://github.com/dask/dask-ml/blob/76a6482d3a2fc87514caa670777e58d5dbb7a672/dask_ml/model_selection/_incremental.py#L211-L218

mrocklin: Keeping all models in memory may not be ideal.  In cases where we've plateaued do we know that a higher score on the validation data actually implies a better model?  

Alternatively, maybe we just hold onto the best that we've seen so far rather than all models?  It would be nice to let things clear out.

stsievert: I think it'd suffice to hold onto one other model to protect against the mentioned case. I'll put in some work for that, hopefully tonight.

dask_ml/model_selection/_incremental.py
stsievert: I was covering that when `patience=2` and `tol=0`, only the last score is decreasing if the model does plateau. I tried to clarify that with:

> ``tol`` and ``patience`` are for stopping training on a "plateau", or when
the hold-out score flattens and/or falls. For example, setting ``tol=0``
and ``patience=2`` will dictate that for estimators that never plateau,
scores will always increase until ``max_iter`` calls are reached.
If it the estimator plateaus, only the last score will decrease.

Let me know what edits you have.

dask_ml/model_selection/_incremental.py
stsievert: The code holds onto fewer models now with 7de8f6a6061ed6b742a5179dd9308686ed0cd3a5.

dask_ml/model_selection/_incremental.py
```diff
@@ -757,7 +859,58 @@ def _get_params(self):
         else:
             return ParameterSampler(self.parameters, self.n_initial_parameters)
 
+    def _stop_on_plateau(self, info, instructions):
+        _best_plateau = {}
```
TomAugspurger: I'm probably missing something, but can you walk through how a model with `patience=0` works? From what I can see, it looks like all the updates to `_best_plateau` are done inside `if self.patience` blocks. So for no patience you'll never update `_best_plateau`.

If you never update `_best_plateau` here, then you'll return an empty set to `_additional_calls`.

Ah, but I see now that `_additional_calls` falls back on the instructions of `out` and `steps`, so hopefully that takes care of decaying us to 0 additional fit calls. Is that correct? I'll leave this comment here, but I think we're OK.

TomAugspurger: Also, style nit: this is a function, so you don't need to prefix variable names with `_` to indicate privacy.

mrocklin:
I'm slightly concerned that PR double the length of the logic in `_additional_calls`.  Was this change necessary for new functionality, or is it only a refactor?

dask_ml/model_selection/_incremental.py
```diff
@@ -757,7 +859,58 @@ def _get_params(self):
...
+            current_calls = records[-1]["partial_fit_calls"]
+            if self.max_iter and current_calls >= self.max_iter:
+                out[k] = 0
+            elif self.patience and current_calls >= self.patience:
+                plateau = [
+                    h["score"]
+                    for h in records
+                    if current_calls - h["partial_fit_calls"] <= self.patience
+                ]
+                if all(score <= plateau[0] + self.tol for score in plateau):
```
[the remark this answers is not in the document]
stsievert: I went with `patience[0]` because it's PyTorch uses `patience[0]` for their `ReduceLROnPlateau` ([source][1]).

Does it matter? I don't think so, but might be wrong (haven't worked an example yet). Both methods are seeing if the most recent scores are within `tol` of each other.

[1]:https://github.com/pytorch/pytorch/blob/294c06538416f5882e9f329695dcae7ff77c218b/torch/optim/lr_scheduler.py#L353-L357

dask_ml/model_selection/_incremental.py
```diff
@@ -757,7 +859,58 @@ def _get_params(self):
...
+
+        out, _best_plateau = self._stop_on_plateau(info, instructions)
+
+        # Hold onto the model that has performed the best, even if it's
+        # plateaud early
+        if (len(self._best_plateaus) == 0) or (
+            _best_plateau
+            and max(_best_plateau.values()) >= min(self._best_plateaus.values())
+        ):
+            self._best_plateaus.update(_best_plateau)
```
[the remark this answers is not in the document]
stsievert: I used `len(self._best_plateau)` to avoid `min` over an empty list later in the condition.

dask_ml/model_selection/_incremental.py
```diff
@@ -757,7 +859,58 @@ def _get_params(self):
...
+        for k, steps in instructions.items():
+            records = info[k]
+            current_calls = records[-1]["partial_fit_calls"]
+            if self.max_iter and current_calls >= self.max_iter:
+                out[k] = 0
+            elif self.patience and current_calls >= self.patience:
+                plateau = [
+                    h["score"]
+                    for h in records
+                    if current_calls - h["partial_fit_calls"] <= self.patience
```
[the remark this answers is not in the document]
stsievert: It is if every `partial_fit` call is coupled with a `score` call, but it may not be in general. It is for TimeDecaySearchCV, but it may not be for different implementations of `_adapt`.

stsievert:
> I'm slightly concerned that PR double the length of the logic in _additional_calls. Was this change necessary for new functionality, or is it only a refactor?

For new functionality. The functionality it adds is

* allowing overriding `_adapt` to provide new adaptive methods, potentially more complex
* has fancier

dask_ml/model_selection/_incremental.py
stsievert: I think your interpretation is correct. Most of it relies on 

> that `_additional_calls` falls back on the instructions of out and steps

dask_ml/model_selection/_incremental.py
```diff
@@ -747,7 +848,8 @@ def __init__(
         self.tol = tol
         self.scores_per_fit = scores_per_fit
         self.max_iter = max_iter
-        super(IncrementalSearch, self).__init__(
+        self._best_plateaus = {}
```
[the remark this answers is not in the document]
stsievert: > Actually, looking at the implementation it seems like self._best_plateaus is only used in _additional_calls, so could you just define it there?

Good point. Implemented.

> What was the motivation for splitting _additional_calls into multiple methods?

Addressed in https://github.com/dask/dask-ml/pull/370#issuecomment-428399420

dask_ml/model_selection/_incremental.py
```diff
@@ -222,19 +222,20 @@ def get_futures(partial_fit_calls):
         _scores = {}
         _specs = {}
         for ident, k in instructions.items():
+            if not k:
+                continue
```
stsievert: I don't think this change is necessary. I need some time to clarify my thoughts on this.

TomAugspurger:
> change plateau definition from "the number of score calls" to "the number of partial_fit calls"

If [my understanding](
https://github.com/dask/dask-ml/pull/370#discussion_r224035160) is correct, which it may not be, I think this is a better concept for users to have in their head. "Number of partial fit calls" feels more concrete". But as I said in that comment, it's not clear to me that the previous way was different.

> return the highest scoring model after the last partial fit call (and always return it, even if plateaued early or score decreases)

I didn't realize we had the caveat of "after the last partial fit call". Suppose patience is 2 and tol is `0.9`. With the following score history:

- `[10, 11, 10, 10.5, 10, 10.5, 10, 10.5, 10.5, 10.5]`

Do we return the model that scored `11` (the second) or `10.5` (the last one)?

dask_ml/model_selection/_incremental.py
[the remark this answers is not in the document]
stsievert: > we could send the instructions "model 1 needs 2 additional calls", so we would have a kind of "gap" in our history where we only see the score for partial fits n and n + 2? Is that correct?

Correct.

> in theory

#221 does this: `_additional_calls` tends to specify more than 1 `partial_fit` call.

I suppose a cleaner implementation would be

``` python
plateau = [h["score"] for h in self.history[-self.patience:] if "score" in h]
```

dask_ml/model_selection/_incremental.py
stsievert: I've clarified my thoughts, and I'll try to implement them tonight. They are simple and put most of the responsibility for choosing a high performing model on `_adapt`.

There are two dimensions in `cv_results_["rank_test_score"]`: the final scores, and the time the model was scored at. This might change one of the tests:

``` diff
- assert search.cv_results_["rank_test_score"][search.best_index_] == 1
+ assert search.cv_results_["rank_test_score"][search.best_index_] <= 3  # 20 total models
```

dask_ml/model_selection/_incremental.py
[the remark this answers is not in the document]
TomAugspurger: So AFAICT, in https://github.com/dask/dask-ml/blob/90fbe69771926b7e43a5bef99d81e268dad63e7e/dask_ml/model_selection/_incremental.py#L231-L233 is the only place where we do something once *per partial_fit_call*. Everything else looks like it's done once per `_fit` call. So right now I don't see how we would end up with an entry in history that doesn't have a score. It seems like history will necessary only report models and scores that are the last in a "batch".

dask_ml/model_selection/_incremental.py
```diff
@@ -757,7 +859,58 @@ def _get_params(self):
...
+                    out[k] = steps
+
+            else:
+                out[k] = steps
+        return out, best_plateau
+
+    def fit(self, *args, **kwargs):
+        self._best_plateaus = {}
+        self._model_scores = {}
+        self._models_kept = []
```
mrocklin: What are these values for?  Can you add a small comment to help other maintainers understand them in the future?

dask_ml/model_selection/_incremental.py
```diff
@@ -757,7 +859,58 @@ def _get_params(self):
...
+        instructions = self._adapt(info)
+        out, best_plateau = self._stop_on_plateau(info, instructions)
+
+        # Hold onto the models that has performed the best (protects against
+        # scores decreasing or early plateauing with high score
+        models_kept = [
+            k
+            for k, v in self._model_scores.items()
+            if (v >= max(self._model_scores.values()) - 1e-5) and k not in out
+        ]
```
mrocklin: What is the 1e-5 number here?

mrocklin: Which models do we keep exactly.  After looking at this code for a while I'm still not sure I understand.  

Would it suffice to keep just the best performing model we've seen so far, and also keep track of when we saw it?  In that case then the plateau check might be something like

```python
if score > best_score:
    best_score = score
    best_model = model
    best_time = time  # or partial_fit_calls or whatnot
if time > best_time + patience:
    stop working
```

dask_ml/model_selection/_incremental.py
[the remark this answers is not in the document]
stsievert: Yup, you're right. Only the final `meta` is appended to `info[ident]`, which definitely has `score` in it at https://github.com/dask/dask-ml/blob/90fbe69771926b7e43a5bef99d81e268dad63e7e/dask_ml/model_selection/_incremental.py#L208

That took me by surprise.

stsievert:
I've implemented a better method to keep models in memory, as mentioned https://github.com/dask/dask-ml/pull/370#discussion_r224115036. This method puts any model persistence in the hands of the adaptive algorithm developer, and is a lot simpler. The implementation is back to the original implementation:

``` python
def _additional_call(self, info):
    instructions = self._adapt(info)
    return self._stop_on_plateau(info, instructions)
```

This required some changes to the tests. Mostly it's possible that the highest scoring model is not actually returned by search. This is to be expected because adaptive algorithms tend to make high probability guarantees that have some probability of failing. This isn't true `IncrementalSearchCV`'s method: there are no guarantees on returning the highest performing model with high probability (though it is inspired by an algorithm with those guarantees). This leads to a change in the tests:

``` diff
- assert (
- search.cv_results_["test_score"][search.best_index_]
-         >= search.cv_results_["test_score"]
- ).all()
- search.cv_results_["rank_test_score"][search.best_index_] == 1
+ assert (
+ search.cv_results_["test_score"][search.best_index_]
+          >= np.percentile(search.cv_results_["test_score"], 90)
+ )
+ search.cv_results_["rank_test_score"][search.best_index_] <= 3
```

This also removed some of the random errors that were happening in this PR: now the model that *is in results* and performed best on the final `partial_fit` call is returned (so no more `KeyError`s on `results.models[best_model_id]`).

Also, I expected the CI tests to pass and it looks like they do for my changes (but sklearn-dev tests have one Tornado timeout issue on one of the incremental tests?). There are also tests failing with `HashingVectorizer` and `QuantileTransformer` classes failing on py27 and py36 tests.

tests/model_selection/test_incremental.py
```diff
@@ -215,25 +240,30 @@ def test_search(c, s, a, b):
...
-def test_search_patience(c, s, a, b):
+def test_search_plateau_patience(c, s, a, b):
     X, y = make_classification(n_samples=100, n_features=5, chunks=(10, 5))
 
     class ConstantClassifier(SGDClassifier):
-        def score(*args, **kwargs):
-            return 0.5
 
-    model = ConstantClassifier(tol=1e-3)
+        def __init__(self, value=0):
```
TomAugspurger: Do you need to set `self.value` here?

dask_ml/model_selection/_incremental.py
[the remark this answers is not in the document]
stsievert: > Second note, do we need to treat the "reference" score (plateau[0] or plateau[1]) specially?

Nice catch. Implemented.

> I think that we should be doing either

The first one is implemented now. But they're both the same thing, past some typos in the second one (which is trivially true if the scores are monotonically increasing). Both cases are testing to see if a the scores are in an envelope of "width" `tol` from the "edges" of the envelope. Here's a visualization of what I mean:

screen shot 2018-10-10 at 9 30 08 pm

tests/model_selection/test_incremental.py
stsievert: Apparently not: I use `self.value` in `score`. But I don't understand how that works, so I added `self.value = value`.

dask_ml/model_selection/_incremental.py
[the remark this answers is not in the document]
stsievert: I think this conversation is resolved. We've clarified that using `history[-self.patience:]` will define patience to be number of score calls. This PR has implemented checking to see if the number of `partial_fit` calls is in the most recent `patience` calls.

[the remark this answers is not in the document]
TomAugspurger: @mrocklin does this sound OK to you? I think users should be thinking in terms of `partial_fit_calls`, not in terms of calls to `score` (which only happens once per "batch").

@stsievert can you think of a short way to explain this that would be suitable for the docstring describing `patience`?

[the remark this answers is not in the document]
mrocklin: > I think users should be thinking in terms of partial_fit_calls, not in terms of calls to score

Fine with me

dask_ml/model_selection/_incremental.py
```diff
@@ -459,43 +463,54 @@ def _get_params(self):
...
+        )
+        cv_results = {k: np.array(v) for k, v in cv_results.items()}
+
+        order = (-1 * cv_results["test_score"]).argsort()
+        cv_results["rank_test_score"] = order.argsort() + 1
+        return cv_results
+
+    def _get_best(self, results, cv_results):
+        scores = {
+            k: v[-1]["score"] for k, v in results.info.items() if k in results.models
```
TomAugspurger: One more time, since I'm still shaky on this: when can we have a model in `results.info` that's not in `results.models`? Under what conditions can we end up in a situation where the highest-scoring model isn't returned to us here?

tests/model_selection/test_incremental.py
```diff
@@ -188,22 +194,41 @@ def additional_calls(scores):
...
+        "test_score",
+        "rank_test_score",
+        "model_id",
+        "params",
+        "partial_fit_calls",
+        "param_alpha",
+        "param_l1_ratio",
+    }.issubset(set(search.cv_results_.keys()))
+    assert all(isinstance(v, np.ndarray) for v in search.cv_results_.values())
+    assert search.cv_results_["test_score"][search.best_index_] >= np.percentile(
```
TomAugspurger: I'd like to avoid this if possible. This seems like a recipe for randomly failing tests in the future.

Is there any way to reconstruct the best model from the output of `search.history_`?

dask_ml/model_selection/_incremental.py
stsievert: > when can we have a model in results.info that's not in results.models

All the time, since this PR also expands `info` so it includes history on every model at ([source](https://github.com/dask/dask-ml/pull/370/files#diff-0a54ded74c8013caf0588e9a5c879e99R260)). I've added that to https://github.com/dask/dask-ml/pull/370#issue-216493783.

> Under what conditions can we end up in a situation where the highest-scoring model isn't ~~returned to us here~~ in `results.models`?

There are two dimensions to the adaptive search:

1. how many `partial_fit` calls happened before a model was scored
2. how high the score is

For adaptive searches (which includes stop on plateau), there's no guarantee that the highest scoring model be scored with the maximum number of `partial_fit` calls.

For passive searches (e.g., `RandomizedSearchCV`) there's only one dimension: how high the final score is.

tests/model_selection/test_incremental.py
stsievert: Yes. If the search is passive (not adaptive), `search.best_index_` is guaranteed to point to the highest scoring model (because there's only one dimension to the scores as mentioned in https://github.com/dask/dask-ml/pull/370#discussion_r224456867). I've implemented this change in 0374b02ffdfa1084115df17bb6303694b0a62b24.

To make the search passive, I set the default `decay_rate=0`. With the default `patience=False`, this means `IncrementalSearchCV` is completely passive by default (added to https://github.com/dask/dask-ml/pull/370#issue-216493783). 

This is the most basic thing and the first thing to try, so I think this should be the default.

TomAugspurger:
FYI, I'm investigating the CI failures now. They seem unrelated.

tests/model_selection/test_incremental.py
mrocklin: Given your previous work showing that adaptivity is almost always a good idea I recommend that we leave the default as-is.  If we should discuss this then lets open up a separate issue and eventually open up a separate PR.  It would be good not to increase the scope of this PR further if possible.

dask_ml/model_selection/_incremental.py
mrocklin: > there's no guarantee that the highest scoring model be scored with the maximum number of partial_fit calls.

Just checking though, we're still going to return the highest scoring model back to the user at the end though, regardless of how long it was trained, right?

mrocklin: Is there anything stopping us from storing the best performing model of each hyper-parameter set that we've scored so far and then returning the best one to the user?

mrocklin: Sorry, I should have said for each hyperparameter set being considered.

I would expect that the patience process would keep track of the best scoring model we've seen so far, and if we've not seen anything better after a certain number of partial fit calls, would stop training that model, keeping a reference to the best seen so far.

Over time we would cull our models based on the scores of their best seen score.

Eventually this set would pare down to one.  That would would be the best model that we've seen so far.

In the diagram that you've shown above, I think that would store the model at the observed peaks.  This is how we would manage patience.

stsievert: I don't see an issue with the current implementation, so I'd rather not file an issue. I'll certainly respond if an issue is filed.

> I would expect that the patience process would keep track of the best scoring model we've seen so far, and if we've not seen anything better after a certain number of partial fit calls, would stop training that model, keeping a reference to the best seen so far.

That's not how I'd expect the patience process/stopping training on plateau to work – I'd expect [early stopping], which is what Scikit-Learn's SGDClassifier does ([docs]). I see early stopping as a proxy for convergence, and a method to

* prevent very bad performance
* limit computation

I do not see it as a method to provide marginal performance gains. At best, holding on to the best model will improve the score by `tol`, at least for real-world use.

I don't think this will be an issue in practice: the variance of validation scores on plateau is typically small (e.g., [CIFAR10 errors] from the [torch blog]). I'm inclined to say this noise is (typically) due to randomness in the dataset or training algorithm.

For real-world models/datasets/problems, adaptive algorithms will succeed in returning the highest scoring model, at least in terms of the final fit score (because they retain the highest scoring model every time they're called). The image I showed earlier isn't realistic, and a low probability result where even a passive search wouldn't return a high performing model.

[docs]:http://scikit-learn.org/stable/modules/sgd.html#stopping-criterion
[torch blog]:http://torch.ch/blog/2016/02/04/resnets.html
[CIFAR10 errors]:https://raw.githubusercontent.com/torch/torch.github.io/master/blog/_posts/images/resnets_CIFARtesterror_modeldepth.png
[3]:http://papers.nips.cc/paper/6768-qsgd-communication-efficient-sgd-via-gradient-quantization-and-encoding
[early stopping]:https://en.wikipedia.org/wiki/Early_stopping
[2]:http://scikit-learn.org/stable/modules/sgd.html#stopping-criterion
[1]:https://github.com/scikit-learn/scikit-learn/blob/c13ba26c4780d6088c40db3cc0f8367df4e27f1f/sklearn/linear_model/sag_fast.pyx#L501-L508

mrocklin:
@stsievert how painful would it be for you to separate this PR into a few PRs, each of which addresses only a single issue?  I'm finding it difficult to review this PR because there is a lot going on, and I'm not sure which parts are relevant to which others.

stsievert:
> how painful would it be for you to separate this PR into a few PRs, each of which addresses only a single issue?

I think it'd be easier for me to have separate commits, one for each issue (e.g., "API: add cv_results_ to IncrementalSearchCV"). I think this would be easy to review: the diffs would be small, and would have a good message. I think that'd also work better with GitHub's UI (which defaults to showing diff from master?).

Creating separate commits will be required for both choices, so let's see how it works then move to separate PRs if need be.

TomAugspurger:
FWIW, that would require force pushing commits, which makes reviewing somewhat harder since it breaks the “new changes” UI.

________________________________

stsievert:
> that would require force pushing commits, which makes reviewing somewhat harder since it breaks the “new changes” UI.

Whoops, already forced pushed. Is this a problem on your end? I'm not sure what "new changes UI" refers to. I could open another PR with these changes, and revert this PR back to it's previous state; would that help?

mrocklin:
I'm ok reviewing things commit by commit, but I'd prefer that they be atomic, so for example you would want to merge https://github.com/dask/dask-ml/pull/370/commits/0ac521fbdb694cd433a303430166e8b6a97e18f0 and https://github.com/dask/dask-ml/pull/370/commits/5299f1b537d72780e7afbd02dff671c997a914f5 into a single commit (there are probably other cases of this)

dask_ml/model_selection/_incremental.py
```diff
@@ -890,13 +890,14 @@ def inverse(time):
...
-        ):
-            next_time_step += 1
+        current_time_step = time_step
+        next_time_step = current_time_step + 1
+        if self.decay_rate:
+            while inverse(current_time_step) == inverse(next_time_step) and (
+                not self.patience
+                or next_time_step - current_time_step < self.scores_per_fit
+            ):
+                next_time_step += 1
```
mrocklin: Thanks for catching this.   It looks like this could probably use a test.

tests/model_selection/test_incremental.py
```diff
@@ -200,7 +200,9 @@ def test_search_basic(c, s, a, b):
 
     params = {"alpha": np.logspace(-2, 2, 100), "l1_ratio": np.linspace(0.01, 1, 200)}
 
-    search = IncrementalSearchCV(model, params, n_initial_parameters=20, max_iter=10)
+    search = IncrementalSearchCV(
+        model, params, n_initial_parameters=20, max_iter=10, decay_rate=0
+    )
```
mrocklin: Why this change?  Why not also engage decay rate here?

stsievert: Because for this dataset and model, I had to make a change to the tests (same change as in https://github.com/dask/dask-ml/pull/370#issuecomment-428788259):

``` diff
- assert (
- search.cv_results_["test_score"][search.best_index_]
-         >= search.cv_results_["test_score"]
- ).all()
- search.cv_results_["rank_test_score"][search.best_index_] == 1
+ assert (
+ search.cv_results_["test_score"][search.best_index_]
+          >= np.percentile(search.cv_results_["test_score"], 90)
+ )
+ search.cv_results_["rank_test_score"][search.best_index_] <= 3
```

I interpret this change as meaning "random searches work best", which makes sense because the model/dataset are simple and not performant. Plus, this test is aimed at testing basic elements of `IncrementalSearchCV` but also relies on the adaptive algorithm working well.

dask_ml/model_selection/_incremental.py
```diff
@@ -502,21 +500,24 @@ def _get_cv_results(self, history, model_hist):
...
-        best_index = -1
-        # history_results is sorted by (model_id, partial_fit_calls)
-        # best is the model_id with the highest partial fit calls
-        for k, v in itertools.groupby(history_results, key=key):
-            v = list(v)
-            best_index += len(v)
-            if k == best_model_id:
-                break
-
-        return results.models[best_model_id], best_index
```
mrocklin: I'm a bit confused about the need for `_get_best` generally.  Why do we need to search for the best result throughout history.  Is this information that `_fit` could just give us directly?  I suspect that we're spending time copying how sklearn does things when it might not be the best approach overall.  Thoughts?

dask_ml/model_selection/_incremental.py
```diff
@@ -849,7 +849,35 @@ def _get_params(self):
...
+
+            else:
+                out[k] = steps
+        return out
+
     def _additional_calls(self, info):
+        instructions = self._adapt(info)
+
+        out = self._stop_on_plateau(info, instructions)
+        return out
```
mrocklin: I think that I would prefer to keep things as a single method until a concrete need arises to separate them.  

While I appreciate the desire to split things apart into smaller chunks I find this adds a cost of indirection, which makes code a bit harder to review and maintain in the future.  I find that code that is finely split often feels better when first writing it, but ages poorly when you have to quickly understand what is going on after a year of not looking at it. 
 There is obviously a length limit to this (functions of several hundred lines are unpleasant), but I don't think that the `_additional_calls` function is yet there.

I haven't yet reviewed the other parts of this commit that change the behavior of `_additional_calls`, partly because it's hard to separate those changes from the refactor.  If possible I'd prefer that we remove the refactor so that it's easier to review and discuss the algorithmic changes.

dask_ml/model_selection/_incremental.py
```diff
@@ -817,6 +818,24 @@ class IncrementalSearchCV(BaseIncrementalSearchCV):
     >>> search = IncrementalSearchCV(model, params, random_state=0,
     ...                              n_initial_parameters=1000,
     ...                              patience=20, max_iter=100)
+
+    ``tol`` and ``patience`` are for stopping training on a "plateau", or when
```
mrocklin: Is the term `plateau` likely to be well understood by novice users?  If yes then we should remove the quotes.  If no then we might want to consider avoiding the term and instead just describe it as you do after using it.  I might even start off this paragraph with 

"Often when training we get to a situation where additional work leads to little to no gain.  In these cases the ``tol`` and ``patience`` parameters can help us stop work early.  ..."

dask_ml/model_selection/_incremental.py
```diff
@@ -817,6 +818,24 @@ class IncrementalSearchCV(BaseIncrementalSearchCV):
...
+    the hold-out score flattens and/or falls. For example, setting ``tol=0``
+    and ``patience=2`` will dictate that scores will always increase until
```
mrocklin: We're not dictating that scores will increase (we don't determine that).  We might need to change the language here.

dask_ml/model_selection/_incremental.py
stsievert: >  Why do we need to search for the best result throughout history

We don't. The current implementation more or less does

``` python
score = {k: v[-1]["score"] for k, v in results.info.items()}
best_model_id = max(score, key=score.get)
```

The previous implementation picked the best model ID before searching through history.

> Is this information that _fit could just give us directly?

I refactored `_fit` to return information that's very useful in calculating that information, but not that information (the "best" model ID?) directly.

In this PR, `_fit` gives us model history separated by model ID for every model (not only the final models `_additional_calls` selects) in `results.info`. This is really useful: it allows getting the final scores for *every model* with

``` python
info, models, history = fit(...)
final_scores = {k: v[-1]["score"] for k, v in info.items()}
best_model_id = max(final_scores, key=final_scores.get)
```

tests/model_selection/test_incremental.py
stsievert: That is, without `decay_rate=0`, this search wouldn't return the highest scoring model (with this small dataset and simple model, the search is too adaptive).

mrocklin:
This is where commit-by-commit reviewing becomes challenging.  I'm going to ask that you rebase to keep reviewable chunks atomic.  (separate PRs would also be welcome of course).

stsievert:
> This is where commit-by-commit reviewing becomes challenging. I'm going to ask that you rebase to keep reviewable chunks atomic. (separate PRs would also be welcome of course).

I see. I'll try to split into separate 3 PRs:

1. API improvements (`cv_results_`, renaming, `model_history_`, etc).
2. Documentation changes
3. Bugs fixes/cleaning stop on plateau

How does that sound? Should the PRs be finer grained?

PR #372, opened 2018-09-19
Added ArrayConverter

TomAugspurger:
This is helpful when for pipelines where you load the data with dask dataframe, and a downstream task like IncrementalSearch requires a dask array with known chunk sizes.

### DIFF
--- a/dask_ml/preprocessing/__init__.py
+++ b/dask_ml/preprocessing/__init__.py
@@ -13,6 +13,7 @@ from .data import (
     DummyEncoder,
     OrdinalEncoder,
 )
+from ._converters import ArrayConverter
 from .label import LabelEncoder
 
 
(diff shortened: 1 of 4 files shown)


mrocklin:
Alternatively I'll put forward the thought that we try hard to make all objects support dask dataframes and dask arrays without known chunk sizes.  Is there a list of these somewhere?

TomAugspurger:
There's no list, but I could compile one.

> we try hard to make all objects support dask dataframes and dask arrays without known chunk sizes.

There are probably places where unknown chunk sizes could be supported without much effort. However, things like https://github.com/dask/dask-ml/blob/b5b6954cbd8a4da89aa4da9612e970a0680290da/dask_ml/model_selection/_split.py#L160 will I think need known sizes (though a clever workaround may be possible). At that point having a way to specify the lengths up-front would be useful.

TomAugspurger:
Closing this in favor of properly handling unknown chunks in more places.

PR #373, opened 2018-09-19
IncrementalSearch edge cases

TomAugspurger:
A couple issues when given a small grid.

For the first, we decided that training 0 models was best, as `inverse(next_time_step)` was 0.

For the second, we got stuck in an infinite `while True` loop since

1. `inverse(current_time_step) == inverse(next_time_step)` was always equal (both were 0), and
2. next_time_step is increasing.

We could detect this and raise when something like `start < current_time_step + 1`, but ideally we would support this. I'm not sure that my fix is correct though.

### DIFF
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -772,13 +772,18 @@ class IncrementalSearch(BaseIncrementalSearch):
 
         current_time_step = time_step + 1
         next_time_step = current_time_step
+
+        if inverse(current_time_step) == 0:
+            # we'll never get out of here
+            next_time_step = 1
+
         while inverse(current_time_step) == inverse(next_time_step) and (
             not self.patience
             or next_time_step - current_time_step < self.scores_per_fit
         ):
             next_time_step += 1
 
-        target = inverse(next_time_step)
+        target = max(1, inverse(next_time_step))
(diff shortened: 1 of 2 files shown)


dask_ml/model_selection/_incremental.py
```diff
@@ -772,13 +773,17 @@ def inverse(time):
...
         current_time_step = time_step + 1
         next_time_step = current_time_step
-        while inverse(current_time_step) == inverse(next_time_step) and (
-            not self.patience
-            or next_time_step - current_time_step < self.scores_per_fit
-        ):
-            next_time_step += 1
+        if inverse(current_time_step) == 0:
+            # we'll never get out of here
+            next_time_step = 1
```
mrocklin: Maybe the `inverse` function itself should never produce non-positive values?  I can imagine placing the `min(1, ` there instead, or perhaps using `math.ceil` rather than `int`

TomAugspurger: I think this additional `if` check is still necessary, even if we change `inverse` to be always positive. When these start off at the same value (whether it's 0 or 1), we get stuck in a cycle where `next_time_step` is increasing, making the `inverse(current_time_step) == inverse(next_time_step)` condition even more false.

Issue #376, opened 2018-09-29
Problem with parallel post fit wrapper (ParallelPostfitWrapper)--Can't drop an axis with more than 1 block. Please use atop instead


firstkingofrome:
Basically I am trying to evaluate some sklearn predictors that I came up with using standard sklearn to produce larger than memory array outputs. This code illustrates the problem:

from dask.base import tokenize
import numpy as np
import dask.array as da
from dask.array import Array
from sklearn.linear_model import LinearRegression
from dask_ml.wrappers import ParallelPostFit
"""
sample of problem
"""
x = np.linspace(0,100,100,dtype=np.int32)
y = np.linspace(0,100,100,dtype=np.int32)
z = np.linspace(0,100,100,dtype=np.int32)

Y = np.random.normal(size=(100,))
X = np.stack([x,y,z],axis=1)
reg = LinearRegression().fit(X,Y)

#now try to compute on dask arrays over the whole space
x= da.linspace(0,100,100,chunks=(10,)).astype(np.int32)
y= da.linspace(0,100,100,chunks=(10,)).astype(np.int32)
z= da.linspace(0,100,100,chunks=(10,)).astype(np.int32)
x,y,z = da.meshgrid(x,y,z,sparse=False,indexing='ij')
stacked = da.stack([x.flatten(),y.flatten(),z.flatten()],axis=1)
clf = ParallelPostFit(estimator=reg)
clf.predict(stacked)

Executing clf.predict throws a value error Can't drop an axis with more than 1 block. Please use atop instead.

which I dont understand how to correct. Thank You for any help.

mrocklin:
I suspect that `ParallelPostFit` assumes that the data has a single block in the columns direction.  @TomAugspurger thoughts about providing an informative error here pointing users to `X.reshape({1: -1})` or doing this for them automatically?

TomAugspurger:
That’s correct. We should at least be calling check_array, which validates this. In principle a rechunk with “auto” and -1 should do what we want.

________________________________

firstkingofrome:
I just wanted to thank you both for how quickly you helped resolve this and for dask.

TomAugspurger:
This will be auto-closed once #377 is merged. Just need to sort out one issue there first.

PR #377, opened 2018-10-01
Auto-rechunk input arrays

TomAugspurger:
Closes https://github.com/dask/dask-ml/issues/376

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -18,7 +18,7 @@ from .metrics import check_scoring, get_scorer
 logger = logging.getLogger(__name__)
 
 
-class ParallelPostFit(sklearn.base.BaseEstimator):
+class ParallelPostFit(sklearn.base.BaseEstimator, sklearn.base.MetaEstimatorMixin):
     """Meta-estimator for parallel predict and transform.
 
     Parameters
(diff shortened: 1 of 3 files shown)


dask_ml/wrappers.py
```diff
@@ -119,6 +119,29 @@ def __init__(self, estimator=None, scoring=None):
...
+        Notes
+        -----
+        The following checks are applied.
+
+        - Ensure that the array is blocked only along the samples.
+        """
+        if isinstance(X, da.Array):
+            if X.ndim == 2 and X.numblocks[1] > 1:
+                logger.debug("auto-rechunking 'X'")
+                X = X.rechunk({0: "auto", 1: -1})
```
mrocklin: I think it's an open question on whether or not we want to include `{0: 'auto'}` here.  I can imagine situations where either including or not including it would be problematic.

TomAugspurger: Just to be clear, my PR only auto-rechunks when `X.numblcoks[1] > 1`. In your mini-example if `X.shape == (40, 10)`, we won't rechunk in `_check_array`.

mrocklin:
+1 from me.  I'm not sure what's up with the Python 2.7 failure though.

TomAugspurger:
It's been around for a few weeks now. I haven't spent the time to debug it yet.

PR #381, opened 2018-10-04
Fixes ShuffleSplit random seed generation bug

### LINKED ISSUE #380: ShuffleSplit makes same split for every iteration if random_state is set
Hi!
ShuffleSplit produces the same data split for every iteration when the random_state is set to a number. I would expect that each iteration would produce a different data split.

```
import dask
from dask_ml.model_selection import ShuffleSplit
import sklearn.datasets

x, y = sklearn.datasets.make_classification(n_samples=10000, n_features=100, n_informative=50, n_clusters_per_class=5, n_classes=5)
x = dask.array.from_array(x, chunks=(100,-1))
y = dask.array.from_array(y, chunks=(100))

cv = ShuffleSplit(n_splits=2, test_size=0.3, train_size=0.7, random_state=0)
for train_idx, test_idx in cv.split(X=x, y=y):
    print(train_idx.compute())

[  25    1   45 ... 9967 9951 9918]
[  25    1   45 ... 9967 9951 9918]
```

Note: It works as expected if random_state is not set
```
cv = ShuffleSplit(n_splits=2, test_size=0.3, train_size=0.7)
for train_idx, test_idx in cv.split(X=x, y=y):
    print(train_idx.compute())

[  46   70   25 ... 9912 9907 9959]
[   5   93   47 ... 9984 9929 9946]
```

The issue seems to lie in `dask-ml/dask_ml/model_selection/_split.py` method `_split_blockwise(self, X):` wherein the rng is set to `rng = check_random_state(self.random_state)` for each new cross validation.

jrbourbeau:
This PR fixes a bug in `ShuffleSplit` that causes the same random seeds to be used for each split when `random_state` is given. Also adds a regression test. 

Closes #380

### DIFF
--- a/dask_ml/model_selection/_split.py
+++ b/dask_ml/model_selection/_split.py
@@ -150,17 +150,16 @@ class ShuffleSplit(BaseCrossValidator):
 
     def split(self, X, y=None, groups=None):
         X = check_array(X)
-
+        rng = check_random_state(self.random_state)
         for i in range(self.n_splits):
+            seeds = rng.randint(0, 2 ** 32 - 1, size=len(X.chunks[0]), dtype="u8")
             if self.blockwise:
-                yield self._split_blockwise(X)
+                yield self._split_blockwise(X, seeds)
             else:
                 yield self._split(X)
 
-    def _split_blockwise(self, X):
+    def _split_blockwise(self, X, seeds):
         chunks = X.chunks[0]
-        rng = check_random_state(self.random_state)
-        seeds = rng.randint(0, 2 ** 32 - 1, size=len(chunks), dtype="u8")
 
         train_pct, test_pct = _maybe_normalize_split_sizes(
             self.train_size, self.test_size
--- a/tests/model_selection/test_split.py
+++ b/tests/model_selection/test_split.py
@@ -51,6 +51,32 @@ def test_blockwise_shufflesplit():
     )
 
 
+def test_blockwise_shufflesplit_rng():
+    # Regression test for issue #380
+    n_splits = 2
+    splitter = dask_ml.model_selection.ShuffleSplit(n_splits=n_splits, random_state=0)
+    gen = splitter.split(dX)
+
+    train_indices = []
+    test_indices = []
+    for train_idx, test_idx in gen:
+        train_indices.append(train_idx)
+        test_indices.append(test_idx)
+
+    assert not np.array_equal(train_indices[0], train_indices[1])
+    assert not np.array_equal(test_indices[0], test_indices[1])
+
+    # Test that splitting is reproducible
+    n_splits = 2
+
+    split1 = dask_ml.model_selection.ShuffleSplit(n_splits=n_splits, random_state=0)
+
+    split2 = dask_ml.model_selection.ShuffleSplit(n_splits=n_splits, random_state=0)
+    for (train_1, test_1), (train_2, test_2) in zip(split1.split(dX), split2.split(dX)):
+        da.utils.assert_eq(train_1, train_2)
+        da.utils.assert_eq(test_1, test_2)
+
+
 @pytest.mark.parametrize("shuffle", [False, True])
 def test_kfold(shuffle):
     splitter = dask_ml.model_selection.KFold(


mrocklin:
I recommending adding a `pip install pip --upgrade` line in the CI script to see if that helps.  I believe that pip has had issues recently.

TomAugspurger:
I haven't seen that issue before. If Matt's suggestion doesn't work then I can take a look.

Could you make a new helper function in dask_ml/_utils.py for this? i.e. anywhere we do `RandomState.randint`? I see the following.

```
dask_ml/cluster/k_means.py:391:    idx = sorted(random_state.randint(0, len(X), size=n_clusters))
dask_ml/cluster/k_means.py:459:        rng2 = random_state.randint(0, 2 ** 32 - 1, chunks=()).compute().item()
dask_ml/cluster/spectral.py:192:                    n_clusters=n_clusters, random_state=rng.randint(2 ** 32 - 1)
dask_ml/utils.py:183:        return da.random.RandomState(random_state.randint())
dask_ml/model_selection/_split.py:163:        seeds = rng.randint(0, 2 ** 32 - 1, size=len(chunks), dtype="u8")
dask_ml/model_selection/_split.py:253:            seeds = rng.randint(0, 2 ** 32 - 1, size=len(chunks), dtype="u8")
dask_ml/model_selection/_split.py:430:        rng = rng.randint(0, 2 ** 32)
dask_ml/decomposition/pca.py:246:            seed = random_state.randint(np.iinfo("int32").max)
```

Something like

```
def draw_seed(random_state, size=None, dtype=None, chunks=None):
    kwargs = {'size': size, 'dtype': dtype}
    if chunks is not None:
        kwargs['chunks'] = chunks
    return random_state.randint(0, 2 ** 32 - 1, **kwargs)
```

dask_ml/cluster/k_means.py
```diff
@@ -456,7 +457,7 @@ def init_scalable(
     else:
         # Step 7, 8 without weights
         # dask RandomState objects aren't valid for scikit-learn
-        rng2 = random_state.randint(0, 2 ** 32 - 1, chunks=()).compute().item()
+        rng2 = draw_seed(random_state, 0, 2 ** 32 - 1, chunks=()).compute().item()
```
TomAugspurger: I think we need `dtype='uint'` here.

dask_ml/cluster/spectral.py
```diff
@@ -189,7 +190,7 @@ def fit(self, X, y=None):
         if isinstance(self.assign_labels, six.string_types):
             if self.assign_labels == "kmeans":
                 km = KMeans(
-                    n_clusters=n_clusters, random_state=rng.randint(2 ** 32 - 1)
+                    n_clusters=n_clusters, random_state=draw_seed(rng, 2 ** 32 - 1)
```
TomAugspurger: Again, `dtype='uint'`.

dask_ml/model_selection/_split.py
```diff
@@ -150,17 +152,16 @@ def __init__(
 
     def split(self, X, y=None, groups=None):
         X = check_array(X)
-
+        rng = check_random_state(self.random_state)
         for i in range(self.n_splits):
+            seeds = draw_seed(rng, 0, 2 ** 32 - 1, size=len(X.chunks[0]), dtype="u8")
```
TomAugspurger: Maybe `dtype='uint'` instead of `'u8`? I'm not sure what happens to `u8` on 32-bit platforms.

dask_ml/model_selection/_split.py
```diff
@@ -250,7 +251,7 @@ def split(self, X, y=None, groups=None):
         seeds = [None] * len(chunks)
         if self.shuffle:
             rng = check_random_state(self.random_state)
-            seeds = rng.randint(0, 2 ** 32 - 1, size=len(chunks), dtype="u8")
+            seeds = draw_seed(rng, 0, 2 ** 32 - 1, size=len(chunks), dtype="u8")
```
TomAugspurger: `dtype='uint'`

jrbourbeau:
Thanks for merging in master. I'll add `dtype='uint'`

TomAugspurger:
xfailing that test warnings. I've never been able to reproduce it locally, and I don't really understand how it could be failing in the first place.

PR #387, opened 2018-10-06
Update joblib documentation for scikit-learn 0.20

mrocklin:


mrocklin:
It would be nice to add a list at the bottom of common operations within scikit-learn that can benefit from this approach.

mrocklin:
> Also, is there a reason for removing the scatter= section for parallel_backend?

While still available, this feature isn't really necessary any more.  Joblib will automatically detect and scatter data during computation.

jrbourbeau:
Sweet, I hadn't realized that was done automatically. Good to know!

TomAugspurger:
Added a note that this is useful for CPU-bound workloads.

PR #389, opened 2018-10-07
Added DeprecationDict

### LINKED ISSUE #383: Compat with scikit-learn dev
Handle removal of DeprecationDict (and the deprecation it was supporting).

jrbourbeau:
This PR adds `DeprecationDict`, which was removed from scikit-learn, to `dask_ml/model_selection/utils.py`. Moving `DeprecationDict` over to `dask_ml` seems reasonable to be, but I'm open to suggestions if other have them. 

Closes  #383

### DIFF
--- a/dask_ml/model_selection/_search.py
+++ b/dask_ml/model_selection/_search.py
@@ -54,7 +54,7 @@ from .methods import (
     pipeline,
     score,
 )
-from .utils import is_dask_collection, to_indexable, to_keys, unzip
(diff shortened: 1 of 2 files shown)


TomAugspurger:
FYI, I'm going to be bumping the min scikit-learn version in https://github.com/dask/dask-ml/issues/391

I'll come back to this after I've done that. We'll still need something, but it'll hopefully be a bit simpler.

TomAugspurger:
This was included in #392

PR #393, opened 2018-10-08
Support dataframes for k-means

### LINKED ISSUE #390: KMeans.fit on uncertain lengths
This seems like a reasonable thing that ought to work.  Any ideas on how we could make this happen?

https://stackoverflow.com/questions/52583316/how-to-pass-dask-dataframe-as-input-to-dask-ml-models

```python
import dask.dataframe as dd
import pandas as pd
from dask_ml.cluster import KMeans

df = dd.from_pandas(pd.DataFrame({'A': [1, 2, 3, 4, 5], 
                                  'B': [6, 7, 8, 9, 10]}),
                    npartitions=2)

kmeans = KMeans()
kmeans.fit(df)
```

1.  Do we compute lengths explicitly but warn that we're doing so?
2.  Do we actually need explicit lengths (I was somewhat surprised by this)

mrocklin:
Fixes #390 

I went with the easier approach of just forcing chunk lengths.  I did this after I realized that we're going to be doing multiple passes anyway.

### DIFF
--- a/dask_ml/cluster/k_means.py
+++ b/dask_ml/cluster/k_means.py
@@ -157,7 +157,7 @@ class KMeans(TransformerMixin, BaseEstimator):
             X = X.values
 
         if isinstance(X, dd.DataFrame):
-            X = X.to_dask_array(True)
+            X = X.to_dask_array(lengths=True)
 
         X = check_array(
             X,
--- a/dask_ml/cluster/k_means.py
+++ b/dask_ml/cluster/k_means.py
@@ -156,10 +156,8 @@ class KMeans(TransformerMixin, BaseEstimator):
         if isinstance(X, pd.DataFrame):
             X = X.values
 
-        elif isinstance(X, dd.DataFrame):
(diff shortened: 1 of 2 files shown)


dask_ml/cluster/k_means.py
```diff
@@ -156,10 +156,8 @@ def _check_array(self, X):
         if isinstance(X, pd.DataFrame):
             X = X.values
 
-        elif isinstance(X, dd.DataFrame):
-            raise TypeError(
-                "Cannot fit on dask.dataframe due to unknown " "partition lengths."
-            )
+        if isinstance(X, dd.DataFrame):
+            X = X.to_dask_array(True)
```
TomAugspurger: style-nit: specify the `lengths=` keyword for boolean arguments.

LGTM otherwise.

TomAugspurger:
Looks like `tests/test_kmeans.py::TestKMeans::test_dask_dataframe_raises` will need to be removed.

PR #395, opened 2018-10-09
Support dataframes in _partial.py::fit/predict

mrocklin:
Previously these functions would fail on dask dataframes.
Now they coerce to dask arrays, and predict also converts back

I'm not sure if this logic is also handled elsewhere.  I was running into it with the criteo example though.

### DIFF
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -162,11 +162,11 @@ def fit(model, x, y, compute=True, shuffle_blocks=True, random_state=None, **kwa
     >>> da.learn.predict(sgd, z)  # doctest: +SKIP
     dask.array<x_11, shape=(400,), chunks=((100, 100, 100, 100),), dtype=int64>
     """
-    if not hasattr(x, 'chunks') and hasattr(x, 'to_dask_array'):
+    if not hasattr(x, "chunks") and hasattr(x, "to_dask_array"):
         x = x.to_dask_array()
     assert x.ndim == 2
     if y is not None:
-        if not hasattr(y, 'chunks') and hasattr(y, 'to_dask_array'):
+        if not hasattr(y, "chunks") and hasattr(y, "to_dask_array"):
             y = y.to_dask_array()
         assert y.ndim == 1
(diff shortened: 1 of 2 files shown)


mrocklin:
It is possible to reproduce on master.  Raised here: https://github.com/dask/dask-ml/issues/396

dask_ml/_partial.py
```diff
@@ -213,13 +218,21 @@ def predict(model, x):
...
     assert x.ndim == 2
     if len(x.chunks[1]) > 1:
         x = x.rechunk(chunks=(x.chunks[0], sum(x.chunks[1])))
     func = partial(_predict, model)
     xx = np.zeros((1, x.shape[1]), dtype=x.dtype)
     dt = model.predict(xx).dtype
-    return x.map_blocks(func, chunks=(x.chunks[0], (1,)), dtype=dt).squeeze()
+    result = x.map_blocks(func, chunks=(x.chunks[0], (1,)), dtype=dt).squeeze()
+    if index is not None:
+        result = dd.from_dask_array(result, index=index)
```
TomAugspurger: Hmm I'm having second thoughts about this. I think this is the wrong place to determine whether a Series should be returned. I think that should be up to `model.predict`. And given the current implementation, `model.predict` is only going to return arrays, since it's only ever getting arrays.

So for now I'd recommend reverting the `predict` related changes, and just keep this to support dataframes in `fit`

TomAugspurger: My main priority is matching the behavior of the underlying estimator. Ideally `Incremental(estimator)` is behaves as close as possible to

```
for block in array:
    estimator.partial_fit(block)
```

And prediction is a simple `map(predict, blocks)`.

So if `estimator.predict(pd.DataFrame) -> ndarray`, we would want `estimator.predict(dd.DataFrame) -> da.Array`. At the moment, IIUC, that would return a `Series`.

---

And now I realize I should have read your whole comment. Let's just delete `predict`?

Regardless that can be done elsewhere.

Issue #399, opened 2018-10-11
Is it safe to replace _apply_partitionwise with map_partitions?


mrocklin:
I'm not understanding why this code is here:

https://github.com/dask/dask-ml/blob/90fbe69771926b7e43a5bef99d81e268dad63e7e/dask_ml/wrappers.py#L504-L522

My guess is that it resolves some deficiency in `map_partitions`.  Is that still true?

TomAugspurger:
From 
https://github.com/dask/dask-ml/pull/132#discussion_r166096031

> Is this something that we would want to push upstream into Dask?
 
We were blocked by `map_partitions` returning an array. We should be able to use it now.

It's funny how consistent humans can be :)

jrbourbeau:
I think this was resolved in #400 and this issue can now be closed

TomAugspurger:
Yes, thanks.

PR #400, opened 2018-10-11
Pass models in predict rather than their methods

mrocklin:
Previously we would pass around Estimator.predict methods.  These methods are
opaque to serialization heuristics used in dask.distributed that are used to
determine what should move and how to serialize it.

Now we pass around bare functions that take in estimators as parameters.

### DIFF
--- a/dask_ml/feature_extraction/text.py
+++ b/dask_ml/feature_extraction/text.py
@@ -37,15 +37,13 @@ class HashingVectorizer(sklearn.feature_extraction.text.HashingVectorizer):
 
         See the :doc:`examples/text-vectorization` for more.
         """
-        transformer = super(HashingVectorizer, self).transform
-
         msg = "'X' should be a 1-dimensional array with length 'num_samples'."
 
         if not dask.is_dask_collection(X):
-            return transformer(X)
+            return super(HashingVectorizer, self).transform(X)
 
         if isinstance(X, db.Bag):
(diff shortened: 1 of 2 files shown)


mrocklin:
I would also like to replace this code in this PR if sensible: https://github.com/dask/dask-ml/issues/399

TomAugspurger:
+1 removing apply_partitionwise.

FYI, I'm debuging the CI failures now. I don't have a minimal example for `tests.feature_extraction.test_text.py::test_hashing_vectorizer[bag]` yet. It's failing in the new atop-fusion code. This doesn't hit it unfortunately

```python
In [28]: X = da.random.uniform(size=(10, 10), chunks=5)

In [29]: X.map_blocks(scipy.sparse.csr_matrix).blocks[0].compute()
Out[29]:
<5x10 sparse matrix of type '<class 'numpy.float64'>'
        with 50 stored elements in COOrdinate format>
```

mrocklin:
I'm happy to take a look if you like

TomAugspurger:
Hmm yeah I might appreciate that :)

It seems that all 3 are atop-fusion related... `tests/preprocessing/test_data.py::TestQuantileTransformer::test_basic` seems like the easiest. The other two involve cross dataframe or bag -> dask array boundaries.

mrocklin:
Cool.  I'll finish up what I'm working on now (more of this PR) and then
switch to that.  Should be there in 10min or so

PR #402, opened 2018-10-11
Allow compute=False in ParallelPostFit.score

mrocklin:
I notice that we are not consistent in the various `score` methods in our classes.  Some of them are lazy, some of them are immediate.  None of them support a `compute=` keyword. 

All of the others are trivial to add a `compute=` keyword, what should the default be?  True or False?

### DIFF
--- a/dask_ml/metrics/scorer.py
+++ b/dask_ml/metrics/scorer.py
@@ -1,3 +1,4 @@
+import functools
 import six
 from sklearn.metrics import make_scorer
 from sklearn.metrics.scorer import check_scoring as sklearn_check_scoring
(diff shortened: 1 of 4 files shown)


TomAugspurger:
On estimators, the `.score` method has to be immediate.

I don't think there's a similar restriction on the scoring functions, so having them be lazy by default makes the most sense to me.

mrocklin:
> On estimators, the .score method has to be immediate.

Could it be immediate by default, but still provide a compute=False keyword?

Also, I think that some `.score` functions may currently be lazy.  The dask-glm ones perhaps?  I could be wrong though, I haven't actually run these yet, just looked through code.

TomAugspurger:
A keyword would be fine.

Not sure about dask-glm. I may be wrong, but I think this would only potentially show up in something like a `GridSearchCV(model, params)`

Issue #388, opened 2018-10-07
IncrementalSearchCV runs an adaptive algorithm by default


stsievert:
There are two dimensions in which model selection searches can vary:

> Does the search...
>
> 1. use `partial_fit` or `fit`?
> 2. use previous evaluations to choose which models to evaluate?
>
> (adapted from https://github.com/dask/dask-ml/pull/370#discussion_r221819839, which is a response to https://github.com/dask/dask-ml/pull/370#discussion_r221573023)

I think (1) should be named "incremental" and (2) should be named "adaptive".

Currently, `IncrementalSearchCV` implements

* a complex adaptive algorithm
* stopping on plateau
* an incremental search that's used with `_incremental.fit`.

I'd like to clean the naming of `IncrementalSearchCV`.

mrocklin:
I think that fundamentally there is a tension here between being precise and being concise.  A name that fully encompassed all of the characteristics of the current implementation probably isn't feasible in less than twenty characters.  It's not clear to me how to make this compromise.  Do you have any thoughts on the right way forward?

stsievert:
> Do you have any thoughts on the right way forward?

I think my biggest complaint is the mixing of the incremental model selection search and the complex adaptive algorithm. Maybe something like

``` python
class IncrementalSearchCV(...):
    def _additional_calls(self, info):
        ...

    def _adapt(self, info):
        # default: don't adapt (allows mirroring RandomizedSearchCV
        # and GridSearchCV, but stopping on plateau)
        return {k: 1 for k in info}

    def _stop_on_plateau(self, info):
        return ...

class TimeDecaySearchCV(IncrementalSearchCV):
    def _adapt(self, info):  # as in #370
        return ...
```

mrocklin:
Is your only concern about the current design that it doesn't allow for a hierarchy of names?  

Originally the plan was to have a super-class that people would inherit from and override `_additional_calls` in order to implement their own solution.  We did this and implemented a couple solutions, but found that it was pretty easy to combine all of the traits that we wanted into a single implementation.  The ability to subclass remains, but the default solution is just decently capable. 

I think I see reducing the pool of models over time to be similar to early stopping or how we select parameters to start with.  They're all features of the same general process.  The extent to which we can make them orthogonal so that they can be changed individually by overloading some method seems nice.  I don't think that any of them need to be pulled out to form a separate class hierarchy to achieve this though.

stsievert:
[...]

PR #404, opened 2018-10-15
API: improve and clean IncrementalSearch API

stsievert:
The PR makes 4 improvements to the incremental search API:

* renames `IncrementalSearch` to `IncrementalSearchCV`
* renames `history_results_` to `history_`
* adds `cv_results_` to mirror Scikit-learn's `RandomizedSearchCV`
* returns all model information from `_incremental.fit`
    * which is implemented as `model_history_` in `IncrementalSearchCV`
* allow `incremental._fit` to return multiple models
    * in `IncrementalSearchCV`, choose the best among those models.

### DIFF
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -1,6 +1,5 @@
 from __future__ import division
 
-import itertools
 import operator
 from collections import defaultdict, namedtuple
 from copy import deepcopy
@@ -18,7 +17,6 @@ from sklearn.model_selection import ParameterGrid, ParameterSampler
 from sklearn.utils import check_random_state
 from sklearn.utils.metaestimators import if_delegate_has_method
 from sklearn.utils.validation import check_is_fitted
-from toolz import first
 from tornado import gen
 
 from ..utils import check_array
@@ -502,21 +500,24 @@ class BaseIncrementalSearchCV(BaseEstimator, MetaEstimatorMixin):
         cv_results["rank_test_score"] = order.argsort() + 1
         return cv_results
 
-    def _get_best(self, results, history_results):
-        # type: (Dict, Dict) -> Estimator
-        """Select the best estimator from the set of estimators."""
-        best_model_id = first(results.info)
-        key = operator.itemgetter("model_id")
-        best_index = -1
-        # history_results is sorted by (model_id, partial_fit_calls)
-        # best is the model_id with the highest partial fit calls
-        for k, v in itertools.groupby(history_results, key=key):
-            v = list(v)
-            best_index += len(v)
-            if k == best_model_id:
-                break
+    def _get_best(self, results, cv_results):
+        scores = {
+            k: v[-1]["score"] for k, v in results.info.items() if k in results.models
+        }
 
-        return results.models[best_model_id], best_index
+        # Could use max(scores, key=score.get), but what if score is repeated?
+        # Happens in the test case a lot
+        model_ids = list(scores.keys())
+        scores = [scores[k] for k in model_ids]
+        model_idx = np.argmax(scores)
+        best_model_id = model_ids[model_idx]
+
+        best_est = results.models[best_model_id]
+
+        idx = cv_results["model_id"] == best_model_id
+        assert idx.sum() == 1
+        best_idx = np.argmax(idx)
+        return best_idx, best_est
 
     def _process_results(self, results):
         """Called with the output of `fit` immediately after it finishes.
@@ -554,6 +555,7 @@ class BaseIncrementalSearchCV(BaseEstimator, MetaEstimatorMixin):
         model_history, models, history = results
 
         cv_results = self._get_cv_results(history, model_history)
+        best_idx, best_est = self._get_best(results, cv_results)
         best_estimator = yield best_est
 
         # Clean up models we're hanging onto
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -465,6 +465,42 @@ class BaseIncrementalSearchCV(BaseEstimator, MetaEstimatorMixin):
         """
         return ParameterGrid(self.parameters)
 
+    def _get_cv_results(self, history, model_hist):
+        cv_results = {}
+        best_scores = {}
+        best_scores = {k: hist[-1]["score"] for k, hist in model_hist.items()}
+
+        cv_results = {}
+        for k, hist in model_hist.items():
+            pf_times = list(toolz.pluck("partial_fit_time", hist))
+            score_times = list(toolz.pluck("score_time", hist))
+            cv_results[k] = {
+                "mean_partial_fit_time": np.mean(pf_times),
+                "mean_score_time": np.mean(score_times),
+                "std_partial_fit_time": np.std(pf_times),
+                "std_score_time": np.std(score_times),
+                "test_score": best_scores[k],
+                "model_id": k,
+                "params": hist[0]["params"],
+                "partial_fit_calls": hist[-1]["partial_fit_calls"],
+            }
+        cv_results = list(cv_results.values())  # list of dicts
+        cv_results = {k: [res[k] for res in cv_results] for k in cv_results[0]}
+
+        # Every model will have the same params because this class uses either
+        # ParameterSampler or ParameterGrid
+        cv_results.update(
+            {
+                "param_" + k: v
+                for params in cv_results["params"]
+                for k, v in params.items()
+            }
+        )
+        cv_results = {k: np.array(v) for k, v in cv_results.items()}
+
+        order = (-1 * cv_results["test_score"]).argsort()
+        cv_results["rank_test_score"] = order.argsort() + 1
+        return cv_results
 
     def _get_best(self, results, history_results):
         # type: (Dict, Dict) -> Estimator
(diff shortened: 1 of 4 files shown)


stsievert:
I'm running into some test difficulty with git. I don't have time to resolve them today.

tests/model_selection/test_incremental.py
```diff
@@ -188,22 +193,42 @@ def additional_calls(scores):
...
+        "test_score",
+        "rank_test_score",
+        "model_id",
+        "params",
+        "partial_fit_calls",
+        "param_alpha",
+        "param_l1_ratio",
+    }.issubset(set(search.cv_results_.keys()))
+    assert all(isinstance(v, np.ndarray) for v in search.cv_results_.values())
+    assert (
```
TomAugspurger: This is failing for me occasionally (locally too).

TomAugspurger: The immediate failure is because

```
(Pdb) pp pd.DataFrame(search.cv_results_).iloc[[5, 12]][['model_id', 'test_score']]
    model_id  test_score
5          5        0.68
12        12        0.65
```

But looking at the history, I can't convince myself that we should have picked model 12 as the best.

dask_ml/model_selection/_incremental.py
```diff
@@ -459,43 +463,61 @@ def _get_params(self):
...
+        cv_results.update(
+            {
+                "param_" + k: v
+                for params in cv_results["params"]
+                for k, v in params.items()
+            }
+        )
+        cv_results = {k: np.array(v) for k, v in cv_results.items()}
+
+        order = (-1 * cv_results["test_score"]).argsort()
```
TomAugspurger: Could replace this with

```python
        cv_results = {k: np.array(v) for k, v in cv_results.items()}
        cv_results["rank_test_score"] = scipy.stats.rankdata(
            -cv_results["test_score"], method="min"
        ).astype(int)
        return cv_results
```

With `method="min"`, we assign ties the same rank. I'm not sure what scikit-learn does in the case of ties, but this seems more correct. It makes the assertion below that `assert search.cv_results_["rank_test_score"][search.best_index_] == 1` always true.

tests/model_selection/test_incremental.py
stsievert: I can get this to fail to on my local machine. When I run the tests until failure then plot `search.history_` for a subset of model IDs, I get

screen shot 2018-10-15 at 10 25 36 am

Model 10 had the highest final score, so the same assert statement fails, but model 16 is the model corresponding to `search.best_index_`. This happens because the scores decrease over time. I view this as a test of the adaptive algorithm, not a test of basic `IncrementalSearchCV` functionality.

I believe the search in this test should be passive, or use `decay_rate=0` because of the small dataset and simple model. The adaptive algorithm removed some models, and the score decreased over time (=> models removed had higher scores than retained models).

I think there are two methods to resolve it:

1. Set `decay_rate=0` (which would require parts of #406, 8c7153d929c5fc67355301f26e9be566944b9ec0 and b5ce47c1d1bb973a46a471aac14c30d4d997499b).
2. Implement the test change in https://github.com/dask/dask-ml/pull/370#issuecomment-428788259 then revert it in #406.

TomAugspurger: I don't think a probabilistic test like implemented in 
https://github.com/dask/dask-ml/pull/370#issuecomment-428788259 is good.

I think on IncrementalSearchCV we can't really make any claims about the relative scores of the models. We can't claim that the "best" model (the model who was trained through to the last round) has the highest score (average or final). I think we just drop this assert from the test.

stsievert: > I don't think a probabilistic test like implemented in #370 (comment) is good.

I agree.

> I think on IncrementalSearchCV we can't really make any claims about the relative scores of the models

For passive incremental searches, we can make claims about the relative scores of models.
For adaptive incremental searches, we can not make claims about the relative scores of models.

We can do this because completely passive searches (`patience=False`) don't drop models. Every model is fitted for `max_iter` calls to `partial_fit`. All of these models are considered in `_get_best`, and the model with the highest final score is selected.

I'll merge 8c7153d929c5fc67355301f26e9be566944b9ec0 and b5ce47c1d1bb973a46a471aac14c30d4d997499b later today.

TomAugspurger: Are  8c7153d and b5ce47c related to this specific issue?

We should have a test with the default params, and I think this is a good test, aside from the assertion that `search.cv_results_["test_score"][search.best_index_]` has at least the highest score.

dask_ml/model_selection/_incremental.py
TomAugspurger: Scikit-learn also uses rankdata https://github.com/scikit-learn/scikit-learn/blob/e0e738760625129be71839fac2dad9325fb90225/sklearn/model_selection/_search.py#L24

TomAugspurger:
74d7739 changes the assert to just be basic sanity checks on the results. We can followup with stronger asserts later if we can justify them.

dask_ml/model_selection/_incremental.py
```diff
@@ -459,43 +463,61 @@ def _get_params(self):
...
+        cv_results = {}
+        for k, hist in model_hist.items():
+            pf_times = list(toolz.pluck("partial_fit_time", hist))
+            score_times = list(toolz.pluck("score_time", hist))
+            cv_results[k] = {
+                "mean_partial_fit_time": np.mean(pf_times),
+                "mean_score_time": np.mean(score_times),
+                "std_partial_fit_time": np.std(pf_times),
+                "std_score_time": np.std(score_times),
+                "test_score": best_scores[k],
```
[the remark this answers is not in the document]
TomAugspurger: I think this is fine (and useful) as long as we document it clearly.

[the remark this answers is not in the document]
stsievert: The other quantities are all assumed to be static regardless of the number of `partial_fit` calls. However, `score` typically increases with the number of `partial_fit` calls.

Scikit-learn defines `mean_test_score` to be "mean test score over different test sets", not across time.

stsievert:
> 74d7739 changes the assert to just be basic sanity checks on the results. We can followup with stronger asserts later if we can justify them.

I came to the same conclusion; I can make these changes in #406 or when #388 is resolved.

TomAugspurger:
@mrocklin do you have any thoughts on this PR?

TomAugspurger:
I'd like to merge this today if possible.

@stsievert one confirmatory question:

Do we pick the same model as 'best' on this branch as we did on master? I think we still do. I think the "best" model is still the one that has the highest validation score of all the models trained in the last "round". Is that correct?

mrocklin:
> I think we still do. I think the "best" model is still the one that has the highest validation score of all the models trained in the last "round". Is that correct?

Is this preferable to the model with the best validation score?  What should behavior be if model performance degrades with additional training?  (although this may not be a common occurrence)

mrocklin:
I have no strong thoughts about the cv_history stuff.  I don't have enough experience with this topic.

I do remain generally confused about why we need to search through history for this stuff though.  I would expect that this would be something that we could return easily from the higher level `fit` function.  I've brought this up before though, so I suspect that there is something incorrect with my reasoning here.

mrocklin:
Fair enough.  It's an idle comment.  Feel free to proceed.

TomAugspurger:
Roughly

```diff
diff --git a/dask_ml/model_selection/_incremental.py b/dask_ml/model_selection/_incremental.py
index 313a1af..a810aa7 100644
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -23,7 +23,7 @@ from tornado import gen
 from ..utils import check_array
 from ._split import train_test_split
 
-Results = namedtuple("Results", ["info", "models", "history"])
+Results = namedtuple("Results", ["info", "models", "history", "best"])
 
 
 def _partial_fit(model_and_meta, X, y, fit_params):
@@ -256,12 +256,14 @@ def _fit(
     models = {k: client.submit(operator.getitem, v, 0) for k, v in models.items()}
     yield wait(models)
 
+    best = max(scores.items(), key=operator.itemgetter(1))
+
     info = defaultdict(list)
     for h in history:
         info[h["model_id"]].append(h)
     info = dict(info)
 
-    raise gen.Return(Results(info, models, history))
+    raise gen.Return(Results(info, models, history, best))
 
 
 def fit(
@@ -553,12 +555,15 @@ class BaseIncrementalSearchCV(BaseEstimator, MetaEstimatorMixin):
             random_state=self.random_state,
         )
         results = self._process_results(results)
-        model_history, models, history = results
+        model_history, models, history, bst = results
 
         cv_results = self._get_cv_results(history, model_history)
         best_idx, best_est = self._get_best(results, cv_results)
         best_estimator = yield best_est
 
+        self.best_idx_ = toolz.first(bst)
+        self.best_est_ = yield models[toolz.first(bst)]
+
         # Clean up models we're hanging onto
         ids = list(results.models)
         for model_id in ids:
diff --git a/tests/model_selection/test_incremental.py b/tests/model_selection/test_incremental.py
index 58193b3..bed83ee 100644
--- a/tests/model_selection/test_incremental.py
+++ b/tests/model_selection/test_incremental.py
@@ -250,6 +250,8 @@ def test_search_basic(c, s, a, b):
     decision = search.decision_function(X_)
     assert decision.shape == (1000,)
 
+    assert 0
+
 
 @gen_cluster(client=True, timeout=None)
 def test_search_patience(c, s, a, b):
```

would I think replace `_get_best` entirely.

stsievert:
> Do we pick the same model as 'best' on this branch as we did on master?

Yes, with a couple small improvements. Master picks any model retained by `_additional_calls` as the best, this PR picks the model with highest validation score as the best.

The implementation in https://github.com/dask/dask-ml/pull/404#issuecomment-430264859 look good.

TomAugspurger:
It occurs to me that subclass are free to override `_fit` with something like

```python
def _fit(self, X, y, **fit_params):
    super()._fit(X, y, **fit_params)
    my_best = compute_my_best(self.history_)
    self.best_estimator_ = ...
```

if they need to dig through history to find the best fort them.

TomAugspurger:
I think https://dev.azure.com/pandas-dev/pandas/_settings/agentqueues?queueId=3&_a=details may have the info for that worker, if it means anything to anyone.

dask_ml/model_selection/_incremental.py
```diff
@@ -256,7 +258,15 @@ def get_futures(partial_fit_calls):
 
     models = {k: client.submit(operator.getitem, v, 0) for k, v in models.items()}
     yield wait(models)
-    raise gen.Return(Results(info, models, history))
+    scores = yield client.gather(scores)
```
TomAugspurger: @mrocklin is awaiting the scores here a bad idea?

mrocklin: Seems ok to me

TomAugspurger:
Thanks!

PR #405, opened 2018-10-15
DOC: IncrementalSearchCV

stsievert:
This PR

* adds in "Attributes" section to `IncrementalSearchCV`
* adds an explanation of `patience`/`tol`
* changes `max_iter=100` by default (mentioned in the docs, but not implemented correctly?)

This PR is based off of #404. A diff between them can be seen at https://github.com/stsievert/dask-ml/pull/1

### DIFF
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -716,6 +716,72 @@ class IncrementalSearchCV(BaseIncrementalSearchCV):
 
         If None, the estimator's default scorer (if available) is used.
 
+    Attributes
+    ----------
+    cv_results_ : dict of np.ndarrays
+        This dictionary has keys
+
+        * ``mean_partial_fit_time``
+        * ``mean_score_time``
+        * ``std_partial_fit_time``
+        * ``std_score_time``
+        * ``test_score``
+        * ``rank_test_score``
+        * ``model_id``
+        * ``partial_fit_calls``
+        * ``params``
+        * ``param_{key}``, where ``key`` is every key in ``params``.
+
+        The values in the ``test_score`` key correspond to the last score a model
+        received on the hold out dataset. The key ``model_id`` corresponds with
+        ``history_``. This dictionary can be imported into Pandas.
+
+    model_history_ : dict of lists of dict
+        A dictionary of each models history. This is a reorganization of
+        ``history_``: the same information is present but organized per model.
+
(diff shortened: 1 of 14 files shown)


mrocklin:
The docs stuff here seems fine to me.  Did you mean to also include the cv_results changes?

mrocklin:
> The docs stuff here seems fine to me. Did you mean to also include the cv_results changes?

Ah, I see as you said above you're basing this off of #404 .  I'll wait until that one goes in before I review this.

TomAugspurger:
Merged master.

dask_ml/model_selection/_incremental.py
```diff
@@ -705,6 +706,72 @@ class IncrementalSearchCV(BaseIncrementalSearchCV):
...
+        * ``model_id``
+        * ``params``
+        * ``partial_fit_calls``
+
+        The key ``model_id`` corresponds to the ``model_id`` in ``cv_results_``.
+        This list of dicts can be imported into Pandas.
+
+    best_score_ : float
+        Best score achieved on the hold out data, where "best" means "the highest
+        score on the hold out set after a model's last partial fit call."
```
TomAugspurger: So this isn't quite accurate right now, is it? We could have a case of

Model | Round | Score
----- | ----- | -----
A     | 0     | 0.8
B     | 0     | 0.75
A     | 1     | 0.7
A     | 2     | 0.65

Model A is the chosen as the "best". How about something like the following:

```
best_score_ : float
    The best model's score on the hold-out data in the final round
    of training. The "best" model is the model with the highest
    final score, among all models being trained in the final round.
```

stsievert: Maybe 

```
best_score_ : float
    The best model's score on the hold-out data in the final round
    of training.
best_estimator_ : BaseEstimator
    The model with highest final validation score among all the models
    retained by the "inverse decay" algorithm mentioned above.
```

I'm thinking of "above" to be https://github.com/dask/dask-ml/pull/405/files#diff-0a54ded74c8013caf0588e9a5c879e99R628

stsievert: Another edit:

```
best_estimator_ : BaseEstimator
    The model with highest final validation score among all the models
    retained by the "inverse decay" algorithm mentioned above.
best_score_ : float
    The score that `best_estimator_` achieved on the validation set
    after the final `partial_fit` call.
```

PR #406, opened 2018-10-15
MAINT, BUG, TST: incremental API cleaning

stsievert:
This PR

* resolves a bug with `IncrementalSearchCV` where it would be stuck in an infinite loop when `decay_rate=0`.
* defines `patience` to be the number of calls to `partial_fit`
* implement a test for `tol`
* add to the `patience` test

This PR depends on #404. A clean diff can be found at https://github.com/stsievert/dask-ml/pull/2.

### DIFF
--- a/dask_ml/compose/_column_transformer.py
+++ b/dask_ml/compose/_column_transformer.py
@@ -1,3 +1,5 @@
+import warnings
+
 import dask.array as da
 import dask.dataframe as dd
 import numpy as np
@@ -179,7 +181,9 @@ boolean mask array or callable
         if self.sparse_output_:
             return sparse.hstack(Xs).tocsr()
         elif dd.Series in types or dd.DataFrame in types:
-            return dd.concat(Xs, axis="columns")
+            with warnings.catch_warnings():
+                warnings.filterwarnings("ignore", "Concatenating", UserWarning)
+                return dd.concat(Xs, axis="columns")
         elif da.Array in types:
             return da.hstack(Xs)
         elif self.preserve_dataframe and (pd.Series in types or pd.DataFrame in types):
--- a/dask_ml/feature_extraction/text.py
+++ b/dask_ml/feature_extraction/text.py
@@ -37,15 +37,13 @@ class HashingVectorizer(sklearn.feature_extraction.text.HashingVectorizer):
 
         See the :doc:`examples/text-vectorization` for more.
         """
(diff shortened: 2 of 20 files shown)


TomAugspurger:
Merged master.

stsievert:
> (if we deem that necessary).

There aren't many user-facing changes in this PR. I think it updating this could be a point release. It improves on the implementation (`decay_rate=0` is allowed and defines `patience` to be what the user expects (but they're currently the same thing with default params).

The largest user facing issue is with the bug this PR solves (it removes an infinite loop if `decay_rate=0`).

TomAugspurger:
I think that infinite loop was fixed in https://github.com/dask/dask-ml/pull/373

stsievert:
> I think that infinite loop was fixed in #373 

I don't think so. It's with `decay_rate` specifically, and involves the property that for all x > 0, `1 / x**0 == 1`.

TomAugspurger:
Gotcha. I'll try splitting that bug fix out.

TomAugspurger:
I don't think this PR fixes the `decay_rate=0` bug. AFAICT, `test_search_basic` isn't really being run now, since the tests exists as soon as the test helper `_test_search_basic` is called. I believe you need to await it.

```diff
diff --git a/tests/model_selection/test_incremental.py b/tests/model_selection/test_incremental.py
index 5d828b7..c52b06a 100644
--- a/tests/model_selection/test_incremental.py
+++ b/tests/model_selection/test_incremental.py
@@ -196,9 +196,10 @@ def test_explicit(c, s, a, b):
 @gen_cluster(client=True)
 def test_search_basic(c, s, a, b):
     for decay_rate in {0, 1}:
-        _test_search_basic(decay_rate, c, s, a, b)
+        yield _test_search_basic(decay_rate, c, s, a, b)
 
 
+@gen.coroutine
 def _test_search_basic(decay_rate, c, s, a, b):
     X, y = make_classification(n_samples=1000, n_features=5, chunks=(100, 5))
     model = SGDClassifier(tol=1e-3, loss="log", penalty="elasticnet")
```

TomAugspurger:
Alrighty, I had a look and I think these changes look good. I like thinking about `patience` in terms of the number of `partial_fit_calls`.

Planning to merge this evening.

TomAugspurger:
Thanks @stsievert!

PR #411, opened 2018-10-17
Avoid infinite loop in IncrementalSearchCV with decay=0

TomAugspurger:
This avoids the loop, and hopefully does the correct thing.

IIUC, `decay=0` should be "just keep training all the models".


TomAugspurger:
cc @mrocklin @stsievert if you could give a second opinion when you get a chance.

mrocklin:
Looks good to me

PR #412, opened 2018-10-17
API: rename IncrementalSearchCV to AdaptiveSearchCV

### LINKED ISSUE #388: IncrementalSearchCV runs an adaptive algorithm by default

stsievert:
This PR renames `IncrementalSearchCV` to `AdaptiveSearchCV`.

Renaming to `AdaptiveSearchCV` would close #388: how can `AdaptiveSearchCV` not be adaptive by default? I think the name `AdaptiveSearchCV` aligns with the design goal: to use to previous calls to `partial_fit` to determine if training continue.

I think this will help provide clearer recommendations to the user because the name succinctly describes the features and design goals of this class.

I think some documentation changes should go around this too (I can add a WIP label, but documentation shouldn't be blocking).

### DIFF
--- a/dask_ml/model_selection/__init__.py
+++ b/dask_ml/model_selection/__init__.py
@@ -19,8 +19,8 @@ __all__ = [
 
 
 try:
-    from ._incremental import IncrementalSearchCV  # noqa: F401
+    from ._incremental import AdaptiveSearchCV  # noqa: F401
 
-    __all__.extend(["IncrementalSearchCV"])
+    __all__.extend(["AdaptiveSearchCV"])
 except ImportError:
     pass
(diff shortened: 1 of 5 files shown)


stsievert:
`IncrementalSearchCV` is renamed to `AdaptiveSearchCV` in every place in the source (verified by running grep). The tests should pass too (the only failing test in https://github.com/dask/dask-ml/pull/412/commits/9a696e2d2d2ac46b4d1c4911ab8402329a7e10fb is with black, and I've fixed that).

stsievert:
I think this is closed by https://github.com/dask/dask-ml/issues/388#issuecomment-431621311. I'm planning on implementing that comment as part of #221.

PR #417, opened 2018-10-23
Bug-Fix in Polynomial-Features

datajanko:
While working on #366 I realized that the PolynomialFeatures transformer does not respect the passed parameters correctly. The PR fixes this and provides a test

PR #418, opened 2018-10-23
Remove get= keyword

jrbourbeau:
This PR removes an occurrences of the `get=` keyword in `dask-ml`

Related to error in #417

### DIFF
--- a/ci/environment-2.7.yml
+++ b/ci/environment-2.7.yml
@@ -19,7 +19,7 @@ dependencies:
   - pytest
   - pytest-cov
   - pytest-mock
-  - python=2.7
+  - python=2.7.14
   - sortedcontainers
   - scikit-learn>=0.20.0
   - scipy
--- a/ci/environment-3.6.yml
+++ b/ci/environment-3.6.yml
@@ -27,7 +27,7 @@ dependencies:
   - pytest
   - pytest-cov
   - pytest-mock
-  - python=3.6
+  - python=3.6.6
   - sortedcontainers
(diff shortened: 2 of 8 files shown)


jrbourbeau:
The failing test here should be fixed when https://github.com/dask/dask-glm/pull/72 is merged into `master`

TomAugspurger:
haven't looked closely, but pandas had an issue with the latest CPython point releases and matplotlib (possibly related to CPython bumping the version of openssl, but Qt being built with an older version).

For now, we can maybe pin to the previous point release (2.7.14 and 3.6.6 I think).

jrbourbeau:
Thanks @TomAugspurger! You beat me to the warnings

TomAugspurger:
Hmm, https://github.com/dask/dask/commit/11d30b5154820289bfa58fd5bdf19844f5d10add may have broken some things here...

TomAugspurger:
I haven't looked closely at the failures or the changes there, and I'm going offline for a few hours now.

Happy to take a look once I get back though.

mrocklin:
It sounds like that change might need to be reverted in some cases.  Presumably there are valid use cases that we're now disallowing.

mrocklin:
I think it's generally better to raise errors early at graph construction
time if possible.  That way users get a better sense of which of their
lines of code causes the error.

Thinking about this a bit more.  I suspect that there are cases where
argmin is fine, for example if we aggregate along some non-nan axis, or if
there is a single chunk.  I'm a bit curious what dask-ml was doing that was
ok in this situation.

PR #419, opened 2018-10-26
BUG: properly store 'param_*' keys in incremental search

stsievert:
This fixes a bug where each `param_*` key in `cv_results` was being overwritten by previous values.

A quick illustration of the issue:

``` python
>>> search = IncrementalSearchCV(..., n_initial_parameters=3)
>>> search.fit(X, y)
>>> search.cv_results_["param_alpha"]
array(0.3)  # expected: array([0.1, 0.2, 0.3]) because 3 model ids
>>> search.cv_results_["model_id"]
array([0, 1, 2])
```

There are a couple lines added to cover the case when some parameter sets have different keys. I should probably test that -- I think it's a small but non-negligible corner case.


stsievert:
I've ripped out the logic for missing parameters. The classes that inherit from `BaseIncrementalSearchCV` either use either `ParameterGrid` or `ParameterSampler`, so they all have the same parameters.

TomAugspurger:
What's the expected output for param_alpha in your little example? An ndarray with 3 elements?

tests/model_selection/test_incremental.py
```diff
@@ -267,6 +269,30 @@ def _test_search_basic(decay_rate, c, s, a, b):
...
+            pass
+        def partial_fit(self, *args, **kwargs):
+            pass
+
+    params = [{"value": 0.1}, {"rate": 0.2}]
+    model = SetKwargs()
+
+    search = BaseIncrementalSearchCV(model, params)
+    search.fit(X, y)
+    1 / 0
```
TomAugspurger: debugging?

stsievert: Yup, and on a test I didn't need too.

stsievert:
Yup, I expected an array of 3 number, not just a scalar. I've updated https://github.com/dask/dask-ml/pull/419#issue-225955062 to reflect that.

stsievert:
The tests are passing on CI, except for errors unrelated to this PR.

stsievert:
The CI tests for this PR are passing.

TomAugspurger:
Thanks!

Issue #366, opened 2018-09-14
Add a BlockTransformer helper


TomAugspurger:
This would be similar in spirit to http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html#sklearn.preprocessing.FunctionTransformer

Users provide a `func` that is assumed to operate on ndarrays / pandas dataframes. The `BlockTransformer` will take care of mapping it to each block / partition.

We may need some information from the user when the function changes dtype / shape.

datajanko:
May I work on this as well?

TomAugspurger:
I was going to suggest it for you. I think it'll end up being quite similar to the `PolynomialTransformer`. In the end, you may be able to re-implement the PolynomialTransformer as an instance of `BlockTransformer`, but maybe not. LMK if you need help getting started.

datajanko:
Perfect. I also think it's very similar.

datajanko:
Assume you have a function like

```
def f(arr):
    return arr[:,[0]]+arr[:,[1]]
```
putting such a function into `map_blocks` will only work if all blocks have at least a width of 2.
As data frames are only partitions w.r.t rows,  I wouldn't have a problem width allowing functions that can operate on full rows.

I think one could then also infer the output shape, by simply running the function on the first row (and potentially making the transformer "stateful" only to be able to throw errors).

So what kinds of functions should be applicable here?

datajanko:
@TomAugspurger any thoughts on my last comment? which kind (input shape, output shape) of functions shall we allow? I think it will be quite easy, to accept functions that use information on a specific row or just on a cell will be quite easy (Poly transformer takes information from a row). However, we'd need to enforce for numpy arrays, that blocks have the width of the array to avoid effects as shown above.

Functions that take the information of all rows for a column probably will be more difficult to implement.

TomAugspurger:
The user-provided function should operate on ndarrays / pandas data frames.

In BlockTransformer.transform, dask-ml will ensure that dask Arrays are only partitioned along the samples using `dask_ml.utils.check_array` before calling `X.map_blocks`.

I'd say let's just handle the easy things right now. Anything where dask's metadata inference fails can I think be ignored for now.

> Functions that take the information of all rows for a column probably will be more difficult to implement.


Right, those functions wouldn't work with this helper. This is just for the class of estimators where 

1. They're stateless (fit does nothing)
2. They can transform blockwise

datajanko:
[...]

PR #420, opened 2018-10-26
Adds Block Transformer (Implements #366)

datajanko:
Tries to implement #366.

Sorry, for starting after such a long time with such a small implementation.
I encountered some questions/points for discussion for the most suitable implementation (please apologize my inexperience here).

In the thread of the issue @TomAugspurger suggested to call check_array in transform. However, the scikit-learn's `FunctionTransformer`. Suggestions on what to do here? Of course we could avoid the fit function of FunctionTransformer entirely. This will lead to additional overhand in `inverse_transform` as we need to validate there as well. In my opinion, this would even make sense as it seems a bit illogical to only check the input on the set needed to fit the estimator.

What about DataFrames? Should they be included as well?

ToDo's:

- [X] clarify closeness to functiontransformer
- [X] implement further types if desired
- [X] add documentation
- [X] add tests
- [x] adds example for pass through keywords and also a test


dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,51 @@
...
+            inverse_func=inverse_func,
+            validate=validate,
+            accept_sparse=accept_sparse,
+            check_inverse=check_inverse,
+            kw_args=kw_args,
+            inv_kw_args=inv_kw_args,
+        )
+
+    def fit(self, X):
+        return self
```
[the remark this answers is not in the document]
datajanko: The fit function of `FunctionTransformer` performs the input validation. You suggested, that we should perform the validation before the transformation. The fit function performs some tasks on a numpy array. Thus, extending to dask arrays, we need to override fit

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,51 @@
...
+from pandas.api.types import is_categorical_dtype
+from scipy import stats
+from sklearn.preprocessing import FunctionTransformer
+from sklearn.utils.validation import check_is_fitted, check_random_state
+
+from dask_ml._utils import copy_learned_attributes
+from dask_ml.utils import check_array, handle_zeros_in_scale
+
+
+class BlockTransformer(FunctionTransformer):
```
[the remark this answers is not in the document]
datajanko: essentially, it would give us `fit`, `transform`, `inverse_transform` and further utility functions (used in those functions) we could pass directly to `map_blocks`. I had the impression, that I should follow the pattern we used in the `PolynomialFeatures`, i.e. just using scikit-learn class properly in combination with `map_blocks` etc.

datajanko:
So a short summary of my and your comments.

There are two alternatives we could use to implement the functionality.

1. Just implement from scratch, maybe even dropping `inverse_transform`
2. Inherit from `FunctionTransformer` and follow the implementation there.

Concerning the scikit-learn option: Currently, there are some deprecations in the interface of `FunctionTransformer`. Additionally, only the training input set is validated, which seems a bit odd in my opinion.

In light of that, I'd suggest to take the first route as the code as we avoid a rather unnecessary dependency here.

datajanko:
Sorry, I just found some time again to work on this. I decided to go for the "simplest" possible transformer. Thus, I'm not inheriting from FunctionTransformer.

Questions right now: What to do with dataframes?
I will work on the tests later this week.

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -5,47 +5,56 @@
...
+        self.func = func
+        self.validate = validate
+        self.accept_sparse = False
+        self.kw_args = kw_args
 
     def fit(self, X):
         return self
 
     def transform(self, X):
+        def func_forwarded(arr):
```
datajanko: defining this function, will use the name 'func_forwared' in the desk computation graph: `dask.array<func_forwarded, shape=(5, 2), dtype=int64, chunksize=(3, 2)>` is there a suggested way to solve this issue here? Probably, it would have been better to wrap the function properly and use the `@wraps` decorator. Right idea?

TomAugspurger:
No rush on this @datajanko, we're all a bit busy I think :)

Feel free to ignore dataframes if it seems too messy, though I *think* that `.transform` should be relatively straightforward. `inverse_transform` may be harder.

dask_ml/preprocessing/_block_transformer.py
TomAugspurger: Hmm, something like that.

Can you instead do

```python
kw_args = self.kw_args or {}
# array
XP = X.map_blocks(self.func, dtype=X.dtype, chunks=X.chunks, **kw_args)
# dataframe
...

```

datajanko: I'll try that, and investigate if `map_blocks` passes the arguments correctly to `func`

datajanko:
Actually, I think transform is more important right now. data frame, shouldn't be too messy as well.

datajanko:
I think I implemented the "forward" direction of the Transformer. If desired, one could add the inverse transform as well.

Moreover, I realized that `preserve_dataframe` and `validate` do not play so nicely together. This also implies, that one needs to add some more tests for PolynomialFeatures. If desired I can do this of course.

Are the tests sufficient so far?
I think it would make sense to add a column parameter for this transformer. What is the expected output here? just modify the columns in question, return only the columns in question or something else?

Any ideas suggestion on how to fix the broken tests?

TomAugspurger:
Sorry for the delay. A bit busy trying to unblock a pandas release. I'll
get to this eventually :)

datajanko:
So apparently, this has not been solved yet. What's the state here? Is there still interest on this topic? Sorry I also forgot this 🙈

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+    def fit(self, X):
+        return self
+
+    def transform(self, X):
+        kwargs = self.kw_args if self.kw_args else {}
+
+        if isinstance(X, da.Array):
+            if self.validate:
+                X = check_array(X, accept_dask_array=True, accept_unknown_chunks=True)
+            XP = X.map_blocks(self.func, dtype=X.dtype, chunks=X.chunks, **kwargs)
```
TomAugspurger: I don't think we know that `dtype=X.dtype= is correct. That's only true if the user-defined function preserves the dtype.

tests/preprocessing/test_block_transformer.py
```diff
@@ -0,0 +1,67 @@
+import dask
+import dask.array as da
+import dask.dataframe as dd
+import numpy as np
+import pandas as pd
+import pytest
+from dask.array.utils import assert_eq as assert_eq_ar
```
TomAugspurger: Rather than aliasing, can you just use `da.utils.assert_eq` and `dd.utils.assert_eq`?

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+        Dictionary of additional keyword arguments to pass to func.
+    """
+
+    def __init__(self, func, *, validate=True, kw_args=None, preserve_dataframe=True):
+        self.func = func
+        self.validate = validate
+        self.kw_args = kw_args
+        self.preserve_dataframe = preserve_dataframe
+
+    def fit(self, X):
```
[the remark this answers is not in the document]
datajanko: adds `y=None` to `fit` and `transform`

dask_ml/preprocessing/_block_transformer.py
datajanko: Got removed

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,63 @@
...
+from sklearn.base import BaseEstimator, TransformerMixin
+
+from dask_ml.utils import check_array, handle_zeros_in_scale
+
+
+class BlockTransformer(BaseEstimator, TransformerMixin):
+    """Construct a transformer from a an arbitrary callable
+
+    The BlockTransformer forwards the blocks of the X arguments to a user-defined
+    function or function-object an returns the result of this operation.
```
TomAugspurger: function or function-object -> callable.

an -> and.

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,63 @@
...
+    This is useful for stateless operations, that can be performed on the cell or
+    block level, such as taking the log of frequencies. In general the transformer
+    is not suitable for e.g. standardization tasks as this requires information for
+    a complete columns
```
TomAugspurger: Missing a "."

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,70 @@
...
+
+    validate : bool, optional default=True
+         Indicate that the input X array should be checked before calling
+        ``func``.
+
+    kw_args : dict, optional
+        Dictionary of additional keyword arguments to pass to func.
+    """
+
+    def __init__(self, func, *, validate=True, kw_args=None, preserve_dataframe=True):
```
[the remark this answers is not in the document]
datajanko: Hmm, I was removing this and received some errors as check_array typically converts data back to an array. This is called if `validate = True`. However, now the tests seem to run fine, though I'm not returning a pandas data frame. Is there a `check_dataframe` function or something similar? Suggestions?

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,63 @@
...
+    ----------
+    func : callable
+        The callable to use for the transformation.
+
+    validate : bool, optional default=True
+         Indicate that the input X array should be checked before calling
+        ``func``.
+
+    kw_args : dict, optional
+        Dictionary of additional keyword arguments to pass to func.
```
TomAugspurger: An example here would be welcome. Perhaps a DataFrame hasher like

```python
import dask.datasets
import pandas as pd

df = dask.datasets.timeseries()
df

trn = BlockTransformer(pd.util.hash_pandas_object)
trn.transform(df)
```

TomAugspurger: And you can show an example like `index=False` for passing through kwargs.

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,63 @@
...
+    block level, such as taking the log of frequencies. In general the transformer
+    is not suitable for e.g. standardization tasks as this requires information for
+    a complete columns
+
+    Parameters
+    ----------
+    func : callable
+        The callable to use for the transformation.
+
+    validate : bool, optional default=True
```
TomAugspurger: scikit-learn is switching to `validate=False` by default. I think we should do the same. https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html

dask_ml/preprocessing/_block_transformer.py
datajanko: and and additional "s" in columns :)

dask_ml/preprocessing/_block_transformer.py
datajanko: good idea

dask_ml/preprocessing/_block_transformer.py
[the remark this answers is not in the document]
TomAugspurger: There's just `check_array` right now, which will optionally allow pandas / dask dataframes with `accept_dask_dataframe=True, preserve_pandas_dataframe=True`.

I think the user's function should get the blocks / partitions of whatever dask collection is inside.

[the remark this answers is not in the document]
datajanko: Sorry, I just realized, that `check_array` on a Dask data frame returns the Dask data frame in any case. So this is resolved :)

datajanko:
please check again. I added your example - I didn't build the docs though and refactored the tests to contain kwargs pass through. However, I defined a helper function which feels a bit off here. Any suggestion how to do things more idiomatic and using pytest better?

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -0,0 +1,76 @@
...
+    kw_args : dict, optional
+        Dictionary of additional keyword arguments to pass to func.
+
+    Examples
+    --------
+
+    >>> import dask.datasets
+    >>> import pandas as pd
+    >>> df = dask.datasets.timeseries()
+    >>> df
```
TomAugspurger: This should be like

```
>>> import dask.datasets
>>> import pandas as pd
>>> df = dask.datasets.timeseries()
>>> df  # doctest: +SKIP
Dask DataFrame Structure:
                   id    name        x        y
npartitions=30
2000-01-01      int64  object  float64  float64
2000-01-02        ...     ...      ...      ...
...               ...     ...      ...      ...
2000-01-30        ...     ...      ...      ...
2000-01-31        ...     ...      ...      ...
Dask Name: make-timeseries, 30 tasks
```

Basically, the input *and output* of a python session.

May also want `from dask_ml.preprocessing import BlockTransformer` at the start of the example.

TomAugspurger:
The tests look fine, having a helper at the top of the test module is not a problem.

dask_ml/preprocessing/_block_transformer.py
datajanko: is there a way to quickly build the documentation and maybe only parts of it?

Initially I added the output as you did but the dockets just failed. (didn't use skip though).

datajanko:
Ok, so I was somehow able to generate docs (slowly). skipping for first doctest is still there. Otherwise, I think I addressed your comments. What's next?

datajanko:
Strange. Timeout when downloading newsgroups 20. Is there a way to retriever those tests? Tests didn't fail earlier

datajanko:
thanks for retriggering. Next steps?

TomAugspurger:
I think we're good, thanks!

PR #422, opened 2018-10-29
Minor glm updates

jrbourbeau:
This PR makes a few of minor changes to the `_GLM` base estimator. Namely: 

- Adds `max_iter` parameter to the docstring

- Renames the `multiclass` keyword to `multi_class` to match the corresponding scikit-learn keyword (ref: [scikit-learn's LogisticRegression](http://scikit-learn.org/dev/modules/generated/sklearn.linear_model.LogisticRegression.html)). Even though this keyword is currently ignored in dask-ml, I still think it's worth being consistent with scikit-learn. 

- Moves the position of the `max_iter` keyword the `_GLM.__init__` signature to have the same order as scikit-learn.


TomAugspurger:
Thanks!

PR #424, opened 2018-11-02
API: Lazy score, predict for IncrementalSearchCV

### LINKED ISSUE #423: IncrementalSearchCV should inherit from ParallelPostFit
I haven't tried it out yet, but I think this should work.

That'd make the handling of `Incremental(...).score` and `IncrementalSearchCV(...).score` consistent.

TomAugspurger:
This changes the API of IncrementalSearchCV to match
Incremental for postfit tasks like predict, score, etc.
For Dask inputs, these return Dask outputs, so the result is
lazy and computed in parallel.

Closes #423


TomAugspurger:
Merging later today if no objections.

PR #425, opened 2018-11-05
Changing seed dtype

EmanuelFontelles:
I got the error

`ValueError: high is out of bounds for int32`

Probably because the size of seed generator isn't correct. I changed just `train_test_split` method, but I think that have the same issue in the others methods. I think the numpy type isn't correct, probably we are importing numpy 32 bits.

I think 32 bits could solve but 64 bits could be more correct.

### DIFF
--- a/dask_ml/model_selection/_split.py
+++ b/dask_ml/model_selection/_split.py
@@ -428,7 +428,7 @@ def train_test_split(*arrays, **options):
             blockwise = False
 
         rng = check_random_state(random_state)
-        rng = draw_seed(rng, 0, 2 ** 32 - 1, dtype="uint")
+        rng = draw_seed(rng, 0, 2 ** 32 - 1, dtype="uint32")
         return list(
             itertools.chain.from_iterable(
                 arr.random_split([train_size, test_size], random_state=rng)


TomAugspurger:
Do you get that value error on master? I don't see how your change would fix that error.

On 32-bit platforms, `"uint"` should be `"uint32"`.
On 64-bit platforms, it'll be `"uint64"`, which will hold the values we draw.

EmanuelFontelles:
> Do you get that value error on master? I don't see how your change would fix that error.
> 
> On 32-bit platforms, `"uint"` should be `"uint32"`.
> On 64-bit platforms, it'll be `"uint64"`, which will hold the values we draw.

There is a error with numpy types. My pc works fine with uint64 but when I ran `train_test_split` I got the error `ValueError: high is out of bounds for int32` but my current numpy installation is 64 bits. I think the size of random seed was 32 bits which is incompatible with numpy arrays.

In this case I mean `"uint64"`. Do you understand?

TomAugspurger:
Can you paste the output of

1. A small code snippet and the full traceback for `ValueError: high is out of bounds for int32`. Your fix here isn't consistent with that error message, since you're changing an `uint` dtype, but the error message says `int`.
2. The output of `dask_ml.__version__`
3. The output of `dask_ml._utils.draw_seed(np.random.RandomState(0), 0, high=2**32, dtype='uint')`

Thanks.

TomAugspurger:
Gotcha, thanks.

PR #430, opened 2018-11-26
Bug: Handle a value not being passed for Y in euclidian distance.

asgersoerensen:
For euclidian distance calculation: Added handling of not passing a value for Y and initialising it to X.


TomAugspurger:
Sorry, I missed this PR when it was made.

I've merged master and added a small test for this. Will merge when the CI passes.

TomAugspurger:
Thanks! Sorry for the delay.

PR #432, opened 2018-11-30
DOC: clarify model selection issues

### LINKED ISSUE #388: IncrementalSearchCV runs an adaptive algorithm by default

stsievert:
This PR makes the following changes:

* Makes the documentation clear how different model selection problems can be constrained.
* Changes the `IncrementalSearchCV` default `decay_rate` to `decay_rate=0` (but recommends changing `decay_rate=1` in lead part of docs)
* Adds documentation on determining Hyperbands inputs and a performance comparison
* Wording change: `*SearchCV`s are "estimators" and they receive "models"
* Small documentation fixes (clearing warnings, style fixes)
* Makes the `test_same_models_with_random_state` test more stable. It passed with low probability on my machine; now it passes all the time.

The biggest change is clearly explaining the issues in model selection. It reorganizes into this:

> Two scenarios can occur during hyper-parameter optimization. The hyper-parameter search can be both
>
> 1. compute constrained
> 2. memory constrained
>
> [examples of problems]
>
> [sections mentioning classes best suited for each problem]
>
> ...
>
> [section on adaptive hyperparameter optimization]

With this documentation, changing the default `decay_rate` in `IncrementalSearchCV` makes sense (see https://github.com/dask/dask-ml/pull/432#issuecomment-443255765). If that's the default, it's the only class that fills the "memory constrained by not compute constrained" class of problems (almost; some of dask-searchcv spills in here too).

**Other issues/PRs**
This PR closes #388.

The documentation refactor has introduced the ability to add new classes, now that it's clear what the limitations of each approach are.

* https://github.com/stsievert/dask-ml/pull/3 implements a black box optimizer as mentioned in section 3.4 of [the Hyperband paper][1].

This fits well with the "computationally constrained but not memory constrained" class of model selection problems. I'm leaving it as a separate branch right now because I think ~~#221 is too bloated~~ it needs a lot more work.

[1]:https://arxiv.org/pdf/1603.06560.pdf


stsievert:
My criteria for changing the default: where would `IncrementalSearchCV` provide the most value?

I think it provides the most value in "memory but not compute constrained" problems. This means that `decay_rate` should default to 0 because this isn't compute constrained.

The reason I think it fits best here is because `IncrementalSearchCV` provides access to `partial_fit`, something dask-searchcv doesn't do. dask-searchcv fits to each chunk of the data independently ([_search.py#L416-L422]), which may not be what the user wants (and wasn't clear to me until I looked at the source). I should add to dask-searchcv's docs to reflect this.

This matches the documentation:

> If you need to do hyper-parameter optimization on larger-than-memory datasets, we recommend dask_ml.model_selection.IncrementalSearch
>
> —http://ml.dask.org/incremental.html

Without changing the default, there isn't a class that fits into "memory but not compute constrained". I don't think `DaskGridSearchCV(Incremental(model), params)` would work, especially if more than one pass through data wanted (which might warrant a change?).

TODO:

- [ ] add to dask-searchcv documentation to clarify how Dask arrays are handled
- [ ] change default `decay_rate` of `IncrementalSearchCV`
- [ ] take dask-searchcv off "memory but not compute constrained" (but add a note below).

[_search.py#L416-L422]:https://github.com/dask/dask-ml/blob/dd2c616086779018ea8aa589005753ca2c949ddb/dask_ml/model_selection/_search.py#L416-L422

dask_ml/model_selection/_incremental.py
```diff
@@ -653,10 +647,10 @@ class IncrementalSearchCV(BaseIncrementalSearchCV):
 
         Alternatively, you can set this to ``"grid"`` to do a full grid search.
 
-    decay_rate : float, default 1.0
+    decay_rate : float, default 0.0
         How quickly to decrease the number partial future fit calls.
         Higher `decay_rate` will result in lower training times, at the cost
-        of worse models.
+        of worse models. If changed, we recommend ``decay_rate == 1``.
```
TomAugspurger: Why exactly 1?

docs/source/hyper-parameter-search.rst
```diff
@@ -3,31 +3,82 @@ Hyper Parameter Search
 
 *Tools for performing hyperparameter optimization of Scikit-Learn API-compatible models using Dask*.
 
-There are two kinds of hyperparameter optimization estimators
-in Dask-ML. The appropriate one to use depends on the size of your dataset and
-whether the underlying estimator implements the `partial_fit` method.
+Issues in hyper-parameter searches
```
TomAugspurger: Maybe call this `Scaling Hyper-parameter Searches`?

docs/source/hyper-parameter-search.rst
```diff
@@ -3,31 +3,82 @@ Hyper Parameter Search
...
+----------------------------------
+Two scenarios can occur during hyper-parameter optimization. The
```
TomAugspurger: Blank line after header.

docs/source/hyper-parameter-search.rst
```diff
@@ -3,31 +3,82 @@ Hyper Parameter Search
...
 
-If your data is large and the underlying estimator implements ``partial_fit``, you can
-Dask-ML's :ref:`*incremental* hyperparameter optimizers <hyperparameter.incremental>`.
+These issues are independent and both can happen the same time. Being memory
+constrained has to do with dataset size, and being compute constrained has to
+do with estimator complexity and number of possible hyper-parameter
+combinations.
+
+An example of being compute constrained is with almost any neural network or
+deep learning framework. An example of being memory constrained is when the
```
TomAugspurger: First sentence is a bit awkward. Perhaps "Training a deep neural network is a compute-constrained problem. Training on a larger-than-memory dataset is a memory-constrained problem."

docs/source/hyper-parameter-search.rst
```diff
@@ -3,31 +3,82 @@ Hyper Parameter Search
...
+dataset doesn't fit in RAM. Dask-ML covers all 4 combinations of these two
+constraints.
+
+Neither compute nor memory constrained
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Scikit-learn handles this case.
```
TomAugspurger: Though, note that Dask-ML's pipeline re-writing / caching can still be helpful here.

docs/source/hyper-parameter-search.rst
```diff
@@ -3,31 +3,82 @@ Hyper Parameter Search
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+More detail in :ref:`hyperparameter.incremental`.
+
+.. autosummary::
+   dask_ml.model_selection.IncrementalSearchCV
+
+:class:`~dask_ml.model_selection.IncrementalSearchCV` calls ``partial_fit`` on
+each chunk of the data. :class:`~dask_ml.model_selection.GridSearchCV` and
+:class:`~dask_ml.model_selection.RandomizedSearchCV` also work with Dask
+objects, but call ``fit`` on each chunk of the data individually.
```
TomAugspurger: Will need to remove this, if it's incorrect.

docs/source/hyper-parameter-search.rst
```diff
@@ -3,31 +3,82 @@ Hyper Parameter Search
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. autosummary::
+   dask_ml.model_selection.IncrementalSearchCV
+
+This requires changing the ``decay_rate`` to be non-zero. The recommended value
+is ``decay_rate=1``.
+
+These searches can reduce time to solution by (cleverly) deciding which
+parameters to evaluate.  These searches `adapt` to history to decide which
```
TomAugspurger: Use asterisks for emphasis instead of backticks.

docs/source/hyper-parameter-search.rst
```diff
@@ -3,31 +3,82 @@ Hyper Parameter Search
...
+parameters to continue evaluating and are called "`adaptive` model selection
```
TomAugspurger: Asterisks

docs/source/hyper-parameter-search.rst
```diff
@@ -93,6 +144,8 @@ to pull it locally to your computer:
     # Pass to fit without ever leaving the cluster
     search.fit(df[['x', 'x2']], df['y'])
 
+The data is not collected to one machine because the ``estimator.fit`` is
```
TomAugspurger: May need to remove.

dask_ml/model_selection/_search.py
```diff
@@ -1130,6 +1130,13 @@ def fit(self, X, y=None, groups=None, **fit_params):
             train/test set.
         **fit_params
             Parameters passed to the ``fit`` method of the estimator
+
+        Notes
```
TomAugspurger: May need to remove if this is incorrect.

docs/source/hyper-parameter-search.rst
```diff
@@ -185,34 +238,33 @@ expensive early steps, this can be a big win when performing a grid search.
...
-
 .. note::
 
    These estimators require the optional ``distributed`` library.
 
-Here's an example training on a "large" dataset (a Dask array) with the
-``IncrementalSearchCV``.
+These are make repeated calls to the ``partial_fit`` method of the estimator.
+Naturally, these classes determine when to stop calling ``partial_fit`` by
+`adapting to previous calls`. The most basic level of this is to stop training
```
TomAugspurger: asterisks.

docs/source/hyper-parameter-search.rst
```diff
@@ -185,34 +238,33 @@ expensive early steps, this can be a big win when performing a grid search.
...
+if the score doens't improve, which ``IncrementalSearchCV`` does.
```
TomAugspurger: typo: doesn't.

docs/source/hyper-parameter-search.rst
TomAugspurger: This reads a bit awkward. Maybe:

IncrementalSearchCV implements the most basic adaptation scheme: training stops when the score stops improving.

docs/source/hyper-parameter-search.rst
```diff
@@ -185,34 +238,33 @@ expensive early steps, this can be a big win when performing a grid search.
...
 
 .. ipython:: python
 
     from dask.distributed import Client
     client = Client()
     import numpy as np
     from dask_ml.datasets import make_classification
-    X, y = make_classification(n_samples=5000000, n_features=20,
-                               chunks=100000, random_state=0)
+    # X, y = make_classification(n_samples=5000000, n_features=20,
```
TomAugspurger: A smaller example is fine, but let's remove the commented out code.

docs/source/hyper-parameter-search.rst
```diff
@@ -185,34 +238,33 @@ expensive early steps, this can be a big win when performing a grid search.
...
+    #                           chunks=100000, random_state=0)
+    X, y = make_classification(chunks=20, random_state=0)
 
-Our underlying estimator is an SGDClassifier. We specify a few parameters
+Our underlying estimator is an ``SGDClassifier``. We specify a few parameters
```
TomAugspurger: Maybe make this a reference to the scikit-learn class :class:\`sklearn.linear_model.SGDClasifier\`

docs/source/hyper-parameter-search.rst
```diff
@@ -236,8 +288,10 @@ train-and-score them until we find the best one.
 
     from dask_ml.model_selection import IncrementalSearchCV
 
-    search = IncrementalSearchCV(model, params, random_state=0)
-    search.fit(X, y, classes=[0, 1])
+    search = IncrementalSearchCV(model, params, 9, random_state=0)
+    _ = search.fit(X, y, classes=[0, 1])
```
TomAugspurger: Style nit: I prefer a trailing a semicolon to hide the output if you don't want it shown (agreed that we don't want it shown here).

docs/source/hyper-parameter-search.rst
```diff
@@ -248,16 +302,10 @@ to use post-estimation features like scoring or prediction, we recommend using
...
-   params = {'estimator__alpha': np.logspace(-2, 1, num=1000),
-             'estimator__l1_ratio': np.linspace(0, 1, num=1000),
-             'estimator__average': [True, False]}
-
-   model = ParallelPostFit(SGDClassifier(tol=1e-3,
-                                         penalty="elasticnet",
-                                         random_state=0))
-   search = IncrementalSearchCV(model, params, random_state=0)
-   search.fit(X, y, classes=[0, 1])
+   params = {'estimator__alpha': np.logspace(-2, 1, num=1000)}
```
TomAugspurger: Why these changes?

docs/source/hyper-parameter-search.rst
```diff
@@ -248,16 +302,10 @@ to use post-estimation features like scoring or prediction, we recommend using
...
+   model = ParallelPostFit(SGDClassifier(tol=1e-3, random_state=0))
+   search = IncrementalSearchCV(model, params, 9, random_state=0)
```
TomAugspurger: Use keywords for the parameters. The first two are probably fine as positional, but everything else should be named.

dask_ml/model_selection/_incremental.py
TomAugspurger: It'd be good to note that this is useful for compute-bound (& memory-bound) problems.

docs/source/hyper-parameter-search.rst
stsievert: Oooo that's nice. I didn't know that worked!

docs/source/hyper-parameter-search.rst
stsievert: I think to make it shorter. I added `estimator__l1_ratio` back in.

dask_ml/model_selection/_search.py
stsievert: I've revised this note to

> This class performs best when each cross validation split fits into RAM because fit is called on each cross validation split. For example, if using KFold with k chunks, a 1−1/k fraction of the dataset should fit into RAM.

Let me know if that's correct.

stsievert:
> The actual `fit` calls will take place on CV splits of the training dataset.

Right. I've written the docs to mention that each CV split is gathered to one machine for most estimators. Can you review this?

I know `dask_ml.Incremental` aliases `fit` to call `partial_fit` on each chunk of the Dask array. Would that work with dask-searchcv's searches? I believe dask-searchcv computes each CV split regardless of what `fit` can take, but I'm having a hard time seeing that.

stsievert:
This PR now depends on #221 (but implicitly, which is why the Travis CI tests are failing).

stsievert:
I'm drafting a paper for the SciPy 2019 proceedings. In this, I explicitly mention that `decay_rate=0` by default. I use this to show that `IncrementalSearchCV` is best suited for memory-constrained problems.

stsievert:
@TomAugspurger or @mrocklin could you give this documentation a review? I need another set of eyes.

I've updated it pretty significantly, and updated https://github.com/dask/dask-ml/pull/432#issue-234845927 to reflect the changes. The most significant changes:

* Adding usage example of `HyperbandSearchCV` with rule-of-thumb to determine input parameters
* Adding performance comparison of `HyperbandSearchCV` and `IncrementalSearchCV`
* Makes the documentation clear how different model selection problems can be constrained.
    * I give examples of the being both compute- and memory-constrained
    * I highlight the different cases where different estimators can be applied. Not many estimators can be applied to only a single case.
* Changes the IncrementalSearchCV default decay_rate to decay_rate=0
    * And recommending using `decay_rate=1` in a "compute but not memory constrained"

The CI is failing, and looks related. The failing test (`test_same_models_with_same_random_state`) is passing on my machine.

docs/source/cross_validation.rst
```diff
@@ -4,9 +4,15 @@ Cross Validation
...
-The simplest way to split one or more Dask arrays is with :func:`dask_ml.model_selection.train_test_split`.
+The simplest way to split one or more Dask arrays is with :func:`dask_ml.model_selection.train_test_split`:
+
+.. autosummary::
+   dask_ml.model_selection.train_test_split
+
+Here's an example usage:
 
 .. ipython:: python
+   :okwarning:
```
TomAugspurger, line 15: What's the warning here?

docs/source/cross_validation.rst
```diff
@@ -4,9 +4,15 @@ Cross Validation
 See the `scikit-learn cross validation documentation`_ for a fuller discussion of cross validation.
 This document only describes the extensions made to support Dask arrays.
 
-The simplest way to split one or more Dask arrays is with :func:`dask_ml.model_selection.train_test_split`.
+The simplest way to split one or more Dask arrays is with :func:`dask_ml.model_selection.train_test_split`:
+
+.. autosummary::
+   dask_ml.model_selection.train_test_split
```
TomAugspurger, line 10: Maybe not necessary? We'll have a link to the full page just above.

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,193 @@
...
+
+1. "**memory constrained"**, which happens when the dataset size is too large
+   to fit in memory.  This typically happens when a model needs to be tuned for
+   a larger-than-memory dataset after local development.
+2. "**compute constrained**", which happen when the computation takes too long
+   even with data that can fit in memory.  This typically happens when many
+   hyperparameters need to be tuned, which especially common with neural
+   networks.
+
+Here's an example of what "memory constrained" means:
```
TomAugspurger, line 32: A bit of prose on each of these would be welcome. For example

- Memory contained: "The original DataFrame or array is larger than your single machine's memory"
- Compute constrained: "The dataset fits in memory, but training takes too long."

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,193 @@
...
+   }
+
+These issues are independent and both can happen the same time. Dask-ML has
+tools to address all 4 combinations. Let's look at each case.
+
+Neither compute nor memory constrained
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This case happens when there aren't many hyperparameters to tune and the data
+fits in memory. This is common when the search can run on your laptop.
```
TomAugspurger, line 75: Maybe strike the "on your laptop" and replace it with something like "doesn't take too long to train." Many groups use single large nodes, which works well if you have that option.

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,193 @@
...
+
+This case happens when the data doesn't fit in memory but there aren't many
+hyperparameters to search over. The data doesn't fit in memory, so it makes
+sense to call ``partial_fit`` on each chunk of a Dask Array/Dataframe. These
+estimators do that, possibly with some configuration:
+
+.. autosummary::
+   dask_ml.model_selection.IncrementalSearchCV
+   dask_ml.model_selection.RandomizedSearchCV
+   dask_ml.model_selection.GridSearchCV
```
TomAugspurger, line 114: I'm not sure I would put these here. The caveat about Incremental is a big one. I think these fall more naturally under "Compute, but not memory constrained".

stsievert: Thanks. Let me know if you have another edits on where different estimators should go.

docs/source/cross_validation.rst
stsievert: I wanted to keep the consistency of linking to Dask-ML documentation with `autosummary`.

docs/source/cross_validation.rst
stsievert: With NumPy 1.18.1:

``` 
>>>-------------------------------------------------------------------------
Warning in /Users/scott/Developer/stsievert/dask-ml/docs/source/cross_validation.rst at block ending on line 22
Specify :okwarning: as an option in the ipython:: block to suppress this message
----------------------------------------------------------------------------
/Users/scott/anaconda3/envs/dask-ml-test/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject
  return f(*args, **kwds)
<<<-------------------------------------------------------------------------
```

stsievert:
I'm having some testing difficulties. The CI is randomly failing but all tests pass locally under the `ci/environment-3.7.yaml` environment. I'm having a hard time figuring out why. I think it's related to this PR because all the failing tests are in `tests/model_selection/test_incremental.py`. Do you have any insight or guidance?

The failures are not consistent: `win64`, `linux sklearnDev` and `linux linux` randomly fail; however, I don't recall seeing `linux earliest` fail. The failing tests include `test_transform` and `test_numpy_array`, and I've also seen failures on `test_param_random_determinism` and `test_model_random_determinism`.

It looks like these tests are failing because of a Client timeout error. Here's one relevant error for `test_numpy_array` from the `linux linux` CI machine (but with the traceback first).


<details>

```
=================================== FAILURES ===================================

----------------------------- Captured stderr call -----------------------------
tornado.application - ERROR - Exception after Future was cancelled
  File "/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/home/vsts/work/1/s/tests/model_selection/test_incremental.py", line 416, in test_numpy_array
    yield search.fit(X, y, classes=[0, 1])
  File "/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/home/vsts/work/1/s/dask_ml/model_selection/_incremental.py", line 577, in _fit
    random_state=self.random_state,
  File "/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/tornado/gen.py", line 742, in run
    yielded = self.gen.throw(*exc_info)  # type: ignore
  File "/home/vsts/work/1/s/dask_ml/model_selection/_incremental.py", line 209, in _fit
    metas = yield client.gather(new_scores)
  File "/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
concurrent.futures._base.CancelledError

_______________________________ test_numpy_array _______________________________
    def test_func():
        result = None
        workers = []
        with clean(timeout=active_rpc_timeout, **clean_kwargs) as loop:

            async def coro():
                with dask.config.set(config):
                    s = False
                    for i in range(5):
                        try:
                            s, ws = await start_cluster(
                                nthreads,
                                scheduler,
                                loop,
                                security=security,
                                Worker=Worker,
                                scheduler_kwargs=scheduler_kwargs,
                                worker_kwargs=worker_kwargs,
                            )
                        except Exception as e:
                            logger.error(
                                "Failed to start gen_cluster, retrying",
                                exc_info=True,
                            )
                        else:
                            workers[:] = ws
                            args = [s] + workers
                            break
                    if s is False:
                        raise Exception("Could not start cluster")
                    if client:
                        c = await Client(
                            s.address,
                            loop=loop,
                            security=security,
                            asynchronous=True,
                            **client_kwargs
                        )
                        args = [c] + args
                    try:
                        future = func(*args)
                        if timeout:
                            future = asyncio.wait_for(future, timeout)
                        result = await future
                        if s.validate:
                            s.validate_state()
                    finally:
                        if client and c.status not in ("closing", "closed"):
                            await c._close(fast=s.status == "closed")
                        await end_cluster(s, workers)
                        await asyncio.wait_for(cleanup_global_workers(), 1)

                    try:
                        c = await default_client()
                    except ValueError:
                        pass
                    else:
                        await c._close(fast=True)

                    for i in range(5):
                        if all(c.closed() for c in Comm._instances):
                            break
                        else:
                            await asyncio.sleep(0.05)
                    else:
                        L = [c for c in Comm._instances if not c.closed()]
                        Comm._instances.clear()
                        # raise ValueError("Unclosed Comms", L)
                        print("Unclosed Comms", L)

                    return result

            result = loop.run_sync(
>               coro, timeout=timeout * 2 if timeout else timeout
            )

/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/distributed/utils_test.py:957:
/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/tornado/ioloop.py:532: in run_sync
    return future_cell[0].result()
/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/distributed/utils_test.py:927: in coro
    result = await future

fut = <Future cancelled>, timeout = 10

    async def wait_for(fut, timeout, *, loop=None):
        """Wait for the single Future or coroutine to complete, with timeout.

        Coroutine will be wrapped in Task.

        Returns result of the Future or coroutine.  When a timeout occurs,
        it cancels the task and raises TimeoutError.  To avoid the task
        cancellation, wrap it in shield().

        If the wait is cancelled, the task is also cancelled.

        This function is a coroutine.
        """
        if loop is None:
            loop = events.get_event_loop()

        if timeout is None:
            return await fut

        if timeout <= 0:
            fut = ensure_future(fut, loop=loop)

            if fut.done():
                return fut.result()

            fut.cancel()
            raise futures.TimeoutError()

        waiter = loop.create_future()
        timeout_handle = loop.call_later(timeout, _release_waiter, waiter)
        cb = functools.partial(_release_waiter, waiter)

        fut = ensure_future(fut, loop=loop)
        fut.add_done_callback(cb)

        try:
            # wait until the future completes or the timeout
            try:
                await waiter
            except futures.CancelledError:
                fut.remove_done_callback(cb)
                fut.cancel()
                raise

            if fut.done():
                return fut.result()
            else:
                fut.remove_done_callback(cb)
                # We must ensure that the task is not running
                # after wait_for() returns.
                # See https://bugs.python.org/issue32751
                await _cancel_and_wait(fut, loop=loop)
>               raise futures.TimeoutError()
E               concurrent.futures._base.TimeoutError

/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/asyncio/tasks.py:449: TimeoutError


```

</details>

docs/source/hyper-parameter-search.rst
```diff
@@ -5,15 +5,18 @@ Hyper Parameter Search
 models using Dask, and to scale hyperparameter optimization to either* **larger
 data** *or* **more computational power.**
 
-This page will presume knowledge of hyperparameter tuning/optimization; if not,
-see Scikit-learn's documentation on "`Tuning the hyper-parameters of an
-estimator <https://scikit-learn.org/stable/modules/grid_search.html>`_." The
-issues that arise in these hyperparameter searches will be explicitly mentioned
-in the following section:
+Hyperparameter searches are a require process in machine learning. Briefly,
```
dhirschfeld, line 8:
```suggestion
Hyperparameter searches are a required process in machine learning. Briefly,
```

docs/source/cross_validation.rst
TomAugspurger: OK can you remove this okwarning then? That sounds like an issue with either your local install or our CI script.

Ideally okwarning is only used when we knowingly emit a warning.

docs/source/cross_validation.rst
stsievert: Maybe that's better suited for a separate PR. I'll revert this change.

Here's a page with the "consistency" envisioned: https://ml.dask.org/clustering.html.

Here are other pages with links to estimators without autosummaries:

* `dask_ml.preprocessing.OneHotEncoder` on https://ml.dask.org/preprocessing.html
* `dask_ml.model_selection.train_test_split()` on https://ml.dask.org/cross_validation.html
* `dask_ml.compose.ColumnTransformer` on https://ml.dask.org/compose.html
* `wrappers.ParallelPostFit` on https://ml.dask.org/meta-estimators.html

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
...
-If your data is large and the underlying estimator implements ``partial_fit``, you can
-Dask-ML's :ref:`*incremental* hyperparameter optimizers <hyperparameter.incremental>`.
+* ":ref:`hyperparameter.scaling`" will mention problems often seen in
+  hyperparameter optimization searches, and tools to get around these issues.
+
+Then, some tooling will be mentioned to circumvent the issues mentioned, and
+expanded on in the following sections:
+
+1. ":ref:`hyperparameter.drop-in`" will mention classes that are drop-in
+   replacements for Scikit-learn but work nicely with Dask objects.
```
TomAugspurger, line 25:
```suggestion
   replacements for Scikit-learn but work nicely with Dask objects and can offer better performance.
```

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
...
+   larger-than-memory dataset after local development.
+2. "**compute constrained**". This happen when the computation takes too long
+   even with data that can fit in memory.  This typically happens when many
+   hyperparameters need to be tuned, which especially common with neural
+   networks.
+
+"Memory constrained" searches happen when the data doesn't fit in the memory of
+a single machine:
+
+.. code:: python
```
TomAugspurger, line 50:
```suggestion
.. code-block:: python
```

and all the following ones.

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
...
+
+   # not memory constrained
+   import pandas as pd
```
TomAugspurger, line 53: Format these like you're at the interpreter. Start with `>>>` and put the outputs below.

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
...
+fits in memory. This is common when the search doesn't take too long to run.
+
+Scikit-learn can handle this case:
+
+.. autosummary::
+   sklearn.model_selection.GridSearchCV
+   sklearn.model_selection.RandomizedSearchCV
+
+Dask-ML also has some drop in replacements for the Scikit-learn versions that
+works well with Dask.
```
TomAugspurger, line 105:
```suggestion
works well with Dask collections.
```

and maybe link to https://docs.dask.org/en/latest/user-interfaces.html#high-level-collections

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
...
+   dask_ml.model_selection.GridSearchCV
+   dask_ml.model_selection.RandomizedSearchCV
+
+
+These models work well with Dask Arrays/DataFrames. By default, these
+estimators will pass the entire dataset to ``fit`` even if a Dask
+Array/DataFrame is passed.  More detail is in
+":ref:`works-with-dask-collections`".
+
+These estimators make some progress into hyperparameter searches that are
```
TomAugspurger, line 117: What are "these estimators" and what does "make some progress" mean?

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
...
+
+Compute constrained, but not memory constrained
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This case happens when the data fits on in the memory of one machine but when
+there are a lot of hyperparameters to search. The best class for this case is
+:class:`~dask_ml.model_selection.HyperbandSearchCV`:
+
+.. autosummary::
+   dask_ml.model_selection.HyperbandSearchCV
```
TomAugspurger, line 153: Does including this here imply that it doesn't work with compute *and* memory constrained? Or should I not read it that way?

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
...
+with the data. For more detail, see "`Tuning the hyper-parameters of an
+estimator <https://scikit-learn.org/stable/modules/grid_search.html>`_."
 
-If your dataset is relatively small or the underlying estimator doesn't implement
-``partial_fit``, you can use :class:`dask_ml.model_selection.GridSearchCV` or
-:class:`dask_ml.model_selection.RandomizedSearchCV`.
-These are drop-in replacements for their scikit-learn counterparts, that should offer better performance and handling of Dask Arrays and DataFrames.
-The underlying estimator will need to be able to train on each cross-validation split of the data.
-See :ref:`hyperparameter.drop-in` for more.
+These searches often take weeks or months to complete, typically when searching
```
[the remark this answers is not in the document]
stsievert: I've changed it to "These searches can take an ample time (days or weeks), typically ...."

Should I find references to back this claim up? It's certainly true in personal experience.

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
 Hyper Parameter Search
 ======================
 
-*Tools for performing hyperparameter optimization of Scikit-Learn API-compatible models using Dask*.
+*Tools to perform hyperparameter optimizaiton of Scikit-Learn API-compatible
+models using Dask, and to scale hyperparameter optimization to either* **larger
+data** *or* **more computational power.**
```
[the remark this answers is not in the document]
stsievert: I've changed this to "larger data and/or larger searches". I've used "larger searches" as a stand-in for "more intense search that requires a cluster".

docs/source/hyper-parameter-search.rst
```diff
@@ -1,33 +1,232 @@
...
 
-If your dataset is relatively small or the underlying estimator doesn't implement
-``partial_fit``, you can use :class:`dask_ml.model_selection.GridSearchCV` or
-:class:`dask_ml.model_selection.RandomizedSearchCV`.
-These are drop-in replacements for their scikit-learn counterparts, that should offer better performance and handling of Dask Arrays and DataFrames.
-The underlying estimator will need to be able to train on each cross-validation split of the data.
-See :ref:`hyperparameter.drop-in` for more.
+These searches often take weeks or months to complete, typically when searching
+for great performance or with massive datasets. This is common when preparing
+for production or a paper publication. These issues will be clarified in the
```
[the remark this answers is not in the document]
stsievert: I've tried to remove that (very passive) phrase.

``` 
 The following section clarifies the issues that can occur:
```

I've removed all instances of "will be", and tend to use "detail" or "details".

TomAugspurger:
> The CI is randomly failing

I've seen those hanging tests elsewhere. I haven't been able to diagnose them yet.

docs/source/hyper-parameter-search.rst
stsievert: > it doesn't work with compute and memory constrained?

No. Hyperband the like reduce the amount of computation required, regardless of data size. That is, they fit in "compute constrained" regardless of "memory constrained" status. I've added a note in "compute and memory constrained":

> **Compute and memory constrained**
> * HyperbandSearchCV
> * ...
>
>  These classes work well with data that does not fit in memory. They also reduce the computation required as described in "Compute constrained but not memory constrained"

Does that clear your confusion?

stsievert:
`IncrementalSearchCV` takes an algorithm-specific parameter, `decay_rate` and has an algorithm-specific implementation. This means the children of `IncrementalSearchCV` have to  take a disjoint set of parameters, breaking [Liskov's substitution principle][1]. For example, `SuccessiveHalvingSearchCV` does not accept `decay_rate` even though it's an instance of `IncrementalSearchCV`.

I'm planning on implementing this inheritance structure:

``` python
class BaseIncrementalSearchCV(...):
    ...


class IncrementalSearchCV(BaseIncrementalSearchCV):
    def __init__(self, ...):  # does not take decay_rate

    def _adapt(self, info):
        return {k: self.fits_per_score for k in info}  # same as decay_rate=0


class InverseDecaySearchCV(IncrementalSearchCV):
    def __init__(self, decay_rate=1, ...):

    def _adapt(self, info):
        ...  # current version of IncrementalSearchCV._adapt


class SuccessiveHalvingSearchCV(IncrementalSearchCV):
    def __init__(self, aggressiveness=3, ...):

    def _adapt(self, info):
        ...
```

I'd appreciate knowing if there are any issues with this inheritance structure.

This will have an added documentation benefit: `IncrementalSearchCV` will only live in "memory- but not compute constrained", not every category except "not compute-constrained nor memory-constrained" with recommendations on changing `decay_rate`.

[1]:https://en.wikipedia.org/wiki/Liskov_substitution_principle

(originally mentioned in https://github.com/dask/dask-ml/pull/528#discussion_r409076757)

stsievert:
I've done the following:

* Implemented and documented `InverseDecaySearchCV` as mentioned in https://github.com/dask/dask-ml/pull/432#issuecomment-614253784
* Tested `InverseDecaySearchCV`. Tests are implemented to ensure `_adapt` works as expected.
* Deprecated `IncrementalSearchCV.decay_rate`. This warns by default and prints a message which specifies how to remove the warning and how to keep the same behavior.

Also, I'm questioning including `decay_rate` in `InverseDecaySearchCV`'s parameters. Let me do some more research to see if anything other than`decay_rate == 1` make sense, or recommendations on how to change it.

TomAugspurger:
To the extent possible, I'd recommend keeping documentation changes, additional tests, and behavior changes separate. That makes things much easier for me to review.

stsievert:
I've tried to do that with two PRs:

* This PR is focused around documentation improvements to `docs/source/hyperparameter.rst`. In addition to making these improvements, this PR also...
    * Deprecates `decay_rate`. This makes `IncrementalSearchCV` better fit in "memory but not compute constrained".
    * Implements a new class `InverseDecaySearchCV`. This class is a wrapper for the previous implementation of `IncrementalSearchCV._adapt`.
    * Makes other small documentation improvements (adding to documentation of dask-searchcv, linking in Hyperband docs, etc)
* #651: cleaning `InverseDecaySearchCV`. This PR implements and tests improvements to `InverseDecaySearchCV._adapt`.

Will that be sufficient, or should I remove the `IncrementalSearchCV` deprecation as well? That'll mean this PR is only modifies `*.rst` files/docstrings and #651 only modifies `*.py` files.

dask_ml/model_selection/_incremental.py
```diff
@@ -958,6 +951,25 @@ def __init__(
         )
 
     def fit(self, X, y=None, **fit_params):
+        if "IncrementalSearchCV" in str(type(self)):
```
TomAugspurger, line 954: Mmm let's avoid the `str(type(self))` stuff. A little less direct, but how about you define a `_check_decay_deprecation()` method

```python
def _check_decay_deprecation(self):
     if self.decay_rate is None: ...
```

then in the subclass (`InverseDecaySearchCV`?) define that method to do nothing.

dask_ml/model_selection/_incremental.py
```diff
@@ -958,6 +951,25 @@ def __init__(
...
+            if self.decay_rate is None:
```
TomAugspurger, line 955: Ideally users wouldn't need to use a filter to silence this. I think we can achieve it with something like

```
no_default = object()


class IncrementalSearchCV(...):
    def __init__(..., decay_rate=no_default, ...):
        ...
    def fit(self):
        if decay_rate is no_default: warnings.warn("deprecated... specify 'decay_rate=None' or use InverseDecaySearchCV to silence")
         # then set decay_rate=1.0 for backwards compat
        elif decay_rate is not None: warnings.warn("deprecated... use InverseDecaySearchCV")
```

That might obviate the need for `_check_decay_deprecation` since presumably `IncrementalSearchCV` would have a different default.

Then the three cases are

1. Not specified. Warn about using InverseDecaySearch if you want decay, o/w decay=None, set `decay_rate=1.0`, proceed
2. Specified with `decay_rate=None`: no warning
3. Specified with `decay_rate=number`: warn instructing to use InverseDecaySearchCV.

dask_ml/model_selection/_incremental.py
```diff
@@ -958,6 +951,25 @@ def __init__(
...
+                    "    >>> with warnings.catch_warnings():\n"
+                    "    >>>     warnings.filterwarnings('ignore', "
+                    "category=FutureWarning, module='dask_ml')\n"
+                    "    >>>     search.fit(X, y, **fit_params)\n",
+                    FutureWarning,
+                )
+            else:
+                warn(
+                    "decay_rate is deprecated in InverseDecaySearchCV. "
+                    f"Use InverseDecaySearchCV to use decay_rate={self.decay_rate}",
```
TomAugspurger, line 971: Use a FutureWarning in both places.

dask_ml/model_selection/_incremental.py
stsievert: I avoided this because I didn't want the user to change their code twice; once to set `decay_rate=None`, and once to remove `decay_rate` at some point in the future.

I've implemented your suggestion. Two deprecation cycles is clear.

stsievert:
Here are some arguments for deprecating `decay_rate`:

1. (documentation) `IncrementalSearchCV` is the only class to scale "memory- but not compute-constrained" searches. I think these searches is important and common enough to include an implementation that works by default..<br><br>I would imagine this is the most prominent usage of  Dask-ML's hyperparameter optimization: "memory- but not compute constrained" includes training linear models or decision trees on large datasets.
1. (API) `decay_rate` is a algorithm specific parameter. Other adaptive algorithms won't need `decay_rate`. Case in point: `SuccessiveHalvingSearchCV` doesn't take `decay_rate`. Other adaptive algorithms like (the popular) median elimination won't either. This leads to a violation of Liskov substitution principle.
3. (weaker API) Without deprecating `decay_rate`, `IncrementalSearchCV` will have two useful implementations:
    * An implementation best suited for large data with `decay_rate=0`
    * An implementation of an adaptive algorithm with `decay_rate=1` (at least after #651).

I don't know how  `decay_rate` performs if `decay_rate not in [0, 1]`. I don't even really know how `decay_rate=1` performs, but at least know the motivation and have performance results for a related implementation, `SuccessiveHalvingSearchCV`.

(1) is the motivation for included the deprecation in this PR.

[1]:https://en.wikipedia.org/wiki/Liskov_substitution_principle

tests/model_selection/test_incremental.py
```diff
@@ -25,13 +25,19 @@
...
+)
 from dask_ml.model_selection._incremental import _partial_fit, _score, fit
 from dask_ml.model_selection.utils_test import LinearFunction, _MaybeLinearFunction
 from dask_ml.utils import ConstantFunction
-from dask_ml.wrappers import Incremental
 
-pytestmark = pytest.mark.skipif(not DISTRIBUTED_2_5_0, reason="hangs")
+pytestmark = [
+    pytest.mark.skipif(not DISTRIBUTED_2_5_0, reason="hangs"),
+    pytest.mark.filterwarnings("ignore:decay_rate"),
```
stsievert, line 39: Why is this ignored?

tests/model_selection/test_incremental.py
```diff
@@ -414,10 +414,14 @@ def test_numpy_array(c, s, a, b):
     X, y = make_classification(n_samples=100, n_features=5, chunks=(10, 5))
     X, y = yield c.compute([X, y])
     model = SGDClassifier(tol=1e-3, penalty="elasticnet")
-    params = {"alpha": np.logspace(-2, 10, 10), "l1_ratio": np.linspace(0.01, 1, 20)}
+    params = {
+        "alpha": np.logspace(-5, -3, 10),
```
stsievert, line 418: Why is `alpha` changed from `logspace(-2, 10)` to `logspace(-5, -3)`?

tests/model_selection/test_incremental.py
```diff
@@ -414,10 +414,14 @@ def test_numpy_array(c, s, a, b):
...
+        "l1_ratio": np.linspace(0, 1, 20),
+    }
 
-    search = IncrementalSearchCV(model, params, n_initial_parameters=10)
+    search = IncrementalSearchCV(model, params, n_initial_parameters=10, max_iter=10)
     yield search.fit(X, y, classes=[0, 1])
+    assert search.best_score_ > 0  # smoke test
```
stsievert, line 424: Smoke test to ensure what?

tests/model_selection/test_incremental.py
```diff
@@ -529,48 +530,47 @@ def test_high_performing_models_are_retained_with_patience(c, s, a, b):
 
 @gen_cluster(client=True)
 def test_same_params_with_random_state(c, s, a, b):
-    X, y = make_classification(n_samples=100, n_features=5, chunks=(10, 5))
-    model = SGDClassifier(tol=1e-3, penalty="elasticnet")
+    X, y = make_classification(n_samples=100, n_features=10, chunks=10, random_state=0)
```
stsievert, line 533: Why the change in `n_features`?

dask_ml/model_selection/_search.py
```diff
@@ -1181,6 +1181,14 @@ def fit(self, X, y=None, groups=None, **fit_params):
...
+        -----
+        This class performs best when each cross validation split fits into
+        RAM because ``fit`` is called on each cross validation split. For
+        example, if using :class:`~sklearn.model_selection.KFold` with k
+        chunks, a :math:`1 - 1/k` fraction of the dataset should fit into RAM.
```
stsievert, line 1190: What happens if the sklearn impl is wrapped with `dask_ml.wrappers.Incremental`?

tests/model_selection/test_incremental.py
```diff
@@ -702,12 +702,12 @@ def test_verbosity_types(c, s, a, b):
     params = {"value": scipy.stats.uniform(0, 1)}
 
     for verbose in [-1.0, 1.2]:
-        search = IncrementalSearchCV(model, params, verbose=verbose)
+        search = IncrementalSearchCV(model, params, verbose=verbose, max_iter=3)
```
stsievert, line 705: Why was `max_iter=3` added?

tests/model_selection/test_incremental.py
stsievert: Because the warnings are tested in `test_incremental_warns.py`.

tests/model_selection/test_incremental.py
stsievert: Because `logspace(-2, 10)` creates values from `1e-2` to `1e10`. That's unreasonable, especially because `alpha` is a weight on the regularization term. `alpha=1e10` will result in all of `model.coef_` being 0. Small `alpha`s are more realistic and are provided.

tests/model_selection/test_incremental.py
stsievert: Because this is a test on verbosity and `max_iter=3` will speed up the test.

tests/model_selection/test_incremental.py
stsievert: Comment added. Now, this is the only assert in this test function.

tests/model_selection/test_incremental.py
stsievert: Because this is a test to ensure determinism. Increasing the number of features so each chunk shape is `(10, 10)` will make each `partial_fit` call extremely unstable.

This isn't required, but does make this test a little stronger.

dask_ml/model_selection/_search.py
stsievert: Doesn't matter. The model doesn't influence the size of the passed NumPy array.

Dask-searchcv always passed NumPy arrays of size `len(X) * (1 - 1/k)` with K-fold cross validation. I think the relevant test is https://github.com/dask/dask-ml/blob/b812fe5db8fcf6f37d805adafe1e7c8f21a18df8/tests/model_selection/dask_searchcv/test_model_selection.py#L151-L155

`cv` could be `sklearn.model_selection.KFold`. If I'm interpreting this test correctly, this asserts that the data is divided in 3 chunks for both Dask arrays and NumPy arrays. In both cases, 2/3rds of the data is used for fitting and 1/3rd for scoring.

TomAugspurger:
FYI, going to ignore the sklearn dev failures for now. We'll need to add `n_features_in_` to each estimator, which will take a bit of work.

stsievert:
It looks like all the CI failures are unrelated. There's a couple other warnings besides the one with `n_features_in_`:

`n_features_in_`:

<details>

``` python

# data = dask.array<array, shape=(10, 4), dtype=float64, chunksize=(5, 4), chunktype=numpy.ndarray>

@pytest.mark.parametrize("data", [X, dX, df, ddf])
def test_fit(data):
    a = sklearn.impute.SimpleImputer()
    b = dask_ml.impute.SimpleImputer()
    
    a.fit(X)
    b.fit(data)
    
    assert_estimator_equal(a, b)
```

This raises an `AssertionError: {'n_features_in_'}`. The raised FutureWarning is

> As of scikit-learn 0.23, estimators should expose a n_features_in_ attribute, unless the 'no_validation' tag is True. This attribute should be equal to the number of features passed to the fit method. An error will be raised from version 0.25 when calling check_estimator(). See SLEP010: https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep010/proposal.html

</details>

`PartialPassiveAggressiveClassifier.average_intercept_`:

<details>

``` python
# file: TestPassiveAggressiveClassifier.test_basic

def test_basic(self, single_chunk_classification):
    X, y = single_chunk_classification
    a = lm.PartialPassiveAggressiveClassifier(classes=[0, 1], random_state=0, max_iter=100, tol=1e-3)
    b = lm_.PassiveAggressiveClassifier(random_state=0, max_iter=100, tol=1e-3)
    a.fit(X, y)
    b.partial_fit(*dask.compute(X, y), classes=[0, 1])
    assert_estimator_equal(a, b, exclude=["loss_function_"])
```

This also raises a FutureWarning:

> FutureWarning: Attribute average_intercept_ was deprecated in version 0.23 and will be removed in 0.25.

</details>

Passing keyword arguments:

<details>

``` python
    def test_check_cv():
        # No y, classifier=False
       cv = check_cv(3, classifier=False)
```

> FutureWarning: Pass classifier=False as keyword args. From version 0.25 passing these as positional arguments will result in an error

This also shows up in `TestPolynomialFeatures`:

``` python
# ______________________ TestPolynomialFeatures.test_basic _______________________

# self = <tests.preprocessing.test_data.TestPolynomialFeatures object at 0x7ff8848d8a90>

def test_basic(self):
    a = dpp.PolynomialFeatures()
```


> FutureWarning: Pass interaction_only=False, include_bias=True as keyword args. From version 0.25 passing these as positional arguments will result in an error

</details>

TomAugspurger:
I fixed the `PartialPassiveAggressiveClassifier.average_intercept_` in another branch.

I think this is ready to go. All good on your end?

TomAugspurger:
Going to merge so we can include this in the 1.4.0 release. I can make another release anytime if needed.

stsievert:
> I think this is ready to go. All good on your end?

Yup! I'm glad to see this merged.

Issue #415, opened 2018-10-19
Release finished model/job during cross validation


zexuanz-zefr:
I'm wondering if there is a way to release the job/worker when a fitting job finishes during RandomizedSearchCV.

I'm doing something like
`random_search_cv = RandomizedSearchCV(
    estimator=model, 
    param_distributions=params_grid, 
    n_iter=n,
    scoring=score,
    cv=5,
    scheduler=client,
    n_jobs=-1
)
random_search_cv.fit(x_train, y_train)`
And I wonder if there is a parameter like 'release_job=True' to set so that when each model finished in the cv it will be release instead of waiting for the entire cv to finish and then return.
By releasing finished work during cv we can efficiently prevent working being killed/restarted.
Thanks

vecchp:
This is also something I've been pondering.  After looking into a bit myself there may be two possible places to update the behavior in _search.py:

1) DaskBaseSearchCV
2) build_graph 

Here is the pseudo code that demonstrates what I'm trying to accomplish in DaskBaseSearchCV

``` Python
def fit(self, X, y=None, groups=None, **fit_params):
  ... some additional logic for handling futures vs non-future computation ...
  if schedulers is 'dask.distributed':
    out = scheduler(dsk, keys, num_workers=n_jobs, sync=False)    
    results = []
    with future, result as_completed(out, with_results=True):
       #gather results as they come in
       #cancel the task so it won't recompute
```

The implementation appears quite hacky as you have to cancel futures as they complete.  What would be desirable is that each cv task releases itself once the computation is complete instead of having a supervisor watch as they complete.  Maybe someone more familiar with Dask internals could provide some guidance to a cleaner solution?

mrocklin:
These are, I think, two different cases.

In the original case with `RandomSearchCV` I suspect that @zexuanz-zefr is using scikit-learn's `RandomSearchCV` and is using Dask with the joblib extension.  I fully agree with @zexuanz-zefr 's assessment.  Probably this needs to be pushed upstream to Scikit-Learn, or better yet, the [joblib issue tracker](https://github.com/joblib/joblib/issues/new).  Also cc'ing @ogrisel who seems to manage most Joblib work today.

TomAugspurger:
Since @zexuanz-zefr used `scheduler=client`, I suspect they're using `dask_ml.model_selection.RandomizedSearchCV`.

vecchp:
[...]

Issue #230, opened 2018-06-27
Error of random seed when using train_test_split()


PeterFogh:
I get this error when executing the following code:

```
import dask.array as da
from dask_ml.datasets import make_regression
from dask_ml.model_selection import train_test_split

print('Local dask.__version__: {}'.format(dask.__version__))
print('Local dask_ml.__version__: {}'.format(dask_ml.__version__))
print('Client dask.__version__: {}'.format(dask.delayed(dask.__version__).compute()))
print('Client dask_ml.__version__: {}'.format(dask.delayed(dask_ml.__version__).compute()))
```
Local dask.__version__: 0.16.1
Local dask_ml.__version__: 0.6.0
Client dask.__version__: 0.16.1
Client dask_ml.__version__: 0.6.0
```
X, y = make_regression(n_samples=10000, n_features=4, random_state=0, chunks=4)
X
```
dask.array<da.random.normal, shape=(10000, 4), dtype=float64, chunksize=(4, 4)>

```
X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)
```
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-51-bd652d57a653> in <module>()
----> 1 X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)

~\AppData\Local\Continuum\anaconda3\envs\py36_all\lib\site-packages\dask_ml\model_selection\_split.py in train_test_split(*arrays, **options)
    276                             train_size=train_size, blockwise=blockwise,
    277                             random_state=random_state)
--> 278     train_idx, test_idx = next(splitter.split(*arrays))
    279 
    280     train_test_pairs = ((_blockwise_slice(arr, train_idx),

~\AppData\Local\Continuum\anaconda3\envs\py36_all\lib\site-packages\dask_ml\model_selection\_split.py in split(self, X, y, groups)
    137         for i in range(self.n_splits):
    138             if self.blockwise:
--> 139                 yield self._split_blockwise(X)
    140             else:
    141                 yield self._split(X)

~\AppData\Local\Continuum\anaconda3\envs\py36_all\lib\site-packages\dask_ml\model_selection\_split.py in _split_blockwise(self, X)
    144         chunks = X.chunks[0]
    145         rng = check_random_state(self.random_state)
--> 146         seeds = rng.randint(0, 2**32 - 1, size=len(chunks))
    147 
[...]

PR #433, opened 2018-12-01
Gather CV Results as Completed

vecchp:
In progress PR to address #415.


vecchp:
@TomAugspurger I'll think a little about the tests, though your current suggestion seem reasonable.

dask_ml/model_selection/_search.py
```diff
@@ -1194,6 +1200,25 @@ def fit(self, X, y=None, groups=None, **fit_params):
         if scheduler is dask.threaded.get and n_jobs == 1:
             scheduler = dask.local.get_sync
 
+        if dask.distributed and isinstance(getattr(scheduler, '__self__', None), dask.distributed.Client):
+            cv_results_key = next(k for k in dsk.keys() if 'cv-results' in k)
+            score_keys = dsk[cv_results_key][1]
```
jcrist: I'd change how the graph is generated for the distributed scheduler, instead of mutating the output of `build_graph` here. Also, as written this adds extra tasks on the first call to the scheduler (cv_results, and the refit (if needed)), but then ignores them until the second call (which is inefficient and harder to reason about). It'd be better to just separate these into two calls in all cases.

I'd handle this by decoupling the refitting from the initial graph, and move it to be 2 computes (one for the initial fit, a second for the refit, handling determining the best result locally). The `cv_results` key could then be removed, instead passing a single large list of keys (one for each estimator) to the scheduler. Local logic would then look like:

```python
# One key for each estimator/param combo
dsk, keys, n_splits = build_graph(...)

if is_distributed:
    futures = scheduler(dsk, keys, num_workers=n_jobs, sync=False)
    scores_map = {f.key: res for f, res in as_completed(futures, with_results=True)}
    scores = [scores_map[k] for k in keys]
else:
    scores = scheduler(dsk, keys, num_workers=n_jobs)

cv_results = determine_cv_results_from_scores(scores, ...)

if self.refit:
    # Build the refit graph (IIRC it's just a single task) and call compute again to do the refit.
```

Does that make sense?

dask_ml/model_selection/_search.py
```diff
@@ -1194,6 +1200,25 @@ def fit(self, X, y=None, groups=None, **fit_params):
...
+            futures = scheduler(dsk, score_keys, num_workers=n_jobs, sync=False)
+
+            score_map = {}
+            for future, result in dask.distributed.as_completed(futures, with_results=True):
+                if future.status == 'finished':
+                    score_map[future.key] = result
+                    future.cancel()
```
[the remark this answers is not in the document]
vecchp: I can double check, however from what understanding the future may sit in memory until the entire graph completes computing.

dask_ml/model_selection/_search.py
```diff
@@ -1194,6 +1200,25 @@ def fit(self, X, y=None, groups=None, **fit_params):
         if scheduler is dask.threaded.get and n_jobs == 1:
             scheduler = dask.local.get_sync
 
+        if dask.distributed and isinstance(getattr(scheduler, '__self__', None), dask.distributed.Client):
```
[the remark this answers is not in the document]
vecchp: works for me

dask_ml/model_selection/_search.py
vecchp: Yah it does, i was thinking about taking this approach however I wanted to get the POC working before i made any major changes to the generation of the graph.

dask_ml/model_selection/_search.py
```diff
@@ -11,6 +11,12 @@
 import packaging.version
 from dask.base import tokenize
 from dask.delayed import delayed
+
+try:
+    import dask.distributed
+except ImportError:
+    dask.distributed = None
```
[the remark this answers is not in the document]
vecchp: Sure, can make this change.

vecchp:
@jcrist @TomAugspurger made some changes w.r.t the PR comments.  All tests pass except for the test_cache_cv.  I believe it fails because the changes cause the cache to be hit more times than the previous expected values.   What are your thoughts on these changes now?

I also haven't got around to confirming if the futures no longer need to  be canceled our writing robust tests for the new as_completed feature.

vecchp:
@TomAugspurger do you think you can get a chance to look this over?  Finally got around to the tests and I think the code is in a place to get some more feedback.  

Side note, I'm noticing some CI failures relating to 64bit vs 32bit which i think are unrelated to my PR.  I think it's related to https://github.com/dask/dask-ml/issues/230

TomAugspurger:
I'm working to get the CI passing in https://github.com/dask/dask-ml/pull/439.

At least one issue is from a bug upstream in dask, though it doesn't look too difficult to fix.

TomAugspurger:
All the CI issues should be resolved. Will try to take a look today or tomorrow.

dask_ml/model_selection/_search.py
```diff
@@ -1193,21 +1177,55 @@ def fit(self, X, y=None, groups=None, **fit_params):
         if scheduler is dask.threaded.get and n_jobs == 1:
             scheduler = dask.local.get_sync
 
-        out = scheduler(dsk, keys, num_workers=n_jobs)
+        if "Client" in type(getattr(scheduler, "__self__", None)).__name__:
```
TomAugspurger: Will this approach be heavier on communication? Or is it the same amount, just spread out over time?

My real question is should we provide an option to disable this behavior?

dask_ml/model_selection/_search.py
```diff
@@ -1193,21 +1177,55 @@ def fit(self, X, y=None, groups=None, **fit_params):
...
+            futures = scheduler(
+                dsk, keys, allow_other_workers=True, num_workers=n_jobs, sync=False
+            )
+            result_map = {}
+            while len(result_map) != len(keys):
+                for f in as_completed(futures):
+                    if f.status == "finished":
+                        result_map[f.key] = f.result()
+                    elif f.status == "error":
+                        f.retry()
```
TomAugspurger: We should log that we're retrying this key.

TomAugspurger: This will be a blocking, synchronous call. Probably not what we want.

Could you instead build up a list of failed keys and call `as_completed` on those? Retrying is important for your original use-case, right?

dask_ml/model_selection/utils_test.py
```diff
@@ -1,7 +1,19 @@
+from ast import literal_eval
+
 import numpy as np
+import pytest
+
 from sklearn.base import BaseEstimator, ClassifierMixin
 from sklearn.utils.validation import _num_samples, check_array
 
+try:
```
TomAugspurger: You can safely assume distributed is present in the test environment.

dask_ml/model_selection/utils_test.py
```diff
@@ -188,3 +200,36 @@ def score(self, X=None, Y=None):
...
+        self.killed_workers = killed_workers
+        self.lock = lock
+        self.min_complete = min_complete
+        self.counter = counter
+
+    def fit(self, X, y):
+        w = get_worker()
+        for e in w.executing:
+            t = literal_eval(e)
+            self.lock.acquire()
```
TomAugspurger: Use a context manager.

dask_ml/model_selection/_search.py
vecchp: From my understanding we do want to do a synchronous call.  Looking through the code, Dask futures keep the results on the worker until .result() is called (or we use with_results=True in as_completed), if a worker dies before we gather the result we will recompute won't we?

vecchp:
@TomAugspurger , I'm having trouble figuring out why the Linux python 3.6.6 tests are failing.  Tests seem to be passing on 2.7, Windows 3.6.6 (my machine at home), OSX Python 3.6.6 (At work). And the dependencies seem to be the same across both OSX and Linux.  

> tornado.application - ERROR - Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOLoop object at 0x7fe27224bcc0>>, <Future finished exception=TypeError("No dispatch for <class 'dask_ml.model_selection.utils_test.AsCompletedEstimator'>",)>)

dask_ml/model_selection/_search.py
vecchp: From what I can see I don't see why this would add additional overhead.  This "should" submit effectively the same task graph in the traditional dask scheduler as well as distributed.  The difference here is that the distributed scheduler allows you to submit the graph asynchronously and return back a bunch of futures with you can then yield the results as they return.  Given this I don't see why it would be preferable to disable this feature if a distributed client is present.

vecchp:
@TomAugspurger I believe I have resolved all the issues with the tests and linting.  This should be ready for merging if you're happy with the current implementation.

vecchp:
@TomAugspurger checking in again to see what else should be done.  All tests pass.

TomAugspurger:
I won't have time to look at this for at least the next couple weeks.

TomAugspurger:
> All tests pass.

It seems like coverage is decreasing. Does that make sense, or is this a false positive? https://codecov.io/gh/dask/dask-ml/compare/cfa57e5789fc29bdcee5d846b5116c07f5fbe292...9da5b8f0cfa5960a7ee70f216a684bff96a450b2/changes

vecchp:
> > All tests pass.
> 
> It seems like coverage is decreasing. Does that make sense, or is this a false positive? https://codecov.io/gh/dask/dask-ml/compare/cfa57e5789fc29bdcee5d846b5116c07f5fbe292...9da5b8f0cfa5960a7ee70f216a684bff96a450b2/changes

Makes sense, looks like it's dead code given the as_completed implementation no longer uses it.  I'll update.

vecchp:
@TomAugspurger looks like all stages pass. I think codecov was removed but that should be back to normal as well given the dead code is now gone.

dask_ml/model_selection/_search.py
```diff
@@ -42,28 +43,31 @@
...
 
+logger = logging.getLogger(__name__)
+
 try:
     from cytoolz import get, pluck
 except ImportError:  # pragma: no cover
     from toolz import get, pluck
 
+try:
+    from dask.distributed import as_completed
```
stsievert: Is this try/except needed? `distributed` is a dependency now (#466).

vecchp: I'm fine with removing the optional check.  Previous guidance was to make it option as it wasn't a dependency at the time.

vecchp:
@stsievert all optional distributed checks in this PR should be removed now. 

@TomAugspurger any guidance left to get this merged?

dask_ml/model_selection/_search.py
```diff
@@ -102,7 +103,20 @@ def __call__(self, est):
...
+        fit_params = {
+            k: (k, v) for (k, v) in zip(fit_params, to_keys(dsk, *param_values))
+        }
+    else:
+        fit_params = {}
+
+    return fit_params
+
+
+def build_cv_graph(
```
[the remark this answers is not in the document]
TomAugspurger: I think we can just add back `build_graph` with a comment that it's going to be removed (not even a deprecation warning). This is a private module, but I was lazy when adding dask support to TPOT. Once this is in, I can cut a release and make a PR to tpot updating their call.

TomAugspurger:
I'm looking into the compatibility for tpot now. Will push an update.

vecchp:
Sounds good!

TomAugspurger:
Thanks @vecchp!

vecchp:
@TomAugspurger thanks so much for getting this in!

PR #450, opened 2019-01-31
Fix sklearn-dev build

jrbourbeau:
The `r2_score` metric in scikit-learn was recently updated (see https://github.com/scikit-learn/scikit-learn/pull/12855) to raise a warning when passed single sample. This is causing our `sklearn-dev` build to fail. 

This PR updates `test_search_cv_results_none_param` in `tests/model_selection/dask_searchcv/test_model_selection_sklearn.py` to not pass a single sample to `r2_score`.

### DIFF
--- a/tests/preprocessing/test_encoders.py
+++ b/tests/preprocessing/test_encoders.py
@@ -88,6 +88,8 @@ def test_basic_dataframe(sparse, method, dask_data, dtype):
 
     assert isinstance(result, type(dask_data))
     assert len(result.columns) == expected.shape[1]
+    if sparse:
+        dtype = pd.SparseDtype(dtype, dtype(0))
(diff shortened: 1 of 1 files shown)


jrbourbeau:
Looks like the test failures here are related to the new pandas 0.24 release. I suspect it has to do with `SparseArray`s now implementing the `ExtensionArray` interface. Any thoughts @TomAugspurger?

TomAugspurger:
That look right @jrbourbeau.

I think we may need to update the tests to something like

```python
if sparse:
    dtype = SparseDtype(dtype, dtype(0))
```

just before the dtype assertion.

jrbourbeau:
Thanks @TomAugspurger!

Issue #454, opened 2019-02-12
module 'dask' has no attribute 'sharedict'


akter-pi:
Hi I am trying to use incremental from dask. My code is simple and following the existing example. My code is below. But I got error when it calls inc.fit and the error is "AttributeError: module 'dask' has no attribute 'sharedict'". I am using dask 1.1.1 and dask_ml 0.11.0

``

from dask.distributed import Client
client = Client()

import dask

import dask.dataframe as dd
import dask.array as da

df = dd.read_csv('BreastCancer.csv')
X = df.drop(['Id','Class','Bare.nuclei'],axis=1)
y = df['Class']=='benign'

from dask_ml.model_selection import train_test_split
X_train, X_test, y_train, y_test = train_test_split(X.to_dask_array(True), y.to_dask_array(True))

X_train, X_test, y_train, y_test = dask.persist(X_train, X_test, y_train, y_test)
classes = da.unique(y_train).compute()

from sklearn.linear_model import SGDClassifier
est = SGDClassifier(loss='log', penalty='l2', tol=1e-3)

from dask_ml.wrappers import Incremental
inc = Incremental(est, scoring='accuracy')

inc.fit(X_train, y_train, classes=classes)
``

The error is
`AttributeError                            Traceback (most recent call last)                 
<ipython-input-1-19bb676669f5> in <module>                                                  
     25 inc = Incremental(est, scoring='accuracy')                                          
     26                                                                                     
---> 27 inc.fit(X_train, y_train, classes=classes)                                          
                                                                                            
/mnt/d/PI.X/py3.5_venv/lib/python3.5/site-packages/dask_ml/wrappers.py in fit(self, X, y, **
fit_kwargs)                                                                                 
    462     def fit(self, X, y=None, **fit_kwargs):                                         
    463         estimator = sklearn.base.clone(self.estimator)                              
--> 464         self._fit_for_estimator(estimator, X, y, **fit_kwargs)                      
    465         return self                                                                 
    466                                                                                     
                                                                                            
/mnt/d/PI.X/py3.5_venv/lib/python3.5/site-packages/dask_ml/wrappers.py in _fit_f(self, estim
ator, X, y, **fit_kwargs)                                                                   
    453                 random_state=self.random_state,                                     
    454                 shuffle_blocks=self.shuffle_blocks,                                 
--> 455                 **fit_kwargs                                                        
    456             )                                                                       
    457                                                                                     
                                                                                            
[...]

Issue #220, opened 2018-06-21
Verify Incremental works with dask.dataframe


TomAugspurger:

TomAugspurger:
It does not

```pytb
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-23-505152de83da> in <module>()
----> 1 inc.fit(X2, y2)

~/sandbox/dask-ml/dask_ml/wrappers.py in fit(self, X, y, **fit_kwargs)
    287     def fit(self, X, y=None, **fit_kwargs):
    288         check_scoring(self.estimator, self.scoring)
--> 289         result = fit(self.estimator, X, y, **fit_kwargs)
    290 
    291         # Copy the learned attributes over to self

~/sandbox/dask-ml/dask_ml/_partial.py in fit(model, x, y, compute, **kwargs)
    167     if y is not None:
    168         assert y.ndim == 1
--> 169         assert x.chunks[0] == y.chunks[0]
    170     assert hasattr(model, 'partial_fit')
    171     if len(x.chunks[1]) > 1:

~/sandbox/dask/dask/dataframe/core.py in __getattr__(self, key)
   2436             return new_dd_object(merge(self.dask, dsk), name,
   2437                                  meta, self.divisions)
-> 2438         raise AttributeError("'DataFrame' object has no attribute %r" % key)
   2439 
   2440     def __dir__(self):

AttributeError: 'DataFrame' object has no attribute 'chunks'
```

We have two options:

1. Just get the array with `.values` in `Incremental.fit`
2. Update `_partial.fit`'s check to handle dataframes or arrays

We probably want option 2, since then we can verify that the divisions actually match.

mrocklin:
If we're adding in an if-else tree then I'll also request a special case be
given to List[Delayed] or List[Future]

manugarri:
Is there any update on this? I am trying to find a workaround to using dask.dataframes but Im not sure how I can  handle a pipeline that has columns and make it work with a dask.array.

As of versions:

```
dask_ml=0.11.0
dask=0.18.2
```

Doing something like this:

```
pipeline = make_pipeline(
    Categorizer(columns=["VendorID", "store_and_fwd_flag", "payment_type", "RatecodeID", "PULocationID"]),
    DummyEncoder(columns=["VendorID", "store_and_fwd_flag", "payment_type", "RatecodeID", "PULocationID"]),
    StandardScaler(),
[...]

PR #455, opened 2019-02-13
Fix incremental bug with Dask sharedict

jrbourbeau:
Currently we're passing a dictionary `values` method to `sharedict.merge` instead of the actual values themselves. Fixes #454.

PR #458, opened 2019-02-13
TST: make sure Incremental works with dataframes

stsievert:
This PR makes sure that dataframes work with Incremental. For tests, I modify the basic incremental test and pass both Dask arrays and dataframes.

**Reference issues/PRs**
* #220: "Verify Incremental works with dask.dataframe"
* #454: I found this issue while writing https://github.com/dask/dask-ml/issues/454#issuecomment-463258586
* #456, which is on setting minimum dependency versions for CI jobs

### DIFF
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -10,6 +10,7 @@ import numpy as np
 import six
 import sklearn.utils
 from dask.delayed import Delayed
+import dask.dataframe as dd
 from toolz import partial
 
 from ._utils import copy_learned_attributes
@@ -167,8 +168,17 @@ def fit(model, x, y, compute=True, shuffle_blocks=True, random_state=None, **kwa
     if y is not None:
         if not hasattr(y, "chunks") and hasattr(y, "to_dask_array"):
             y = y.to_dask_array()
+
+    # dask 1.0.0 has a dd.DataFrame.to_dask_array, but not earlier versions
+    if isinstance(x, dd.DataFrame):
+        x = x.values
+    if isinstance(y, (dd.DataFrame, dd.Series)):
+        y = y.values
+
+    if y is not None:
         assert y.ndim == 1
         assert x.chunks[0] == y.chunks[0]
+
     assert hasattr(model, "partial_fit")
     if len(x.chunks[1]) > 1:
         x = x.rechunk(chunks=(x.chunks[0], sum(x.chunks[1])))
--- a/dask_ml/_partial.py
+++ b/dask_ml/_partial.py
@@ -6,7 +6,6 @@ import warnings
 from abc import ABCMeta
 
 import dask
-import dask.dataframe as dd
 import numpy as np
 import six
 import sklearn.utils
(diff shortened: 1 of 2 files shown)


dask_ml/_partial.py
```diff
@@ -167,8 +168,17 @@ def fit(model, x, y, compute=True, shuffle_blocks=True, random_state=None, **kwa
     if y is not None:
         if not hasattr(y, "chunks") and hasattr(y, "to_dask_array"):
             y = y.to_dask_array()
+
+    # dask 1.0.0 has a dd.DataFrame.to_dask_array, but not earlier versions
```
TomAugspurger: I think we're calling 1.0.0 our "minimum dask version" for a while. So feel free to use it.

So IIUC, this shouldn't be necessary, since the `if not hasattr(y, "chunks") and hasattr(y, "to_dask_array")` will handle things.

tests/test_incremental.py
```diff
@@ -29,23 +30,34 @@ def test_set_params():
...
         result = clf.fit(X, y, classes=[0, 1])
-        for slice_ in da.core.slices_from_chunks(X.chunks):
-            est2.partial_fit(X[slice_], y[slice_[0]], classes=[0, 1])
+        for slice_ in da.core.slices_from_chunks(
+            X.chunks if not dataframes else X.values.chunks
+        ):
+            _X, _y = (X[slice_], y[slice_[0]]) if not dataframes else (X, y)
+            est2.partial_fit(_X, _y, classes=[0, 1])
+            if dataframes:
+                with pytest.raises(ValueError, match="cannot convert float NaN to int"):
```
TomAugspurger: What is the cause of the value error? I don't think we want to enshrine that behavior in a test.

stsievert: > enshrine that behavior in a test.

Good point. Resolved.

dask_ml/_partial.py
stsievert: Thanks, that makes life a little easier.

stsievert:
This is no longer a work in progress, and ready for review (and maybe merge).

I had to rebase off #455 to make the tests work.

tests/test_incremental.py
```diff
@@ -60,9 +68,15 @@ def test_incremental_basic(scheduler):
...
-        assert rel_error < 0.2
+        if not dataframes:
+            rel_error = np.linalg.norm(result - expected)
+            rel_error /= np.linalg.norm(expected)
+            assert rel_error < 0.2
+        else:
+            with pytest.raises(
+                ValueError, match="operands could not be broadcast together"
+            ):
+                np.linalg.norm(result - expected)
```
TomAugspurger: What's the type of `result` here? A dask array with unknown chunks or a dask dataframe?

if it's a dataframe, I'd prefer to convert it to an array and make the same assertions, so that we aren't vulnerable to changes in pandas' broadcasting rules.

stsievert: Thanks for that. That cleans up the tests a lot.

TomAugspurger:
Maybe one more merge of master? I think CI should be passing again.

stsievert:
The CI tests for this PR are passing.

TomAugspurger:
Thanks!

PR #461, opened 2019-02-15
Add CI job for oldest supported dependencies

### LINKED ISSUE #456: Add CI job for oldest supported dependencies

jrbourbeau:
This PR adds a CircleCI job to run our tests against the oldest supported version of our dependencies. Closes #456.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,7 +124,7 @@ jobs:
           command: |
             source activate ${ENV_NAME}
             pip uninstall -y dask distributed
-            conda uninstall -y --force pandas scikit-learn multipledispatch numpy numpy-base
+            conda uninstall -y --force pandas scikit-learn multipledispatch
             conda install -y dask=1.0.0 pandas=0.23.4 scikit-learn=0.20 multipledispatch=0.4.9
             conda list -n ${ENV_NAME}
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,6 +124,7 @@ jobs:
           command: |
             source activate ${ENV_NAME}
             conda install -y dask=0.18.2 pandas=0.21.0 scikit-learn=0.20 multipledispatch=0.4.9
+            conda list -n ${ENV_NAME}
       - run:
           command: |
             source activate ${ENV_NAME}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,7 +125,8 @@ jobs:
             source activate ${ENV_NAME}
             pip uninstall -y dask distributed numpy
             conda uninstall -y --force pandas scikit-learn multipledispatch
-            conda install -y dask=1.0.0 pandas=0.23.4 scikit-learn=0.20.0 multipledispatch=0.4.9
+            # pinning numpy version to avoid FutureWarnings coming from scikit-learn
+            conda install -y dask=1.0.0 pandas=0.23.4 scikit-learn=0.20.0 multipledispatch=0.4.9 numpy=1.15.4
             conda list -n ${ENV_NAME}
       - run:
           command: |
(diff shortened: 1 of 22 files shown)


stsievert:
Should the minimum required version of dask by 1.0.0 instead of 0.18.2? https://github.com/dask/dask-ml/pull/455#issuecomment-463297828 and https://github.com/dask/dask-ml/pull/458#discussion_r256968584

jrbourbeau:
To start, I just wanted to match the minimum versions specified in `setup.py` (i.e. dask 0.18.2). But bumping the minimum supported version of dask to 1.0.0 would certainly be okay with me.

jrbourbeau:
Several of the test failures here (ref https://circleci.com/gh/dask/dask-ml/3484?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) are related to newer features in our dependencies (e.g. `dd.DataFrame.to_dask_array` was included in the dask 0.19.0 release, the `dtype` argument for `pandas.get_dummies` was included in the pandas 0.23.0 release, etc.)

What minimum version numbers for our dependencies do we want to support? As @stsievert [pointed out](https://github.com/dask/dask-ml/pull/461#issuecomment-464276184), there has been mention of dask 1.0.0 as a lower bound for dask. @TomAugspurger any thoughts on a minimum version for pandas? `setup.py` currently has pandas 0.21.0 as the minimum (which was released in October 2017).

TomAugspurger:
Dask 1.0 sounds good.

I'm fine with a relatively new version of pandas. 0.23.4 probably makes
sense.

.circleci/config.yml
```diff
@@ -105,10 +105,46 @@ jobs:
...
+          key: deps-{{ .Branch }}-oldest_supported-{{ checksum "ci/environment-3.6.yml" }}
+      - run:
+          name: configure conda
+          command: ./ci/install-circle.sh
+      - run:
+          name: install oldest dependencies
+          command: |
+            source activate ${ENV_NAME}
+            pip uninstall -y dask distributed
+            conda uninstall -y --force pandas scikit-learn multipledispatch numpy numpy-base
```
jrbourbeau: I had to uninstall `numpy` and `numpy-base` here even though it's unrelated to the oldest supported versions of dask, pandas, etc. Based on some local testing, my best guess is that this is because we're installing packages from both the `defaults` and `conda-forge` channels in `ci/environment-3.6.yml`. Any objections to using just the `conda-forge` channel?

TomAugspurger: No objection.

TomAugspurger:
Hmm apparently the flake8 pulled from conda-forge is checking things in type comments? I'd recommend just removing those for now.

TomAugspurger:
Unrelated failures should be fixed, if you could merge master and push again.

jrbourbeau:
I think this should be ready for review now (cc @TomAugspurger). Summary of changes made in this PR:

- Add CircleCI build to test against earliest supported versions of our dependencies
- Update minimum versions in `setup.py`
- Update a few tests that were failing with older dependency versions
- Unpin NumPy version in `ci/environment-2.7.yml` (I suspect this could have been done in #363)
- Remove a couple of Python 2 style type annotations that newer versions of `flake8` didn't like

dask_ml/cluster/k_means.py
```diff
@@ -23,6 +25,9 @@
 
 logger = logging.getLogger(__name__)
 
+if DASK_VERSION < packaging.version.parse("1.1.0"):
```
TomAugspurger: Small nitpick. Can you move this to _compat.py and just import blockwise from there? That

1.  Doesn’t change the user’s dask.array, if they’re using it directly
2. Makes future cleanup easier since it’s all in one place.

TomAugspurger: So _compat will have a top-level blockwise method, and we use that everywhere.

jrbourbeau: Yeah, can do. Thanks for suggesting this change

TomAugspurger:
Thanks.

PR #466, opened 2019-02-19
MAINT: add distributed as a dependency

### LINKED ISSUE #464: Error importing "IncrementalSearchCV"
```
>>> from dask_ml.model_selection import IncrementalSearchCV
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'IncrementalSearchCV'
```
When I am importing IncrementalSearchCV this is the error, and there is no traceback to track the problem!

stsievert:
Fixes #464


stsievert:
This is a work in progress while I figure out the testing.

TomAugspurger:
These are generally hard to test in CI, due to module caching.

I'd recommend a simple change at the top of the file with a

```
try:
    import distributed
except ImportError:
    raise ImportError(nice message here)
```

and just a manual test locally in an environment without distributed.

jrbourbeau:
Do we want to add `distributed` as a dependency? Ref discussion going on in #463

TomAugspurger:
I have a slight preference for keeping this an optional dependency for now.

jrbourbeau:
Sounds good

TomAugspurger:
Ahh sorry @stsievert I may have led you astray... I see now why you started with your dummy class, since we need to provide something to `model_selection/__init__.py` to import, so that it would raise when the class is instantiated.

So, given the headaches of working around that, lets just add a dependency on distributed. Did you want to work on that change here?

stsievert:
> So, given the headaches of working around that, lets just add a dependency on distributed. Did you want to work on that change here?

This solution is a lot easier.

I made that change here, which is a one-line change to `setup.py`. Is that all that's required?

> Ahh sorry @stsievert I may have led you astray... I see now why you started with your dummy class

No worries – I didn't waste too much time on this. Plus, dummy classes give a bunch of overhead.

TomAugspurger:
Yeah, the change to setup.py should be all we need now.

I'll take a look at the failing CI in a bit.

stsievert:
Should a minimum version of distributed be required?

TomAugspurger:
I was just looking to see what a good version would be. It seems like distributed 1.25.0 coincides with Dask 1.0, so let's go with that.

stsievert:
Sounds good. I've added `distributed>=1.25.0`. I'll leave leave #461 to make Dask 1.0 required

TomAugspurger:
CI should be fixed now, if you could merge master and push.

stsievert:
Master merged and CI passing.

TomAugspurger:
Thanks.

PR #471, opened 2019-02-24
update indexable() to just yield dask dataframes (issue #324)

### LINKED ISSUE #324: Update model_selection.utils.make_indexable
In https://github.com/dask/dask-ml/blob/3834756e08a7dc63c3419e9445852e1dd6a48f14/dask_ml/model_selection/utils.py#L43-L46 we do delay `sklearn.utils.validation.indexable` before passing a dask dataframe to it.

Now that DataFrame.iloc is implemented (https://github.com/dask/dask/pull/3805), we should be able to simply yield `x` just like the array case for dask dataframes.

rmccorm4:
closes #324

### DIFF
--- a/dask_ml/model_selection/utils.py
+++ b/dask_ml/model_selection/utils.py
@@ -4,6 +4,7 @@ from distutils.version import LooseVersion
 
 import dask
 import dask.array as da
+import dask.dataframe as dd
 import scipy.sparse as sp
 from dask.base import tokenize
 from dask.delayed import Delayed, delayed
@@ -45,7 +46,7 @@ def to_indexable(*args, **kwargs):
     else:
         indexable = _indexable
     for x in args:
-        if x is None or isinstance(x, da.Array):
+        if x is None or isinstance(x, da.Array) or isinstance(x, dd.DataFrame):
             yield x
         elif is_dask_collection(x):
             yield delayed(indexable, pure=True)(x)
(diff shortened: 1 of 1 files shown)


n8henrie:
For multiple `isinstance` checks you can use a tuple, and it seems like this style is preferred in this codebase: https://github.com/dask/dask-ml/search?q=isinstance&unscoped_q=isinstance

Note that if one type is much more common than the other, using a separate `or` (like you've done) seems to short-circuit and lead to slight performance improvement:

```python
$ python3 --version
Python 3.7.2
$ python3 -m timeit 'isinstance(3, int) or isinstance(3, float)'
5000000 loops, best of 5: 94.1 nsec per loop
$ python3 -m timeit 'isinstance(3, float) or isinstance(3, int)'
1000000 loops, best of 5: 222 nsec per loop
$ python3 -m timeit 'isinstance(3, (int, float))'
2000000 loops, best of 5: 154 nsec per loop
$ python3 -m timeit 'isinstance(3, (float, int))'
2000000 loops, best of 5: 194 nsec per loop
$ python3 -m timeit 'True or isinstance(3, int) or isinstance(3, float)'
20000000 loops, best of 5: 16.8 nsec per loop
```

rmccorm4:
Good observation, updated to use tuple to follow preferred style.

TomAugspurger:
Thanks for this, sorry for the delay in review.

Could you fetch and merge master? The CI should be fixed.

Could you also edit the original post to include "closes #324" so that the issue gets closed?

rmccorm4:
Merged, and no problem!

Looks like there's still some version issue or something with `tornado` in Travis though @TomAugspurger , unless something else broke.

TomAugspurger:
Yep, I'm looking into that failure (that's for the docs).

TomAugspurger:
Thanks!

PR #474, opened 2019-02-28
Fix sklearn dev tests

jrbourbeau:
There are a couple of recent updates to scikit-learn `master` that are causing the `sklearn_dev` CI build to fail (ref https://circleci.com/gh/dask/dask-ml/3634?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). The recent changes are:

- `drop=` keyword was added to `OneHotEncoder` in https://github.com/scikit-learn/scikit-learn/pull/12908

- Internals of `QuantileTransformer` where updated in https://github.com/scikit-learn/scikit-learn/pull/12827

To fix these failures, this PR:

- Adds a `drop=` keyword to the `dask_ml.preprocessing.OneHotEncoder`. By default `drop=None`, which maintains the current behavior. For now, a `NotImplementedError` error is raised when `drop` is not `None`. I also added a test to ensure `NotImplementedError` is raised appropriately. 

- I ported over the changes made to `QuantileTransformer`

### DIFF
--- a/dask_ml/preprocessing/_encoders.py
+++ b/dask_ml/preprocessing/_encoders.py
@@ -1,10 +1,12 @@
 import dask
 import dask.array as da
 import numpy as np
+import packaging.version
 import pandas as pd
 import sklearn.preprocessing
 
 from .label import _encode, _encode_dask_array
+from .._compat import SK_VERSION
 from ..utils import check_array
 
 
@@ -50,6 +52,9 @@ class OneHotEncoder(sklearn.preprocessing.OneHotEncoder):
           numeric values.
 
         The used categories can be found in the ``categories_`` attribute.
+
+    drop : None, default=None
+        The option to drop one of the categories per feature is not yet supported.
 
     sparse : boolean, default=True
         Will return sparse matrix if set True else will return an array.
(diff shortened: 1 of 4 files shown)


jrbourbeau:
Well it looks like I got the tests to pass with scikit-learn `master`...but broke all the other builds. I'll have to go back and add appropriate checks for the `sklearn` version.

jrbourbeau:
Addressing the `sklearn_dev` `DeprecationWarning` in https://github.com/scikit-learn/scikit-learn/pull/13342

jrbourbeau:
The `QuantileTransformer` update in scikit-learn results in a slightly different `transform()` output (ref the conversation starting at https://github.com/scikit-learn/scikit-learn/pull/12827#discussion_r259808030). Because we test against multiple versions of `sklearn`, some of which include this update to `QuantileTransformer` and other that don't, we need to decide how we want to incorporate the update here. I see a couple of ways to do this.

First, we could keep two versions of the relevant method and use the appropriate one depending on which `sklearn` version is being used. Something like:

```python
def _transform_col(self, X_col, quantiles, inverse):
    if SK_VERSION > packaging.version.parse("0.20.2"):
        return self._transform_col_new(X_col, quantiles, inverse)
    else:
        return self._transform_col_legacy(X_col, quantiles, inverse)
```

Second, we can port over the changes made to `QuantileTransformer` in `sklearn` and add a small allowed tolerance, e.g. `atol=1e-7`, when comparing the `sklearn` and `dask_ml` transformations here

https://github.com/dask/dask-ml/blob/e1b21e6dbcfae8341d05d39544ab90dce21fdb2b/tests/preprocessing/test_data.py#L221

Given that `QuantileTransformer` in `dask_ml` already differs from the `sklearn` implementation because it uses approximate quantiles, I have a slight preference for the second option. 

@TomAugspurger do you have a preference here?

jrbourbeau:
I increased the number of samples in the test data for `tests/preprocessing/test_data.py::TestQuantileTransformer` to avoid the new `UserWarning` introduced in https://github.com/scikit-learn/scikit-learn/pull/13333

TomAugspurger:
Option 2 sounds good to me too. Ready to merge?

Thanks a ton for working on this.

jrbourbeau:
Yep, I think this should be good to merge!

PR #476, opened 2019-03-04
Fix imports for isort 4.3.10

jrbourbeau:
The new release of `isort` (version 4.3.10) changed how relative import sorting is handled. This PR makes updates to have v4.3.10 of `isort` pass.

### DIFF
--- a/dask_ml/linear_model/glm.py
+++ b/dask_ml/linear_model/glm.py
@@ -14,8 +14,6 @@ from dask_glm.utils import (
 )
 from sklearn.base import BaseEstimator
 
-# Register multiple dispatch
-from . import utils  # noqa
 from ..utils import check_array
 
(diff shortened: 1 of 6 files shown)


dask_ml/linear_model/glm.py
```diff
@@ -14,8 +14,6 @@
 )
 from sklearn.base import BaseEstimator
 
-# Register multiple dispatch
-from . import utils  # noqa
```
jrbourbeau: Note: this was the only change that was made by me and not by just running `isort`. `black` and `isort` both wanted to change this line in incompatible ways. However, it appears that `utils` wasn't being used in this file, so I just removed the import altogether.

TomAugspurger:
Thanks. Will look into the doc build on travis later.

Issue #488, opened 2019-03-31
Model selection tests are raising ImportError on scikit-learn master 


stsievert:
Here's one traceback for  `tests/linear_model/test_glm.py` on Scikit-Learn master:

``` python-traceback
tests/linear_model/test_glm.py:13: in <module>
    from dask_ml.model_selection import GridSearchCV
dask_ml/model_selection/__init__.py:7: in <module>
    from ._split import ShuffleSplit, KFold, train_test_split
dask_ml/model_selection/_split.py:12: in <module>
    from sklearn.model_selection._split import (
E   ImportError: cannot import name '_validate_shuffle_split_init'
```

Looks like https://github.com/scikit-learn/scikit-learn/pull/13483/ removed `_validate_shuffle_split_init`.

PR #489, opened 2019-04-03
COMPAT: compat with sklearn master

TomAugspurger:
Closes https://github.com/dask/dask-ml/issues/488

### DIFF
--- a/dask_ml/model_selection/_split.py
+++ b/dask_ml/model_selection/_split.py
@@ -9,16 +9,13 @@ import dask.array as da
 import dask.dataframe as dd
 import numpy as np
 import sklearn.model_selection as ms
-from sklearn.model_selection._split import (
-    BaseCrossValidator,
-    _validate_shuffle_split,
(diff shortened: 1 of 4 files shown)


TomAugspurger:
I'm going to irresponsibly merge this. I haven't investigated why these changes were necessary:

1. skip for the spectral clustering test (NaN or inf is apparently showing up somewhere)
2. raised the tolerance for the incremental selection tests

H4dr1en:
Can you release a 0.12.1 including this fix?

TomAugspurger:
I can, but are you having issues with 0.21? I would be surprised if the
changes here have an effect, since they're test-only.

H4dr1en:
Well I am facing #488 too with sklearn==0.21.1 and dask-ml 0.12.0, so if this PR is actually a fix then it would be good to have it. (0.12.0 was released in February)

TomAugspurger:
Ah yeah that would break at import time. I’ll do a release today.

PR #496, opened 2019-04-14
BUG: make IncrementalSearchCV use random_state when sampling parameters

stsievert:
This preserves the `random_state` in `IncrementalSearchCV`'s parameter sampler. Before, the `ParameterSampler` object didn't have `random_state` specified so multiple calls to `IncrementalSearchCV` with the same random state had no effect on the parameters sampled.

**Reference issues**
* Motivating comment: https://github.com/dask/dask-ml/pull/221/files#r272723216


TomAugspurger:
Thanks!

PR #526, opened 2019-06-21
BUG: invalid tol no longer stops on plateau

### LINKED ISSUE #523: Model selection search stops with patience even if tol is invalid or too small
When I create a `SuccessiveHalvingSearchCV` with a `tol` that is infeasible (e.g., `np.nan`), the search that *would* be performed is effected even though it shouldn't be.

This is best characterized by `SuccessiveHalvingSearchCV`, which has `metadata` and `metadata_` attributes to characterize the search that would be performed.

``` python
from sklearn.datasets import make_classification
from dask_ml.model_selection import SuccessiveHalvingSearchCV
from dask_ml.utils import ConstantFunction

X, y = make_classification(n_samples=100, n_features=5)

params = {"value": np.random.RandomState(42).rand(1000)}
model = ConstantFunction()

search = SuccessiveHalvingSearchCV(
    model, params, patience=2, tol=np.nan,
    n_initial_parameters=20, n_initial_iter=2,
)
search.fit(X, y, classes=[0, 1])

print(search.metadata)   # {'partial_fit_calls': 88, 'n_models': 20, 'max_iter': 18}
print(search.metadata_)  # {'partial_fit_calls': 56, 'n_models': 20, 'max_iter': 6}
```

This shouldn't happen; specifying a value of `tol=np.nan` should mean that plateaus are never detected. This should only record the score more frequently (once every two calls).

stsievert:
It's primary purpose is to resolve #523: it makes sure `SuccessiveHalvingSearchCV` *properly* stops on plateau. This is a little more general it turns out, and bled some into `IncrementalSearchCV`.

In addition, it raises a `ValueError` if `patience` is too small. If `patience == 1`, stop-on-plateau will always find a platea (unless `tol` is negative).

**Reference issues/PRs**
This resolves #523.


stsievert:
Small nit: the 1.0 release isn't tagged on GitHub even though I see the 1.0 release on PyPI.

TomAugspurger:
> Small nit: the 1.0 release isn't tagged on GitHub

Fixed. Forgot to push the tag commit.

dask_ml/model_selection/_incremental.py
```diff
@@ -879,10 +879,17 @@ def _get_params(self):
     def _additional_calls(self, info):
         calls = {k: v[-1]["partial_fit_calls"] for k, v in info.items()}
 
+        if self.patience < 0 or self.patience == 1:
```
TomAugspurger: Is patience between 0 and 1 OK?

dask_ml/model_selection/_incremental.py
```diff
@@ -879,10 +879,17 @@ def _get_params(self):
...
+            raise ValueError(
+                "patience={}<=1 will always detect a plateau. "
+                "this, set\n\n    patience >= 2"
+            )
         if self.patience and max(calls.values()) > 1:
             calls_so_far = {k: v[-1]["partial_fit_calls"] for k, v in info.items()}
             adapt_calls = {
-                k: [vi["partial_fit_calls"] + vi.get("_adapt", 0) for vi in v][-1]
+                k: [
```
TomAugspurger: Can this change cause a IndexError? If none of the `vi in v` have an `_adapt` key, then we'll have a empty list, and take the last item of that, causing an IndexError.

dask_ml/model_selection/_incremental.py
stsievert: This is protecting against the default case when `patience=False`. I've made that more explicit in my tests and code.

dask_ml/model_selection/_incremental.py
stsievert: No. `_adapt` will always be in some element of the list because the function `_adapt` will be called first. Otherwise, every value in `calls` is 0 and this indexing will never be performed.

stsievert:
This PR also allows `tol=None` or `tol=np.nan`. This allows recording scores more frequently when `patience` is small.

TomAugspurger:
LGTM. Anything else to do here?

stsievert:
Not to my knowledge

stsievert:
I take that back; I still think some work needs to be done on plateau detection.

TomAugspurger:
Do you think the CI failures are unrelated?

stsievert:
The CI failures should have come up https://github.com/dask/dask-ml/pull/527 (I still get them there). I've manually ripped those values out.

stsievert:
I'm also getting the same error on `test_basic_array` as mentioned in https://github.com/dask/dask-ml/pull/527#issuecomment-505702724.

TomAugspurger:
Merged master. Hopefully CI is green.

TomAugspurger:
Thanks @stsievert. Going to keep an eye on the CI for the next few days, we may have some flaky tests right now.

PR #528, opened 2019-06-24
ENH: provide monitoring of model selection searches

stsievert:
It allows the user to monitor the model selection search. It prints out information on the score received ~~and allows the user to stop the searches via Cntrl-C.~~

* If `verbose` is False (default): log all the time but don't pipe to stdout (aka print)
* If `verbose` is True: log and print all the time
* If `verbose` is an int/float and `0 < verbose <= 1`: log and print `verbose` fraction of the time.
* If `verbose` is zero: do not log past initialization.

This allows:

* The user to see logs in Jupyter notebooks by specifying `verbose=True`
* Advanced users to configure logging themselves by default with `verbose=False`.
* Advanced users to turn off logging with `verbose=0`.

All types are cast to float except bool; valid values include `verbose in [0.0, 1, True, False, 0.5]`. An error will be raised if `not 0 <= verbose <= 1`.


stsievert:
I'm not sure this PR has the best method of printing/logging information:

https://github.com/dask/dask-ml/blob/bde46b74d9a9ee635e431edd184b1d4b386ea1b3/dask_ml/model_selection/_incremental.py#L112-L115

I'm also unsure how to test `KeyboardInterrupt`s.

dask_ml/model_selection/_incremental.py
```diff
@@ -106,6 +109,12 @@ def _create_model(model, ident, **params):
         return model, {"model_id": ident, "params": params, "partial_fit_calls": 0}
 
 
+def _show_msg(msg, verbose):
```
TomAugspurger: Instead of this, I'd prefer to just use the standard logging module for this. Users can get the info printed to stdout by adding their own handler.

We could consider doing that for them if they make an estimator with `verbose=True`.

```python
handler = logging.StreamHandler()
handler.setLevel(logging.INFO)

logger.addHandler(handler)
```

It may be good to get the log formatter from distributed and apply that.

We'd need to make sure to only add the handler once (maybe the logging module does that already).

dask_ml/model_selection/_incremental.py
```diff
@@ -118,7 +127,10 @@ def _fit(
     fit_params=None,
     scorer=None,
     random_state=None,
+    verbose=False,
 ):
+    _show_msg("[CV] train examples = {}".format(len(y_train)), verbose)
```
TomAugspurger: With logging you typically use `%`-style formatters, and pass the values in. This skips the formatting step if it's not needed.

```
logger.info([CV] train examples = %d, len(y_train))
```

stsievert:
I'm having difficult time with implementation that catches the `KeyboardInterrupt`. When I test it manually in a notebook, the code throws an `KeyboardInterrupt` when I manually send a single keyboard interruption and does not stop the training cleanly. Here's the traceback of the error I get:

``` python-traceback
KeyboardInterrupt               Traceback (most recent call last)
<ipython-input-5-3748455ece57> in <module>()
----> 9 df = _test_keyboardinterrupt()
<ipython-input-4-135866acf811> in _test_keyboardinterrupt()
---> 13     search.fit(X, y)
~/Developer/stsievert/dask-ml/dask_ml/model_selection/_incremental.py in fit(self, X, y, **fit_params)
--> 641         return default_client().sync(self._fit, X, y, **fit_params)
~/anaconda3/lib/python3.6/site-packages/distributed/client.py in sync(self, func, *args, **kwargs)
--> 753             return sync(self.loop, func, *args, **kwargs)
~/anaconda3/lib/python3.6/site-packages/distributed/utils.py in sync(loop, func, *args, **kwargs)
--> 329             e.wait(10)
~/anaconda3/lib/python3.6/threading.py in wait(self, timeout)
--> 551                 signaled = self._cond.wait(timeout)
~/anaconda3/lib/python3.6/threading.py in wait(self, timeout)
--> 299                     gotit = waiter.acquire(True, timeout)

KeyboardInterrupt: 
```

TomAugspurger:
I'm learning a bit about this for a distributed issue. Will take a look later today.

TomAugspurger:
Can you talk a bit about what the expected behavior of a KeyboardInterrupt halfway through is? As a user, I might expect the following

1. All currently scheduled fits are cancelled (cluster should become idle)
2. The usual attributes like `best_estimator_` are present, and reflect the state of the world when the training was interrupted.
3. I have some way of knowing that the training was stopped early? An attribute on the model?

Overall, this looks a bit more complicated that adding better logging. It may be worth splitting into its own PR.

stsievert:
> 1. All currently scheduled fits are cancelled (cluster should become idle)
> 2. The usual attributes like best_estimator_ are present, and reflect the state of the world when the training was interrupted.
> 3. I have some way of knowing that the training was stopped early? An attribute on the model?

I would certainly expect 1 and 2. I intend the `KeyboardInterrupt` to be a mechanism for the busy data scientist to get on with their workflow (using the best estimator, etc).

TomAugspurger:
Looks like a merge conflict.

stsievert:
Resolved. I've also personally verified that the messages are printed in a notebook (so more than just with pytest).

stsievert:
There's still some debugging work to do. The logger isn't being created correctly I think for Hyperband.

stsievert:
Okay, resolved. Now there's no excessive printing when `BaseIncrementalSearchCV` has multiple instantiations in the same script (as per Hyperband).

dask_ml/model_selection/_hyperband.py
```diff
@@ -303,6 +303,7 @@ def __init__(
         test_size=None,
         random_state=None,
         scoring=None,
+        verbose=False,
```
TomAugspurger: Need docs.

dask_ml/model_selection/_incremental.py
```diff
@@ -118,7 +122,14 @@ def _fit(
     fit_params=None,
     scorer=None,
     random_state=None,
+    verbose=False,
 ):
+    prefix = ""
+    if isinstance(verbose, tuple):
```
TomAugspurger: Is this standard in scikit-learn?

stsievert: I'm not sure. I did this to simplify the data flow and didn't have to add another parameter to a bunch of private functions. This `verbose` originates in `SuccessiveHalvingSearchCV` initialization.

dask_ml/model_selection/_hyperband.py
stsievert: Docs added.

dask_ml/model_selection/_incremental.py
stsievert: I've removed this. Instead, I set an attribute (`'_prefix'`) on `BaseIncrementalSearchCV` and pass that to `_incremental.fit` (with `getattr(self, "_prefix", "")`).

stsievert:
I've cleaned up this PR. Details are in https://github.com/dask/dask-ml/pull/528#discussion_r336763815. It cleans up the API, and instead sets the prefix for Hyperband through a private attribute.

dask_ml/model_selection/_incremental.py
```diff
@@ -471,6 +503,16 @@ def _check_array(self, X, **kwargs):
         X = check_array(X, **kwargs)
         return X
 
+    def _handle_verbosity(self, verbose):
```
TomAugspurger: I'm not sure that a `.fit` call should affect this. Ideally all the configuration would be done when the module is loaded. You might want to look at how distributed handles logging. I think by default we should follow the config setting there, and users can override it if needed.

dask_ml/model_selection/_hyperband.py
```diff
@@ -341,9 +349,10 @@ def _get_SHAs(self, brackets):
                 test_size=self.test_size,
                 random_state=seed_start + b if b != 0 else self.random_state,
                 scoring=self.scoring,
+                verbose=self.verbose,
             )
-            for b, (n, r) in brackets.items()
-        }
+            sha._prefix = ", bracket=" + str(b)
```
TomAugspurger: FWIW, I think it's fine to add `prefix` as a parameter to SuccessiveHalvingSearchCV, rather than setting it after the fact here. That makes it a bit more robust to thinks like `clone`.

stsievert: Now, only `SuccessiveHalvingSearchCV` has  a `prefix` argument. `IncrementalSearchCV` is un-modified (which made for a slightly unorthodox implementation).

dask_ml/model_selection/_hyperband.py
```diff
@@ -139,6 +139,11 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
 
         If None, the estimator's default scorer (if available) is used.
 
+    verbose : bool, int, optional, default: False
+        If ``True``, print the best validation score received to stdout every
```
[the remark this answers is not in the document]
stsievert: I now mirror distributed's logging (and only set `logger = logging.getLogger(__name__)` at the top.

I've also expanded on the documentation on how to view the logs on stdout: it looks like

``` python
    verbose : bool, int, optional, default: False
        If ``True``, log the best validation score received every
        time possible. If an integer, print ``1 / verbose``
        percent of the time.

        The logger is available under the name ``dask_ml.model_selection``.
        Logs can be piped to stdout with
        ``logging.basicConfig(level=logging.INFO)``,
        or with

        .. code:: python

           import logging
           logger = logging.getLogger('dask_ml.model_selection')
           logger.setLevel(logging.INFO)
           logger.addHandler(logging.StreamHandler())
```

Let me know if that's right.

dask_ml/model_selection/_incremental.py
```diff
@@ -25,6 +28,7 @@
 from ._split import train_test_split
 
 Results = namedtuple("Results", ["info", "models", "history", "best"])
+logger = logging.getLogger(__name__)
```
TomAugspurger: I'm not sure, but I think this should be

```suggestion
logger = logging.getLogger("dask_ml.model_selection")
```

We want users to configure `'dask_ml.model_selection'`, not 'dask_ml.model_selection._incremental'`

alternatively, in `dask_ml/model_selection/__init__.py` you can do


```python
logger = logging.getLogger(__name__)
```

and try `from . import logger`, but that may cause a circular import.

TomAugspurger:
https://docs.python.org/3/howto/logging-cookbook.html#using-a-context-manager-for-selective-logging may be helpful for the "temporarily set the logging context".

stsievert:
Thanks for the link in https://github.com/dask/dask-ml/pull/528#issuecomment-555201253. That helped with the changes in https://github.com/dask/dask-ml/pull/528#pullrequestreview-318613281

I re-configured the API to do the following:

1. Have `verbose=True` configure verbosity for the user
2. Lower values of verbose mean "be less verbose". i.e., if `0 < verbose < 1`, only print about `verbose` percent of the time.

stsievert:
I can test that the messages are received by logging.INFO fine. However, I'm having some difficulty testing that the messages are sent to stdout when `verbose=True`. It works fine when I manually test it (with a Jupyter notebook), but I'm having a hard time getting the test to reflect that.

stsievert:
I've weakened the tests a bit; I expect the CI to be green (the travis CI looks related though?). The implementation works in a Jupyter notebook, and pipes the logging messages to stdout.

TomAugspurger:
@stsievert can you give a summary of where this is at? It seems like `verbose=True` will auto-configure the logging for the user.

What happens when that's not set. IMO, we would support pre-configured loggers. So if a user has configured some handler for `dask_ml.model_selection`, we would just automatically use it. In particular, we would still make calls to `logger.info(...)` even when `verbose=False`. There just wouldn't be anyone who sees it.

stsievert:
Here's the what this PR implements:

> * If `verbose` is False (default): log all the time but don't pipe to stdout
> * If `verbose` is True: log and print all the time
> * If `verbose` is an float and `0 < verbose <= 1`: log and print `verbose` percent of the time.
>
> —https://github.com/dask/dask-ml/pull/528#issue-290931610

I've left the interrupting this search via `KeyboardInterrupts` for another PR I think, as mentioned in https://github.com/dask/dask-ml/pull/528/#issuecomment-506484960 (which I'd also like to see).

dask_ml/model_selection/_incremental.py
```diff
@@ -119,7 +124,12 @@ def _fit(
     fit_params=None,
     scorer=None,
     random_state=None,
+    verbose: int = 1,
+    prefix="",
 ):
+    logger.info(
+        "[CV%s] train, test examples = %d, %d", prefix, len(y_train), len(y_test)
```
TomAugspurger, line 131: What's the type on `y_train`? If it's potentially a dataframe then `len` will be expensive. I'd prefer to omit this, in case we're able to support dataframes eventually.

dask_ml/model_selection/_incremental.py
```diff
@@ -205,9 +216,15 @@ def get_futures(partial_fit_calls):
     start_time = time()
 
     # async for future, result in seq:
-    while True:
+    for _i in itertools.count():
         metas = yield client.gather(new_scores)
 
+        if _i % int(verbose) == 0:
+            idx = np.argmax([m["score"] for m in metas])
```
TomAugspurger, line 223: Can `metas` ever be empty here, causing a ValueError from taking the max of an empty sequence?

dask_ml/model_selection/_incremental.py
```diff
@@ -287,6 +304,8 @@ def fit(
     fit_params=None,
     scorer=None,
     random_state=None,
+    verbose=1,
```
TomAugspurger, line 307: Can you make the default here match the others (False)?

dask_ml/model_selection/_incremental.py
```diff
@@ -326,6 +345,11 @@ def fit(
         If RandomState instance, random_state is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`.
+    verbose : int, default=1
```
TomAugspurger, line 348: I think int or bool

dask_ml/model_selection/_incremental.py
```diff
@@ -560,7 +590,7 @@ def _check_is_fitted(self, method_name):
         return check_is_fitted(self, "best_estimator_")
 
     @gen.coroutine
-    def _fit(self, X, y, **fit_params):
+    def _fit(self, X, y, verbosity=1, **fit_params):
```
TomAugspurger, line 593: Why verbosity rather than verbose?

TomAugspurger: Also, can't this just be `self.verbose`? Why do we need a fit param?

dask_ml/model_selection/_incremental.py
```diff
@@ -614,7 +646,34 @@ def fit(self, X, y=None, **fit_params):
         **fit_params
             Additional partial fit keyword arguments for the estimator.
         """
-        return default_client().sync(self._fit, X, y, **fit_params)
+        if isinstance(self.verbose, float) and not (0 < self.verbose <= 1):
```
TomAugspurger, line 649: Can the verbosity / verbose issue be avoided by doing this in `self._fit`?

dask_ml/model_selection/_incremental.py
```diff
@@ -614,7 +646,34 @@ def fit(self, X, y=None, **fit_params):
...
+        else:
+            msg = "verbose must be an instance of float or bool. Got type={}"
+            raise TypeError(msg.format(type(self.verbose)))
+
+        kwargs = dict(verbosity=verbosity, **fit_params)
+        if self.verbose:
+            logger.setLevel(logging.INFO)
+            h = logging.StreamHandler(sys.stdout)
+            self._h = h
+            self._logging_context = LoggingContext(
```
TomAugspurger, line 668: Do we need to assign this? Why can't we just do `with LoggingContext()...`?

Also, let's keep the call to `sync` the same.

```
if self.verbose:
    context = LoggingContext(...)
else:
    context = contextlib.nullcontext()

with context:
    return default_client().sync(...)
```

dask_ml/model_selection/_incremental.py
```diff
@@ -614,7 +646,34 @@ def fit(self, X, y=None, **fit_params):
...
+            verbosity = 1
+        elif isinstance(self.verbose, float):
+            verbosity = np.round(1 / self.verbose)
+        else:
+            msg = "verbose must be an instance of float or bool. Got type={}"
+            raise TypeError(msg.format(type(self.verbose)))
+
+        kwargs = dict(verbosity=verbosity, **fit_params)
+        if self.verbose:
+            logger.setLevel(logging.INFO)
```
TomAugspurger, line 665: This concerns me. We shouldn't be modifying the logging configuration permanently. Can we do this just inside LoggingContext, and undo any changes we make?

dask_ml/model_selection/_incremental.py
stsievert: Thanks; this is a meaningless commit message anyways.  `*_train` and `*_test` will be Dask Arrays as per #637.

Now, I'm providing a more meaningful log message on the number of examples in each chunk (which gets at the chunk size balancing, an issue I ran into).

dask_ml/model_selection/_incremental.py
stsievert: No, `metas` could never be empty.

Why? Values gets added to `new_scores` for every non-zero value of `instructions: Dict[str, int]`. If `sum(instructions.values()) == 0` the loop is broken.

dask_ml/model_selection/_incremental.py
stsievert: I've renamed `verbose` to `log_freq` to make it more clear.

dask_ml/model_selection/_incremental.py
stsievert: That's a lot simpler!

dask_ml/model_selection/_incremental.py
```diff
@@ -111,15 +118,22 @@ def _create_model(model, ident, **params):
...
     model,
     params,
-    X_train,
-    y_train,
-    X_test,
-    y_test,
+    X_train: da.Array,
+    y_train: da.Array,
+    X_test: da.Array,
+    y_test: da.Array,
```
TomAugspurger, line 124: Let's hold of on adding these types, I don't want it holding up this PR. I think the test versions at least can be numpy arrays too.

dask_ml/model_selection/_incremental.py
```diff
@@ -154,6 +169,10 @@ def _fit(
     y_train = sorted(futures_of(y_train), key=lambda f: f.key)
     assert len(X_train) == len(y_train)
 
+    train_eg = yield client.map(len, y_train)
```
TomAugspurger, line 172: I don't understand this map call. `y_train` is a Dask array right?

dask_ml/model_selection/_incremental.py
```diff
@@ -111,15 +118,22 @@ def _create_model(model, ident, **params):
...
-    y_test,
+    X_train: da.Array,
+    y_train: da.Array,
+    X_test: da.Array,
+    y_test: da.Array,
     additional_calls,
     fit_params=None,
     scorer=None,
     random_state=None,
+    log_freq: float = 1.0,
```
TomAugspurger, line 129: log_freq -> log_frequency here and elsewhere.

Is this a standard in scikit-learn? Or are we starting a new pattern with log_freq.

dask_ml/model_selection/_incremental.py
```diff
@@ -564,6 +602,19 @@ def _fit(self, X, y, **fit_params):
...
+            # If True, logger will be configured to pipe to stdout
+            log_freq = 1.0
+        elif isinstance(self.verbose, float):
+            if not (0 < self.verbose <= 1):
+                msg = "verbose={} must satisfy 0 < verbose <= 1"
+                raise ValueError(msg.format(self.verbose))
+            log_freq = self.verbose
+        else:
+            msg = "verbose must be an instance of float or bool. Got type={}"
+            raise TypeError(msg.format(type(self.verbose)))
```
TomAugspurger, line 616: Does this match the behavior of scikit-learn? You can't pass an integer like `verbose=1`?

dask_ml/model_selection/_incremental.py
```diff
@@ -614,7 +667,15 @@ def fit(self, X, y=None, **fit_params):
         **fit_params
             Additional partial fit keyword arguments for the estimator.
         """
-        return default_client().sync(self._fit, X, y, **fit_params)
+        if self.verbose:
+            h = logging.StreamHandler(sys.stdout)
+            context = LoggingContext(logger, level=logging.INFO, handler=h)
+            self._logging_context = context  # for testing ease; not necessary
```
TomAugspurger, line 673: Let's avoid adding this just for testing. There are other ways to test that logging occurred: https://docs.pytest.org/en/latest/logging.html#caplog-fixture

dask_ml/model_selection/_successive_halving.py
```diff
@@ -211,7 +218,16 @@ def __init__(
             tol=tol,
             random_state=random_state,
             scoring=scoring,
+            verbose=verbose,
         )
+        self.n_initial_parameters = n_initial_parameters
+        self.n_initial_iter = n_initial_iter
+        self.aggressiveness = aggressiveness
+
+        # unusual ordering allows BaseIncrementalSearchCV to have a prefix
```
TomAugspurger, line 227: What does "unusual ordering" mean here?

dask_ml/model_selection/_incremental.py
stsievert: `y_train` is a list of futures to ndarrays. See the reassignment with `future_of` three lines above this.

``` 
>>> type(client.compute(y_train))
list
>>> [type(x) for x in client.compute(y_train)]
[<class 'numpy.ndarray'>, <class 'numpy.ndarray'>]
>>> [x.shape for x in client.compute(y_train)]
[(20,), (20,)]
```

dask_ml/model_selection/_incremental.py
stsievert: Now `verbose: Union[bool, int]` print `verbose` percent of the time if specified as an integer. That satisfies my constraint that higher values of `verbose` print more often.

dask_ml/model_selection/_successive_halving.py
stsievert: Explained in 9df6914d0d920c19fe9bb9925c3c0a2943098533

stsievert:
I've updated https://github.com/dask/dask-ml/pull/528#issue-290931610:

> *    If verbose is False (default): log all the time but don't pipe to stdout
> *    If verbose is True: log and print all the time
> *    If verbose is an int between 1 and 100: log and print verbose percent of the time.

I still don't like the difference between specifying `verbose=True` and `verbose=1`. If `verbose=True` this will log all the time; if `verbose=1` this will log 1% of the time. That's the primary motivation for making `verbose` either a float or bool.

stsievert:
`verbose` is now a float as before, but I always cast to a float now (unless a bool is given). I've updated https://github.com/dask/dask-ml/pull/528#issuecomment-613219797 and https://github.com/dask/dask-ml/pull/528#issue-290931610 to reflect this change.

stsievert:
The CI is failing because `contextlib.nullcontext` is new in Python 3.7 ([source]). Do we want to upgrade the minimum Python version or use a `NullContext` in `dask_ml._utils.py`?

[source]:https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext

TomAugspurger:
Ah, I didn't realize it was that new. I think defining something like this in `dask_ml._compat`

```python
@contextlib.contextmanager
def nullcontext():
    yield
```

should suffice

dask_ml/model_selection/_successive_halving.py
TomAugspurger: In general, subclasses should implement a superset of their parent's API: https://en.wikipedia.org/wiki/Liskov_substitution_principle

So if the only reason for this is `prefix` not really being necessary for IncrementalSearchCV, I'd prefer to just add it.

stsievert: My motivation more clearly: I don't want to expose `prefix` in a public API (like in `IncrementalSearchCV.__init__`). I think it's an internal detail; Scikit-learn doesn't expose it in [RandomizedSearchCV]. I'd prefer not to an attribute after the fact (`search._prefix = ...`).

Maybe the best solution is to create a new dummy class `_IncrementalSearchCV` to pass `prefix` to `BaseIncrementalSearchCV`. Then, `IncrementalSearchCV` and `SuccessiveHalvingSearchCV` could inherit from the new class.

[RandomizedSearchCV]:https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.RandomizedSearchCV.html

TomAugspurger: I don't have a strong opinion on whether or not `prefix` is in the public API, but I do care that all of these have compatible signature. So if we need `prefix` in the base class so that we can consistently do logging, then we need to have it in all of the signatures.

stsievert: I've satisfied both our concerns with a private class to pass `prefix` to `BaseIncrementalSearchCV`:

* `prefix` is not in the public API
* All subclasses implement a superset of the parent's API.

Briefly, the implementation looks like this:

``` python
class BaseIncrementalSearchCV(...):
    ...
class _IncrementalSearchCV(BaseIncrementalSearchCV):
    ...
class IncrementalSearchCV(_IncrementalSearchCV):
    ...
class SuccessiveHalvingSearchCV(_IncrementalSearchCV):
    ...
```

stsievert:
I've also deprecated the `scores_per_fit` parameter in `IncrementalSearchCV`; the name should be `fits_per_score`.

dask_ml/model_selection/_successive_halving.py
TomAugspurger: I'll look more closely, but I'm not a fan of the extra indirection. And I think `BaseIncrementalSearchCV` has the parameter, and so all the subclasses (including IncrementalSearchCV) need it.

It also seems like it might be a generally useful parameter? If I'm running multiple IncrementalSearchCV's simultaneous I might want to provide a prefix to monitor their performance?

stsievert: > might want to provide a prefix to monitor their performance?

I suppose in the context of Dask it does make sense – that's the motivation for `prefix` in `SuccessiveHalvingSearchCV` too, and I can see how it could apply to my current hyperparameters optimization task.

I'll add `prefix` to the public API for `IncrementalSearchCV`.

stsievert: ...but the algorithm-specific parameter `decay_rate` is accepted by `IncrementalSearchCV`. `decay_rate` is not used by any other adaptive algorithm, including `SuccessiveHalvingSearchCV`.

I think that's work for #432, not this PR.

tests/model_selection/test_incremental.py
```diff
@@ -731,3 +822,17 @@ def test_search_invalid_patience(c, s, a, b):
...
+
+def test_warns_scores_per_fit():
+    X, y = make_classification(n_samples=100, n_features=5, chunks=10)
+
+    params = {"value": np.random.RandomState(42).rand(1000)}
+    model = ConstantFunction()
+
+    with pytest.warns(UserWarning, match="deprecated since Dask-ML v1.4.0"):
+        IncrementalSearchCV(model, params, scores_per_fit=2)
+    if datetime.now() >= datetime(2021, 1, 1):
```
stsievert, line 835: I'm unsure about including this date in the test. How should this be handled? I don't want to let this be deprecated forever.

TomAugspurger: This can be removed. We'll update the test when the deprecation is enforced.

To make sure I understand, `scores_per_fit` and `fits_per_score` have identical semantics, we're just giving it a better name? We might just make an API breaking change in that case.

stsievert: Yup, the renaming is purely cosmetic. Should I leave the warning in or raise a `ValueError`?

TomAugspurger: IMO we either just make the change (and remove this test) or we leave in the deprecation and assert that a warning is raised (and remove this date-dependent assert).

stsievert: I'd rather leave a warning in and remove it shortly. I've pushed that change, and it looks like the CI failure is unrelated (documentation warnings in `hyper-parameter-search.rst`, which is unmodified in this PR).

dask_ml/model_selection/_incremental.py
```diff
@@ -326,6 +356,14 @@ def fit(
         If RandomState instance, random_state is the random number generator;
         If None, the random number generator is the RandomState instance used
         by `np.random`.
+    verbose : bool, int, float, default=False
```
TomAugspurger, line 359: Can you make all of these consistent? Assuming they have the same meaning everywhere.

I think the next line is incorrect, or at least unclear. The default is False so people won't see anything by default unless they've configured logging externally.

dask_ml/model_selection/_incremental.py
```diff
@@ -686,10 +745,16 @@ class IncrementalSearchCV(BaseIncrementalSearchCV):
...
 
+    fits_per_scores : int, optional, default=1
+        If ``patience`` is used the maximum number of ``partial_fit`` calls
+        between ``score`` calls.
+
     scores_per_fit : int, default 1
         If ``patience`` is used the maximum number of ``partial_fit`` calls
         between ``score`` calls.
 
+        *Deprecated in Dask-ML v1.4.0.*
```
TomAugspurger, line 756: You can use the deprecated directive: http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

dask_ml/model_selection/_incremental.py
```diff
@@ -935,17 +936,17 @@ def __init__(
         scoring=None,
         verbose=False,
         prefix="",
-        scores_per_fit=None,
+        **kwargs,
```
TomAugspurger, line 939: We don't want to have kwargs here. Too easy to silently ignore typos.

dask_ml/model_selection/_incremental.py
```diff
@@ -853,17 +930,29 @@ def __init__(
...
-        scores_per_fit=1,
+        fits_per_score=1,
         max_iter=100,
         random_state=None,
         scoring=None,
+        verbose=False,
+        prefix="",
+        scores_per_fit=None,
     ):
+        if scores_per_fit is not None and fits_per_score != 1:
```
TomAugspurger, line 941: Sorry, didn't realize this deprecation was in the `__init__`. This should be done in the `fit`. `__init__` isn't supposed to do any validation.

This will also fix the issue you saw with the warnings in the doc build.

stsievert: D'oh, of course. I've made that change, and removed the `:okwarning:`s. This PR should be ready merge now.

TomAugspurger:
Thanks!

PR #527, opened 2019-06-21
ENH: prioritizes fitting by score in model selection

stsievert:
**What does this PR implement?**
This PR prioritizes model training/scoring based on the most recent score in model selection.

PR #529, opened 2019-06-26
trigger CI

TomAugspurger:


stsievert:
Fyi, I edited https://github.com/dask/dask-ml/pull/527#issuecomment-505702724. Now, it makes it more clear that I see the `TimeoutError` on my machine and on the CI. On my machine, I get this error on the `ms-priority` branch and not on master.

PR #530, opened 2019-07-05
DOC Joblib is no longer vendored in scikit-learn

rth:
scikit-learn, as of 0.21.0, no longer vendors joblib which [became a dependency](https://github.com/scikit-learn/scikit-learn/blob/a243d96336cb4f50ca3635b3062a273f3dc5183a/setup.py#L231).

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -48,11 +48,11 @@ optimization, and more.
 .. code-block:: python
 
    from dask.distributed import Client
-   from sklearn.externals.joblib import parallel_backend
+   import joblib
 
    client = Client()  # Connect to a Dask Cluster
 
-   with parallel_backend('dask'):
+   with joblib.parallel_backend('dask'):
        # Your normal scikit-learn code here
 
 See :doc:`Dask-ML Joblib documentation <joblib>` for more information.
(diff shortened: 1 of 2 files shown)


stsievert:
The CI failures are unrelated. Here's some of the output from the tests:

<details>
Here's the output from `linux sklearnDev`:

<pre>
    @functools.wraps(fun)
    def wrapped(*args, **kwargs):
>       warnings.warn(msg, category=DeprecationWarning)
E       DeprecationWarning: Function assertEqual is deprecated; This helper is deprecated in version 0.22 and will be removed in version 0.24. Please use "assert" instead
</pre>

There are a bunch of tests in `tests/test_pca.py` that are failing because of this. Here's the output from pytest on `linux earliest`:

<pre>
@pytest.mark.xfail(
		condition=DASK_200, reason="https://github.com/dask/dask/issues/5008"
)
def test_unknown_category_transform_array():
	x2 = da.from_array(np.array([["a"], ["b"], ["c"], ["d"]]), chunks=2)
	enc = dask_ml.preprocessing.OneHotEncoder()
	enc.fit(dX)

	result = enc.transform(x2)
	with pytest.raises(ValueError) as e:
				result.compute()

	assert e.match("Block contains previously")
	>       assert "d" in str(e)
	E       AssertionError: assert 'd' in '<ExceptionInfo ValueError tblen=10>'
	E        +  where '<ExceptionInfo ValueError tblen=10>' = str(<ExceptionInfo ValueError tblen=10>)
</pre>
</details>

TomAugspurger:
Thanks @rth!

PR #532, opened 2019-07-15
WIP: allow not running passive bracket in HyperbandSearchCV

stsievert:
*This PR is not ready to be merged. I still need to verify the implementation.*

This PR allows skipping the bracket that trains all models to completion if `bool(patience)`.  This bracket is likely of marginal use because the hyperparameters will (almost certainly) influence the score. Why not eliminate this bracket instead of designing `patience=True` to accommodate this bracket?

**Reference issues/PRs**
* the prioritization scheme (#527). This means higher scoring models are run first.
* (future) keyboard interrupts (https://github.com/dask/dask-ml/pull/528#issuecomment-506484960).

These two features mean that the user can exit out of hyper-parameter selection when the score is high enough.

**TODO**

- [x] implement
- [ ] verify that `bracket=0` returns low scores
- [ ] run simulations to see how much this time-to-solution
- [ ] document

[my SciPy19 talk]:https://www.youtube.com/watch?v=x67K9FiPFBQ

### DIFF
--- a/dask_ml/model_selection/_hyperband.py
+++ b/dask_ml/model_selection/_hyperband.py
@@ -317,7 +317,7 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
             scoring=scoring,
         )
 
-    def _get_SHAs(self, brackets):
+    def _get_SHAs(self, brackets, all_shas=False):
         patience = _get_patience(
             self.patience, self.max_iter, self.aggressiveness, self.tol
         )
@@ -328,8 +328,12 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
         self._SHA_seed = seed_start
 
         # These brackets are ordered by adaptivity; bracket=0 is least adaptive
-        SHAs = {
-            b: SuccessiveHalvingSearchCV(
+        SHAs = {}
+        for b, (n, r) in brackets.items():
+            if not all_shas and b == 0 and patience:
+                continue
+
+            SHA = SuccessiveHalvingSearchCV(
                 self.estimator,
                 self.parameters,
                 n_initial_parameters=n,
@@ -342,8 +346,7 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
                 random_state=seed_start + b if b != 0 else self.random_state,
                 scoring=self.scoring,
             )
-            for b, (n, r) in brackets.items()
-        }
+            SHAs[b] = SHA
         return SHAs
 
     @gen.coroutine
@@ -412,7 +415,7 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
         best_index = best_index.flat[0]
 
         meta, _ = _get_meta(
-            {b: SHA.history_ for b, SHA in SHAs.items()}, brackets.keys(), SHAs, key
+            {b: SHA.history_ for b, SHA in SHAs.items()}, SHAs.keys(), SHAs, key
         )
 
         self.metadata_ = {
@@ -445,7 +448,7 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
         bracket_info = list(reversed(sorted(bracket_info, key=lambda x: x["bracket"])))
 
         brackets = _get_hyperband_params(self.max_iter, eta=self.aggressiveness)
-        SHAs = self._get_SHAs(brackets)
+        SHAs = self._get_SHAs(brackets, all_shas=True)
         for bracket in bracket_info:
             b = bracket["bracket"]
             bracket["SuccessiveHalvingSearchCV params"] = _get_SHA_params(SHAs[b])
--- a/tests/model_selection/test_hyperband.py
+++ b/tests/model_selection/test_hyperband.py
@@ -419,3 +419,14 @@ def test_history(c, s, a, b):
     for model_hist in alg.model_history_.values():
         calls = [h["partial_fit_calls"] for h in model_hist]
         assert (np.diff(calls) >= 1).all() or len(calls) == 1
+
+
+@gen_cluster(client=True, timeout=5000)
+def test_patience_brackets(c, s, a, b):
+    X, y = make_classification(n_samples=10, n_features=4, chunks=10)
+    model = ConstantFunction()
+    params = {"value": scipy.stats.uniform(0, 1)}
+    search = HyperbandSearchCV(model, params, max_iter=9, patience=True)
+    yield search.fit(X, y)
+
+    assert all("bracket=0" not in h for h in search.history_)


stsievert:
This PR is motivated by a question from @amueller after [my SciPy19 talk]. Why run this bracket if it's not the best performing? In simulations, `bracket=0` is the best bracket 3.5% of the time:

Screen Shot 2019-07-14 at 9 36 39 PM

This histogram is from the 200 runs of `HyperbandSearchCV` I did in https://github.com/stsievert/dask-hyperband-comparison.

[my SciPy19 talk]:https://www.youtube.com/watch?v=x67K9FiPFBQ

amueller:
Thanks, that's cool! The question I'm most interested in is: what if you only ran bracket 4 in all cases? Would it be substantially worse than running hyperband?

stsievert:
> what if you only ran bracket 4 [the most aggressive bracket] in all cases?

I'll take your question as "how can the search be made more aggressive?"

The aggressiveness of Hyperband's early stopping scheme can be configured (which I accidentally forgot to mention in the talk!<sup>*</sup>). This is controlled by `HyperbandSearchCV`'s `aggressiveness` optional input, which controls how many models are kept when some models are stopped and how long to train between stoppings.

By default, `aggressiveness=3`. If changed, Li et. al. recommended setting `aggressiveness` to 4:

> The value of η [aka `aggressiveness`] can be viewed as a knob that can be tuned based on practical user constraints. Larger values of η correspond to a more aggressive elimination schedule and thus fewer rounds elimination ... If our [infinite horizon] theoretical bounds are optimized (see Section 4) they suggest choosing η = e ≈ 2.718 but in practice we suggest taking η to be equal to 3 or 4 (if you don’t know how to choose η, use η = 3).
> ...
> If overhead is not a concern and aggressive exploration is desired, one can (1) increase η to reduce the number of brackets while maintaining R as the maximum number of configurations in the most exploratory bracket or (2) still use η = 3 or 4 but only try brackets that do a baseline level of exploration,
> —Section 2.6 of the [Hyperband paper].

`aggressiveness=4` in https://github.com/dask/dask-ml/pull/532#issuecomment-511260472. This PR removes the least aggressive bracket(s).

It's also possible to only run bracket 4:

``` python
from dask_ml.model_selection import HyperbandSearchCV, SuccessiveHalvingSearchCV
search = HyperbandSearchCV(model, params, max_iter=max_iter)

bracket_params = search.metadata["brackets"][4]["SuccessiveHalvingSearchCV params"]
bracket = SuccessiveHalvingSearchCV(model, params, **bracket_params)
```

This will require the user deciding how important the training data is, and which bracket to choose.

<sup>* I have removed that typo from [the slides]</sup>

[the slides]:https://github.com/stsievert/talks
[Hyperband paper]:http://www.jmlr.org/papers/volume18/16-558/16-558.pdf

amueller:
Aggressiveness is not what I meant, that's just the base of the "halving", right?
Unless I really misunderstood something. The eta/agressiveness is the same for all brackets, right? Only the number of hyper-parameters that are sampled and the minimum resources allocated are different in each bracket, right?

What I meant was what you showed last. The paper basically says that just running the bracket that puts the least emphasis on the amount of training data basically always wins, and in practice Successive Halving is basically as good as Hyperband.

stsievert:
> that's just the base of the "halving", right? The eta/agressiveness is the same for all brackets, right? Only the number of hyper-parameters that are sampled and the minimum resources allocated are different in each bracket, right?

All correct.

> What I meant was what you showed last. The paper basically says that just running the bracket that puts the least emphasis on the amount of training data basically always wins, and in practice Successive Halving is basically as good as Hyperband.

Good. Yes, almost always<sup>*</sup> (there's one case where the second most exploratory bracket is optimal, and it has a narrow search space). The other experiments all run Hyperband (via `hyperband`) alongside repeats of the most exploratory bracket (via `bracket s=4`):

Screen Shot 2019-07-17 at 12 11 48 PM


Looks like repeating the most exploratory bracket finds slight lower losses than Hyperband on average, at least in their example. I've also simulated this with my data  (the same example as https://github.com/dask/dask-ml/pull/532#issuecomment-511260472):

Screen Shot 2019-07-17 at 12 29 21 PM

Here "2 repeats" means "2 repeats of the most exploratory bracket". It looks like 2 or 3 runs of the most exploratory bracket recovers the Hyperband performance, at least for this example.

I'll run some more simulations to see how this performs over time. I'll probably edit the search space a bit to make it more difficult for the most adaptive bracket to be optimal.

<sup>* The one exception finds the second most aggressive bracket is optimal in Figure 4 (which runs [LeNet])</sup>

[LeNet]:http://deeplearning.net/tutorial/lenet.html

amueller:
Thanks. Yes this is what I meant ;)
What's the x axis on your plot? Number of configurations run?
And how do you simulate the repeats?
How do repeats compare against running more configs in parallel?

stsievert:
> How do repeats compare against running more configs in parallel?

They're the same thing. "Repeating the most exploratory bracket two times" is "running two instances of that bracket in parallel". i.e., "2 repeats" means "running 2 copies of the most exploratory bracket in parallel".

https://github.com/dask/dask-ml/pull/532#issuecomment-512470347 is only concerned with the final score, not any of the intermediate scores. I plan to run those simulations somewhat soon.

> What's the x axis on your plot? Number of configurations run?

Whoops!

The plot above is an estimate of the final validation score for running a different number of the most exploratory bracket in parallel. It's a vertical histogram, so the x-axis is the frequency from my simulations.

> And how do you simulate the repeats?

Earlier, I simulated Hyperband 200 times so I also have the results from 200 runs of the most exploratory bracket (shown in "1 repeat"). The other plots (with more repeats) are generated by pulling from that histogram, and taking the maximum of the values pulled. Something like

``` python
def simulate(final_scores, repeats):
    return np.random.choice(final_scores, size=repeats).max()

n = 200  # I ran Hyperband 200 times in an earlier simulation

# For each run, pull the best validation score from bracket=4
one_run = [simulation_results(i, bracket=4) for i in range(n)]  # values in "1 repeat".

two_runs = [simulate(one_run, 2) for _ in range(1000)]
three_runs = [simulate(one_run, 3) for _ in range(1000)]
```


The plot I showed above is only for the final score. It does not characterize how the score changes over time.

amueller:
> They're the same thing. "Repeating the most exploratory bracket two times" is "running two instances of that bracket in parallel". i.e., "2 repeats" means "running 2 copies of the most exploratory bracket in parallel".

Adding more runs is not the same as repeating a bracket I think, because you'd get the best 50% of the overall, not the best 50% from each repeat. So combining them would be less randomized.

Hyperband is as expensive as 4 repeats but it's worse when looking at the histogram, right? It looks even worse than two repeats to me. So doing repeats seems better than hyperband according to your histogram, right?

stsievert:
I'm currently using doing some hyperparameter optimization for a result I've generated. I find myself wanting to quickly get a rough idea of the parameters, and don't want to do an excessive amount of computation. Currently, I find myself running the most exploratory bracket of Hyperband: it's only one bracket and there's not much computation.

Maybe this would be a good API:

* `explore=True`: very exploratory search. Two copies of most aggressive bracket and (in the future) one copy of InverseDecaySearchCV?
* `isinstance(explore, int)`: run `explore` repeats of most exploratory bracket.

I've pushed this API, with a note that `explore` is still experimental.

stsievert:
I think this PR is ready for review now. The implemented changes summarized in https://github.com/dask/dask-ml/pull/532#issuecomment-634731887 come from my desires from the search I'm currently performing. I'm not interested in finding the best parameters, I'm interested in finding good parameters quickly.

I'm currently manually running the most exploratory bracket, or `explore=1`. This PR has the following benefits:

* `explore` has shown empirical performance, especially if `explore in [2, 3]`.* The most exploratory bracket is very likely optimal, and will often out-perform Hyperband if repeated (i.e, if `isinstance(explore, int)`).**
* `explore` allows easy specification of Hyperband parameters.  can specify the two parameters I care about as per the Hyperband rule-of-thumb (training time and number of hyperparameters to sample).
* `explore` effectively limits the total computation (so my EC2 bill isn't too high).

The only backing simulations are in https://github.com/dask/dask-ml/pull/532#issuecomment-512470347. I have not done any simulations to very how cluster usage varies throughout time as the search progresses (which won't be relevant until #676 is resolved) – most of my concerns are more functional (EC2 cost & time).

<sup>*At least for the experiments in https://github.com/dask/dask-ml/pull/532#issuecomment-512470347</sup>
<sup>**See comparison between Hyperband and `bracket s=4` in Figures 5, 6, and 7 the [Hyperband paper][1]. In Figure 3, the second most aggressive bracket of Hyperband does better for a [search on LeNet][2], which has a section titled [tips and tricks][3].</sup>


[3]:http://deeplearning.net/tutorial/lenet.html#choosing-hyperparameters
[2]:http://deeplearning.net/tutorial/lenet.html
[1]:https://arxiv.org/pdf/1603.06560.pdf

dask_ml/model_selection/_hyperband.py
```diff
@@ -106,6 +107,32 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
         optimal. ``aggressiveness=4`` has higher confidence that is likely
         suitable for initial exploration.
 
+    explore : bool, int, default=False
+        If ``explore=True`` is specified, run a custom exploratory search aimed
```
TomAugspurger, line 111: Can you think of a 1-sentance explanation of the purpose of `explore`? Even after reading the description I'm still unsure.

stsievert: Thanks, done. This PR still isn't ready for merge; I have a couple todo's to complete in https://github.com/dask/dask-ml/pull/532#issue-297450726.

stsievert:
This PR is ready for review.

dask_ml/model_selection/_hyperband.py
```diff
@@ -109,6 +113,36 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
         optimal. ``aggressiveness=4`` has higher confidence that is likely
         suitable for initial exploration.
 
+    explore : int, bool, default=False
+        Run a search geared towards an initial exploratory search if this
+        parameter is specified. ``explore`` is typically specified when
+        not much is known about the hyperparameters and/or model.
+
+        If ``explore`` is a bool, run a search aimed at finding the same
```
TomAugspurger, line 121: "a bool" should be "True"?

dask_ml/model_selection/_hyperband.py
```diff
@@ -109,6 +113,36 @@ class HyperbandSearchCV(BaseIncrementalSearchCV):
...
+        validation accuracy as Hyperband with ``explore=False`` but with
+        less computation.
+        If ``explore`` is an integer, repeat the most exploratory bracket
```
TomAugspurger, line 124: What does "exploratory" mean in this context?

dask_ml/model_selection/_hyperband.py
```diff
@@ -384,8 +419,31 @@ def _get_SHAs(self, brackets):
...
+        if self.explore:
+
+            if isinstance(self.explore, bool):
+                n_repeats = max(2, len(SHAs) // 2)
+            elif isinstance(self.explore, int) and self.explore > 0:
+                n_repeats = self.explore
+            elif isinstance(self.explore, int) and self.explore < 0:
+                n_repeats = len(SHAs) + self.explore + 1
+            else:
+                raise ValueError("explore={self.explore} is not an integer")
```
TomAugspurger, line 434: Missing an fstring.

```suggestion
                raise ValueError(f"explore={self.explore} is not an integer")
```

dask_ml/model_selection/_hyperband.py
```diff
@@ -403,10 +461,9 @@ async def _fit(self, X, y, **fit_params):
         )
         SHAs = {b: SHA for b, SHA in zip(_brackets_ids, _SHAs)}
 
-        # This for-loop rename estimator IDs and pulls out wall times
-        key = "bracket={}-{}".format
-        for b, SHA in SHAs.items():
+        def _rename_model_ids(SHA, b, key=lambda x: x):
```
TomAugspurger, line 464: I've we're only using this once, I'd prefer to just do it inline rather than as a closure.

dask_ml/model_selection/_hyperband.py
stsievert: It means the bracket that samples the largest number of models.

stsievert:
Thanks for the comments; resolved. I'd like to see this merged; I think https://github.com/dask/dask-ml/pull/532#issuecomment-512470347 is enough motivation. There's an additional benchmark that needs to be performed to see how the scores increase throughout time, but that's a tangential concern that's not relevant until #677 is merged.

tests/model_selection/test_hyperband.py
```diff
@@ -440,6 +453,76 @@ def test_history(c, s, a, b):
...
+    assert meta[0]["bracket"] == 0
+    n_brackets = {27: 4, 9: 3, 3: 2, 2: 1}
+    assert meta[-1]["bracket"] == n_brackets[max_iter] - 1
+    assert len(meta) == n_brackets[max_iter]
+    if max_iter >= 3:
+        assert len(meta[0]["decisions"]) < len(meta[-1]["decisions"])
+        assert meta[0]["n_models"] < meta[-1]["n_models"]
+
+
+@gen_cluster(client=True, timeout=5000)
```
TomAugspurger, line 514: Can the timeout be removed?

```suggestion
@gen_cluster(client=True)
```

tests/model_selection/test_hyperband.py
```diff
@@ -440,6 +453,76 @@ def test_history(c, s, a, b):
         assert (np.diff(calls) >= 1).all() or len(calls) == 1
 
 
+@pytest.mark.parametrize("explore", [2, 1, -1, -2])
+def test_explore(explore):
+    @gen_cluster(client=True, timeout=5000)
```
TomAugspurger, line 458:
```suggestion
    @gen_cluster(client=True)
```

Issue #550, opened 2019-10-02
Incorrect warning in train_test_split


fonnesbeck:
When applying `train_test_split` to a `float64`-typed dataset, I get an erroneous warning that says there are mixed types:

![image](https://user-images.githubusercontent.com/81476/66060002-8ffe1700-e502-11e9-9642-1316ce7485b7.png)

Using dask_ml version 1.0.0 on a GCS instance.

TomAugspurger:
What are the types of `X` and `y`? both Dask Dataframe / Series, or a mix of dask and pandas?

fonnesbeck:
Both are pandas; `y` was `pop()`ed from `X`.

TomAugspurger:
Ah, in that case we're doing the correct thing (using scikit-learn). We just shouldn't warn.

fonnesbeck:
I've tried switching to torch tensors, but I get the same error:

![image](https://user-images.githubusercontent.com/81476/66068104-93999a00-e512-11e9-8195-6ddf78b5f95c.png)

Moreover, the split fails entirely:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-42-273a132217bf> in <module>
----> 1 X_train, y_train, X_test, y_test = train_test_split(X, y)

/opt/anaconda3/lib/python3.7/site-packages/dask_ml/model_selection/_split.py in train_test_split(test_size, train_size, random_state, shuffle, blockwise, *arrays, **options)
    458             train_size=train_size,
    459             random_state=random_state,
--> 460             shuffle=shuffle
    461         )

/opt/anaconda3/lib/python3.7/site-packages/sklearn/model_selection/_split.py in train_test_split(*arrays, **options)
   2122 
   2123     return list(chain.from_iterable((safe_indexing(a, train),
-> 2124                                      safe_indexing(a, test)) for a in arrays))
   2125 
   2126 

/opt/anaconda3/lib/python3.7/site-packages/sklearn/model_selection/_split.py in <genexpr>(.0)
   2122 
   2123     return list(chain.from_iterable((safe_indexing(a, train),
-> 2124                                      safe_indexing(a, test)) for a in arrays))
   2125 
   2126 

/opt/anaconda3/lib/python3.7/site-packages/sklearn/utils/__init__.py in safe_indexing(X, indices)
    217                                    indices.dtype.kind == 'i'):
    218             # This is often substantially faster than X[indices]
--> 219             return X.take(indices, axis=0)
[...]

PR #552, opened 2019-10-07
Remove warning in train_test_split

TomAugspurger:
It didn't provide guidance to the user anyway.

Closes https://github.com/dask/dask-ml/issues/550


TomAugspurger:
Actually, rethinking this.

I think the behavior I wanted to warn on was a mix of `dask.Array` X  and `dask.Series` y. That falling back to scikit-learn is a poor UX.

So I think we should do is convert the arrays to a dataframe, or the series to an array. Not sure which yet... If we're doing Series -> Array, we'll need chunk sizes, so it'll have to be controlled by a user parameter.

PR #563, opened 2019-10-25
[BUG] Preserve index in PolynomialFeatures

### LINKED ISSUE #562: PolynomialFeatures does not preserve index
When using the `PolynomialFeatures` transformer with `preserve_dataframe=True`, we lose the index of the data frame. 

```python
import pandas as pd
from dask_ml.preprocessing import PolynomialFeatures

df = pd.DataFrame({"x": [1, 2]}, index=[2, 3])
PolynomialFeatures(degree=1, preserve_dataframe=True).fit_transform(df)
#>     1    x
#>0  1.0  1.0
#>1  1.0  2.0
```

The correct output should be
```python
#>     1    x
#>2  1.0  1.0
#>3  1.0  2.0
```

Same problem occurs with dask data frames.

To fix this, I would like to add `index=X.index` to
https://github.com/dask/dask-ml/blob/17568750c85c2cbc3c41b22a4b3498006cd0ed67/dask_ml/preprocessing/data.py#L993
and 
https://github.com/dask/dask-ml/blob/17568750c85c2cbc3c41b22a4b3498006cd0ed67/dask_ml/preprocessing/data.py#L998

Happy to send the PR. Just want to make sure I'm not missing anything or that there's a reason for the current behavior.

jtilly:
Closes #562.

I'm now adding the index back on. 

I also added a test. If there's a more elegant way to reshuffle the index than `sample`, please let me know. 

I also made one of the existing tests stricter by removing a now unnecessary `reset_index=True` (https://github.com/dask/dask-ml/blob/master/tests/preprocessing/test_data.py#L568).

### DIFF
--- a/tests/preprocessing/test_data.py
+++ b/tests/preprocessing/test_data.py
@@ -568,7 +568,7 @@ class TestPolynomialFeatures:
             res_pandas = a.fit_transform(frame.compute())
             assert dask.is_dask_collection(res_df)
             assert dask.is_dask_collection(res_arr)
-            assert_eq_df(res_df.compute(), res_pandas)
+            assert_eq_df(res_df, res_pandas)
         assert_eq_ar(res_df.values, res_c)
         assert_eq_ar(res_df.values, res_arr)
 
@@ -588,6 +588,4 @@ class TestPolynomialFeatures:
         res_df = dpp.PolynomialFeatures(
             preserve_dataframe=True, degree=1
         ).fit_transform(frame)
-        if daskify:
-            res_df = res_df.compute()
         assert_eq_df(res_df.iloc[:, 1:], frame, check_dtype=False)
--- a/tests/preprocessing/test_data.py
+++ b/tests/preprocessing/test_data.py
@@ -1,3 +1,5 @@
+from copy import copy
+
 import dask
 import dask.array as da
 import dask.dataframe as dd
@@ -561,11 +563,12 @@ class TestPolynomialFeatures:
         res_df = a.fit_transform(frame)
         res_arr = b.fit_transform(frame)
         res_c = c.fit_transform(frame)
+
         if daskify:
             res_pandas = a.fit_transform(frame.compute())
             assert dask.is_dask_collection(res_df)
             assert dask.is_dask_collection(res_arr)
-            assert_eq_df(res_df.compute().reset_index(drop=True), res_pandas)
+            assert_eq_df(res_df.compute(), res_pandas)
         assert_eq_ar(res_df.values, res_c)
         assert_eq_ar(res_df.values, res_arr)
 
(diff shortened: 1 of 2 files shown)


tests/preprocessing/test_data.py
```diff
@@ -561,11 +563,12 @@ def test_df_transform(self, daskify):
         res_df = a.fit_transform(frame)
         res_arr = b.fit_transform(frame)
         res_c = c.fit_transform(frame)
+
         if daskify:
             res_pandas = a.fit_transform(frame.compute())
             assert dask.is_dask_collection(res_df)
             assert dask.is_dask_collection(res_arr)
-            assert_eq_df(res_df.compute().reset_index(drop=True), res_pandas)
+            assert_eq_df(res_df.compute(), res_pandas)
```
TomAugspurger: You can skip the compute here. `assert_eq` will handle it for you.

(though I see now that this had `compute` before).

tests/preprocessing/test_data.py
```diff
@@ -575,3 +578,16 @@ def test_transformer_params(self):
...
+        frame = copy(df)
+        if not daskify:
+            frame = frame.compute()
+        frame = frame.sample(frac=1.0)
+        res_df = dpp.PolynomialFeatures(
+            preserve_dataframe=True, degree=1
+        ).fit_transform(frame)
+        if daskify:
+            res_df = res_df.compute()
+        assert_eq_df(res_df.iloc[:, 1:], frame, check_dtype=False)
```
TomAugspurger: Can you explain

1. Why is `check_dtype=False` needed?
2. Why is `.iloc[:, 1:]` needed?

Ideally, we would construct the expected result exactly and compare against that.

jtilly: 1. This is a float64 vs float32 thing. So there might be a bigger underlying problem here.
2. The polynomial transformer with degree 1 adds a column of ones to the left of the input data frame. So I can either drop this column of ones from `res_df` or add it to `frame`. Happy to change that.

jtilly: To clarify 1: the input data frame somehow has one float32 column. Haven't figured out yet where that comes from. Since everything gets transformed into numpy arrays under the hood, the float32 column gets upcasted to float64.

jtilly: The first column of `df` gets downcasted in https://github.com/dask/dask-ml/blob/master/tests/preprocessing/test_data.py#L201.

That's why we're ending up with float32 and float64 dtypes in the tests afterwards. I would be inclined to change that test and ensure that `df` does not get mutated. Passing in `df` as a fixture everywhere would probably be better yet.

As for the `PolynomialTransformer`: Because everything gets turned into a numpy array somewhere in sklearn, it's not really a surprise that we're ending up with homogeneous dtypes. I could re-cast everything (just like we're doing with the index) or add a note/comment somewhere that `preserve_dataframe=True` doesn't mean dtypes are preserved.

TomAugspurger: Thanks for looking into it. SGTM.

TomAugspurger:
Thanks!

PR #580, opened 2019-11-02
ENH: PCA works with NumPy/Pandas data

### LINKED ISSUE #578: Sklearn pipeline and cross_val_score don't work for some transformers
Hi,

```
from sklearn.pipeline import make_pipeline
from sklearn.model_selection import cross_val_score
from sklearn.ensemble import GradientBoostingClassifier
from dask_ml.decomposition import PCA
from dask_ml.wrappers import ParallelPostFit
from dask_ml.preprocessing import StandardScaler

clf = ParallelPostFit(estimator=GradientBoostingClassifier(), scoring='accuracy')

pipe = make_pipeline(PCA(),clf)
pipe = make_pipeline(StandardScaler(),clf)

mysc = cross_val_score(pipe, dataset.iloc[:,:-1], dataset.iloc[:,-1])
```
Work for pipe = make_pipeline(StandardScaler(),clf) but not for pipe = make_pipeline(PCA(),clf)
Got the error:
AttributeError: 'DataFrame' object has no attribute 'chunks'

If I use RobustScaler() instead of StandardScaler():
AttributeError: 'int' object has no attribute 'ndim'

How can I fix this problem?

stsievert:
This PR allows Dask-ML's PCA implementation to work with NumPy arrays and Pandas DataFrames.

This PR closes #578 (well https://github.com/dask/dask-ml/issues/578#issuecomment-548973996).

### DIFF
--- a/dask_ml/decomposition/pca.py
+++ b/dask_ml/decomposition/pca.py
@@ -3,6 +3,7 @@ import numbers
 import dask.array as da
 import dask.dataframe as dd
 import numpy as np
+import pandas as pd
 import scipy.sparse as sp
 import sklearn.decomposition
 from dask import compute
@@ -11,7 +12,7 @@ from sklearn.utils.validation import check_random_state
 
 from .._compat import check_is_fitted
 from .._utils import draw_seed
-from ..utils import svd_flip
+from ..utils import _copy_attrs, svd_flip
 
 
 class PCA(sklearn.decomposition.PCA):
@@ -187,6 +188,20 @@ class PCA(sklearn.decomposition.PCA):
         self.random_state = random_state
 
     def fit(self, X, y=None):
+        if isinstance(X, (np.ndarray, pd.DataFrame)):
+            est = sklearn.decomposition.PCA(
+                n_components=self.n_components,
+                copy=self.copy,
+                whiten=self.whiten,
+                svd_solver=self.svd_solver,
+                tol=self.tol,
+                iterated_power=self.iterated_power,
+                random_state=self.random_state,
+            )
+            est.fit(X)
+            attrs = [k for k in dir(est) if k[0] != "_" and k[-1] == "_"]
+            _copy_attrs(est, self, attrs)
+            return self
         self._fit(X)
         return self
 
@@ -396,6 +411,17 @@ class PCA(sklearn.decomposition.PCA):
 
         """
         # X = check_array(X)
+        if isinstance(X, (np.ndarray, pd.DataFrame)):
+            est = sklearn.decomposition.PCA(
+                n_components=self.n_components,
+                copy=self.copy,
+                whiten=self.whiten,
+                svd_solver=self.svd_solver,
+                tol=self.tol,
+                iterated_power=self.iterated_power,
+                random_state=self.random_state,
+            )
+            return est.fit_transform(X)
         U, S, V = self._fit(X)
         U = U[:, : self.n_components_]
 
--- a/dask_ml/decomposition/pca.py
+++ b/dask_ml/decomposition/pca.py
@@ -192,7 +192,8 @@ class PCA(sklearn.decomposition.PCA):
             msg = (
                 "Got an unsupported type ({}). To resolve this issue,\n\n"
                 "  * Use sklearn.decomposition.PCA  # recommended\n\n"
-                "Wrapping the input with a Dask Array/DataFrame is "
+                "Wrapping the input with a Dask Array/DataFrame will resolve "
+                "this issue but is "
                 "*not recommended* (Dask-ML's PCA implementation will likely "
                 "be slower because the data fit in memory)"
             )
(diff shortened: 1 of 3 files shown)


stsievert:
> In theory, we could pass non-dask objects through to sklearn.decomposition.PCA.

How in theory?

I'm not sure I like the practice of accepting NumPy/DataFrame objects as well as accepting Dask Array/DataFrame objects ([k_means.py#L177][1], [_incremental.py#L469][2], [_encoders.py#L166][3]). I believe that will create a maintenance burden, and I'd rather users choose the right tool. However, having a NEP-18 style interface would be convenient.

[3]:https://github.com/dask/dask-ml/blob/6c3b91cfafdf230408fccc875a749ce67014e89b/dask_ml/preprocessing/_encoders.py#L166
[1]:https://github.com/dask/dask-ml/blob/6c3b91cfafdf230408fccc875a749ce67014e89b/dask_ml/cluster/k_means.py#L177
[2]:https://github.com/dask/dask-ml/blob/6c3b91cfafdf230408fccc875a749ce67014e89b/dask_ml/model_selection/_incremental.py#L469

TomAugspurger:
My primary concern with passing the ndarray through to the sklearn base
class is that we may end up using entirely different algorithms for
PCA(ndarray) vs. PCA(dask.array.Array).

The other option is to wrap the non-dask object into a single-partition
dask object, and then proceed with our algorithm. But that will typically
be slower than the NumPy version.

All in all, perhaps erroring is best.

stsievert:
> perhaps erroring is best

:+1: I raise an error that pushes the user towards using Scikit-learn's implementation. I do mention that wrapping in Dask Array/Dataframe will resolve this but don't recommend that solution.

dask_ml/utils.py
```diff
@@ -360,6 +360,11 @@ def _num_samples(X):
     return result
 
 
+def _copy_attrs(c1, c2, names):
```
TomAugspurger: Is this needed?

tests/test_pca.py
```diff
@@ -780,3 +780,15 @@ def test_unknown_shapes_n_components_larger_than_num_rows(solver):
...
         with pytest.raises(ValueError, match="n_components is too large"):
             pca.fit(X)
+
+
+@pytest.mark.parametrize("input_type", ["array", "dataframe"])
+def test_pca_sklearn_inputs(input_type):
+    a = dd.PCA()
+
+    assert isinstance(X, np.ndarray)
+    _X = X if input_type != "dataframe" else pd.DataFrame(X)
```
TomAugspurger: To avoid this string comparision, I'd recommend using `input_type=[np.array, pd.DataFrame])` above and `X = input_type(X)`

stsievert:
Thanks for the review. Resolved. This LGTM now.

dask_ml/decomposition/pca.py
```diff
@@ -396,6 +408,8 @@ def fit_transform(self, X, y=None):
 
         """
         # X = check_array(X)
+        if isinstance(X, (np.ndarray, pd.DataFrame)):
```
TomAugspurger: Sorry for not seeing this earlier, but can you replace this with `not dask.is_dask_collection(X)`? That will be a bit nicer w.r.t. any duck typing that scikit-learn may do (I don't know if a scipy.sparse matrix is allowed by scikit-learn here. If not, I'd let them raise on that.)

TomAugspurger: Same comment on L200.

stsievert: Thanks. There was actually a test for sparse inputs, and some code to raise an error if sparse inputs passed. I made slight modifications to existing tests:

* `test_pca_sparse_input`: added an argument to make sure the right `TypeError` is raised
* `test_pca_bad_solver`: passed in Dask array instead of NumPy array to make solver-checking code could be reached.

stsievert:
The CI is green, except for the `sklearnDev` tests. One unrelated test is failing. Here's some of the relevant pytest output from Azure:

<details>

``` python
tests/preprocessing/test_data.py::TestRobustScaler::test_df_values FAILED [ 81%]
...
_______________________ TestRobustScaler.test_df_values ________________________

self = <tests.preprocessing.test_data.TestRobustScaler object at 0x7f92f8e7d7f0>
…
        result_ar = est1.fit_transform(X)
        result_df = est2.fit_transform(df)
        if hasattr(result_df, "values"):
            result_df = result_df.values
>       assert_eq_ar(result_ar, result_df)

tests/preprocessing/test_data.py:207: 

a = array([[-0.44793265, -0.66977198, -0.43820941, ...,  0.73827126,
…a_original = dask.array<truediv, shape=(100, 20), dtype=float64, chunksize=(50, 20), chunktype=numpy.ndarray>
b_original = dask.array<values, shape=(nan, 20), dtype=float64, chunksize=(nan, 20), chunktype=numpy.ndarray>
adt = dtype('float64'), a_meta = array([], shape=(0, 0), dtype=float64)
a_computed = array([[-0.44793265, -0.66977198, -0.43820941, ...,  0.73827126,
        -0.9181568 , -0.98776071],
       [ 0.5152352...748, -1.32027257],
       [ 0.05005361, -0.50989146,  1.38200978, ..., -1.34076828,
        -0.2979689 , -0.17230847]])
bdt = dtype('float64'), b_meta = array([], shape=(0, 0), dtype=float64)

def assert_eq(a, b, check_shape=True, check_graph=True, check_meta=True, **kwargs):
…
             assert type(b_meta) == type(b_computed)
>           assert allclose(a, b, **kwargs)
E           AssertionError

/usr/share/miniconda/envs/dask-ml-test/lib/python3.7/site-packages/dask/array/utils.py:267: AssertionError

```


</details>

TomAugspurger:
Agreed that the sklearn-dev failure is unrelated. Thanks.

Issue #585, opened 2019-11-19
dask-searchcv mentioned in dask-ml docs


lebedov:
It appears that dask-searchcv is still explicitly mentioned in some parts of the dask-ml documentation (e.g., [here](https://github.com/dask/dask-ml/blame/master/docs/source/hyper-parameter-search.rst#L57)) despite having been merged.

TomAugspurger:
Thanks for pointing that out. Those should all be updated.

TomAugspurger:
Closed by https://github.com/dask/dask-ml/pull/586

PR #586, opened 2019-11-20
dask-searchcv -> dask-ml in docs (#585)

lebedov:
Replace dask-searchcv in main dask-ml docs (but left appropriate mention in history).


TomAugspurger:
Thanks!

Issue #592, opened 2019-12-10
dask_ml.decomposition.PCA: ValueError with data > 1 TB


demaheim:
When I do `dask_ml.decomposition.PCA().fit(x)`, where the array `x` has a size > 1 TB, I get the error `ValueError: output array is read-only`.

I use
```
dask-ml                   1.1.1
distributed               2.9.0
```

The script
```python
from dask_jobqueue import SLURMCluster
from dask.distributed import Client
from dask_ml.decomposition import PCA
import dask.array as da

cluster = SLURMCluster()
nb_workers = 58
cluster.scale(nb_workers)
client = Client(cluster)
client.wait_for_workers(nb_workers)

x = da.random.random((1000000, 140000), chunks=(100000, 2000))
pca = PCA(n_components=64)
pca.fit(x)
```
gives the error
```
Traceback (most recent call last):
  File "value_error.py", line 48, in <module>
    pca.fit(x)
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/dask_ml/decomposition/pca.py", line 190, in fit
    self._fit(X)
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/dask_ml/decomposition/pca.py", line 338, in _fit
    raise e
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/dask_ml/decomposition/pca.py", line 325, in _fit
    singular_values,
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/dask/base.py", line 436, in compute
    results = schedule(dsk, keys, **kwargs)
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/distributed/client.py", line 2573, in get
    results = self.gather(packed, asynchronous=asynchronous, direct=direct)
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/distributed/client.py", line 1873, in gather
    asynchronous=asynchronous,
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/distributed/client.py", line 768, in sync
    self.loop, func, *args, callback_timeout=callback_timeout, **kwargs
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/distributed/utils.py", line 334, in sync
    raise exc.with_traceback(tb)
  File "/home/dheim/miniconda3/lib/python3.7/site-packages/distributed/utils.py", line 318, in f
    result[0] = yield future
[...]

PR #595, opened 2019-12-23
Fix svd flip

fujiisoup:
cc @demaheim
This adds a workaround for #592 

I'm not sure why but for some reason, the array becomes sometimes read-only and cannot do the inplace multiplication. Large PCA fails because of this issue.
It may not be a perfect fix but it works for large arrays > 1TB.

Please let me know if anyone has better idea to fix this.

### DIFF
--- a/dask_ml/utils.py
+++ b/dask_ml/utils.py
@@ -23,6 +23,43 @@ from ._utils import ConstantFunction
 logger = logging.getLogger()
 
 
+def svd_flip_fixed(u, v, u_based_decision=True):
+    """
+    This is a replicate of svd_flip() from
+    miniconda3/lib/python3.7/site-packages/sklearn/utils/extmath.py
+    To avoid error "ValueError: output array is read-only", it is changed
+    as described by comments in code
+    """
+    if u_based_decision:
+        max_abs_cols = np.argmax(np.abs(u), axis=0)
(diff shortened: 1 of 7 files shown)


dask_ml/utils.py
```diff
@@ -23,8 +23,37 @@
 logger = logging.getLogger()
 
 
+def svd_flip_with_copy(u, v, u_based_decision=True):
```
TomAugspurger: Rather than duplicating the entire implementation, can you copy the inputs and set them to not be readonly?

fujiisoup:
Thank you, @TomAugspurger, for the reply.

Could you instruct me how to construct a read only dask array?
I'll add a test for this case.

TomAugspurger:
Dask Arrays don't really have a "readonly" concept. It'd be the underlying
NumPy array that's readonly.

fujiisoup:
Ok. Then probably I can do this.
Will update in a few days.

fujiisoup:
Added a test.

mrocklin:
I like that when @fujiisoup says "days" he actually means "hours" :)

TomAugspurger:
> I like that when @fujiisoup says "days" he actually means "hours" :)

:)

Thanks! These changes look perfect. The failure against scikit-learn master looks unrelated.

PR #603, opened 2020-01-16
1.2.0 Changelog is missing information updates

datajanko:
I found these two classes not mentioned in the recent changelog, which are available in the docs.


TomAugspurger:
Thanks!

PR #608, opened 2020-02-06
use nanmean and nanvar to comply with sklearn docs

dmh43:
Under the `Notes` section [here](https://ml.dask.org/modules/generated/dask_ml.preprocessing.StandardScaler.html#dask_ml.preprocessing.StandardScaler) it mentions:

> NaNs are treated as missing values: disregarded in fit, and maintained in transform.

To be consistent with these docs I've used `nanmean` and `nanvar` instead of the usual mean and var methods.

### DIFF
--- a/tests/preprocessing/test_data.py
+++ b/tests/preprocessing/test_data.py
@@ -50,17 +50,6 @@ def pandas_df():
 def dask_df(pandas_df):
     return dd.from_pandas(pandas_df, npartitions=5)
 
-@pytest.fixture
-def pandas_nan_df():
-    df = pd.DataFrame(5 * [range(42)]).T.rename(columns=str)
-    df.iloc[0] = np.nan
-
-
-@pytest.fixture
-def dask_nan_df(pandas_df):
-    return dd.from_pandas(pandas_df, npartitions=5)
-
-
 class TestStandardScaler:
     def test_basic(self):
         a = dpp.StandardScaler()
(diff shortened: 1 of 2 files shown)


dmh43:
Any idea why that last test failed? It seems StandardScaler isnt used in that code.

TomAugspurger:
Thanks. I've restarted the failed builds.

Can you add a test with some missing values, and just assert that the statistics aren't NaN?

dmh43:
Thanks for taking a look. Added the test.

tests/preprocessing/test_data.py
```diff
@@ -50,6 +50,16 @@ def pandas_df():
 def dask_df(pandas_df):
     return dd.from_pandas(pandas_df, npartitions=5)
 
+@pytest.fixture
+def pandas_nan_df():
+    df = pd.DataFrame(5 * [range(42)]).T.rename(columns=str)
+    df.iloc[0] = np.nan
```
TomAugspurger: This would need to be returned, right?

tests/preprocessing/test_data.py
```diff
@@ -50,6 +50,16 @@ def pandas_df():
...
+
+
+@pytest.fixture
+def dask_nan_df(pandas_df):
```
TomAugspurger: And this would need to take `pandas_nan_df` to have NA values?

dmh43:
Thanks for taking a look @TomAugspurger, feedback incorporated.

TomAugspurger:
Thanks @dmh43. The CI failures should be fixed on master.

Issue #357, opened 2018-09-07
preprocessing.DummyEncoder.transform() gives confusing error message with unexpected columns


ecurtin2:
When DummyEncoder encounters a dataframe with different columns than the fit() encountered, the error message prints `self.columns`. If no columns value was passed to __init__ this reports an error message like:

```python

ValueError: Columns of 'X' do not match the training columns. Got Index([.......], dtype='object'), expected None
```

If instead the error printed `self.columns_`, which appears to be set from the dataframe passed to fit() it would make the error much more clear. 

Personally, I've added something like this to our local codebase

```python 

expected = set(self.columns_)  
got = set(X.columns)

raise ValueError(
f"Columns of 'X' do not match the training columns.\n"
f"X is missing: {expected - got or 'No columns missing'}. "
f"X has but these were not expected: {got - expected or 'No unexpected columns'}."
)
```

Which makes it clear where the columns differ from expected. I'm more than happy to submit a PR for this, but have never contributed to open source before so this may take me more time than should be necessary for such a small error. 

I'm also unsure if the library has a way to handle a more general input validation across transformers, or if that is even within the scope of dask-ml.

TomAugspurger:
Thanks. I think that's correct, we should be using `self.columns_` instead of `self.columns`.

If you are interested in making a fix, http://ml.dask.org/contributing.html has instructions for getting started. Otherwise, I'll try to get to it next week.

> I'm also unsure if the library has a way to handle a more general input validation across transformers, or if that is even within the scope of dask-ml.

Could you explain that a bit more? Generally, estimators just do whatever validation is required for their own correctness (checking shapes, missing data, etc.).

ecurtin2:
Thanks for getting back! 

By validation I mean that ensuring the columns passed to transform() have the same name and dtype given during fit() (or at least compatible dtype, say int8 vs int32 or something). We recently ran into a bit of an issue where we did pd.read_sql(...) and the query returned the proper columns / names, but one of the inferred dtypes was unexpected since the column in the DB was NULL / dtype unspecified (I think it inferred to string). 

After writing I'm unsure this falls under the scope of the library, but It's something I'm probably gonna have to implement for work anyway.

TomAugspurger:
> By validation I mean that ensuring the columns passed to transform() have the same name and dtype given during fit()

We typically follow scikit-learn here, which I believe will attempt to coerce things (so the int8 vs int32 wouldn't matter).

[...]

PR #610, opened 2020-02-17
Fix DummyEncoder.transform error message

rmsare:
I ran into the issue in #357 while using DummyEncoder recently, so I figured I would contribute a small fix for the print statement.

It also adds a test with `columns=...` specified. Would any other tests be useful here?


TomAugspurger:
Thanks. The CI failure is fixed on master.

PR #614, opened 2020-02-20
Fix LinearRegression score

### LINKED ISSUE #507: model.score() returns MSE but the docstring states that it returns R2
In dask_ml linear_models LinearRegression. If you use model.score(X_test, y_test) then you get back the Mean Squared Error instead of the R2 which is how sklearn's implementation does it. 
On top of that the score method defined has a docstring that states that it returns the r2 score instead of the MSE. 

We can see in the code here https://github.com/dask/dask-ml/blob/e1b21e6dbcfae8341d05d39544ab90dce21fdb2b/dask_ml/linear_model/glm.py#L304-L328

That the docstring says r2 score, but the method returns the mean_squared_error from dask_ml.utils.

Either the docstring should be changed to reflect the real return value. Or the r2_score method should be defined and then use that on the rerturn statement of the score method

exemplary-citizen:
Fixes #507

### DIFF
--- a/ci/environment-3.6.yaml
+++ b/ci/environment-3.6.yaml
@@ -10,6 +10,7 @@ dependencies:
   - distributed ==2.4.0
   - flake8
   - isort
+  - msgpack-python ==0.6.2
   - multipledispatch
   - numba
   - numpy ==1.17.3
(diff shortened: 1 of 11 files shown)


TomAugspurger:
Thanks for this. CI should be fixed if you merge master and re-push.

exemplary-citizen:
@TomAugspurger is this good to go?

TomAugspurger:
yep, thanks!

PR #615, opened 2020-02-21
DOC: typo

raybellwaves:


raybellwaves:
Accidentally did this on the rtd branch. Hitting the 'Edit on GitHub' button in the docs takes you there. I assume #616 is the right way to update the docs.

PR #619, opened 2020-02-25
Incremental pca

### LINKED ISSUE #617: IncrementalPCA
Hi all,

Although dask-ml already has PCA and wrappers.Incremented, but there is no IncrementalPCA.
Is this in the scope of development?

With dask-ml's PCA, we can handle large data, but if the data is too large, some approximated algorithms may be necessary.
With wrappers.Incremental, we could use scikit-learn's IncrementalPCA, but it internally uses scipy.linalg.svd, which won't run with large feature size of data.

For me it would be nicer if dask-ml has dask-compatible IncrementalPCA.

I implemented it in my folk, most part of which is stolen from scikit-learn but uses dask.array.linalg.svd_compressed internally.
If it is in the scope, I will clean it a bit and send a PR.

Thank you!

cc:
@demaheim
@yasahi-hpc

fujiisoup:
closes #617 

Most part of the script is stolen from scikit-learn but updated to be compatible with dask-arrays.
Tests are also copied.


dask_ml/decomposition/extmath.py
```diff
@@ -0,0 +1,104 @@
...
+                last_unnormalized_variance + new_unnormalized_variance +
+                last_over_new_count / updated_sample_count *
+                (last_sum / last_over_new_count - new_sum) ** 2)
+
+        zeros = last_sample_count == 0
+        # updated_unnormalized_variance[zeros] = new_unnormalized_variance[zeros]
+        updated_unnormalized_variance = np.where(
+            zeros, new_unnormalized_variance, 
+            updated_unnormalized_variance)
+        updated_variance = updated_unnormalized_variance / updated_sample_count
```
fujiisoup: This is one of the updated part. Dask does not support assignment and so I use where instead.

dask_ml/decomposition/incremental_pca.py
```diff
@@ -0,0 +1,340 @@
...
+    def __init__(self, n_components=None, whiten=False, copy=True,
+                 batch_size=None, svd_solver='full', 
+                 iterated_power=0, random_state=None):
+        self.n_components = n_components
+        self.whiten = whiten
+        self.copy = copy
+        self.batch_size = batch_size
+        self.svd_solver = svd_solver
+        self.iterated_power = iterated_power
+        self.random_state = random_state
```
fujiisoup: svd_solver, iterated_power, random_state are added as new optional arguments.

dask_ml/decomposition/incremental_pca.py
```diff
@@ -0,0 +1,340 @@
...
+            col_batch_mean = np.mean(X, axis=0)
+            X -= col_batch_mean
+            # Build matrix of combined previous basis and new data
+            mean_correction = \
+                np.sqrt((self.n_samples_seen_ * n_samples) /
+                        n_total_samples) * (self.mean_ - col_batch_mean)
+            X = np.vstack((self.singular_values_.reshape((-1, 1)) *
+                           self.components_, X, mean_correction))
+
+        solver = self._get_solver(X, self.n_components_)
```
fujiisoup: svd_solver is chosen based on the same strategy with PCA.

dask_ml/decomposition/incremental_pca.py
```diff
@@ -0,0 +1,340 @@
...
+        if solver in {"full", "tsqr"}:
+            if hasattr(X, 'rechunk'):
+                X = da.rechunk(X, (X.chunks[0], -1))
```
fujiisoup: If the array is chunked along the last axis, svd won't work.
I'm not sure if rechunking is a right way here.
Probably we can raise an Exception?

tests/test_incremental_pca.py
```diff
@@ -0,0 +1,404 @@
...
+
+iris = datasets.load_iris()
+
+
+@pytest.mark.parametrize('svd_solver', [
+    'full', 
+    'auto', 
+    'randomized'
+])
+def test_compare_with_sklearn(svd_solver):
```
fujiisoup: This is only one test I added aside from those stolen from scikit-learn.

mrocklin:
I apologize for the delay @fujiisoup .  Many of the Dask maintainers are at a workshop this week.  

At first glance I am surprised that the Scikit-Learn code works well with small modifications.  I haven't looked at it deeply yet, but I have a few questions:

1.  Could we use the Scikit-Learn estimator and the general approach for partial_fit?
2.  How does this perform?  In particular, if it is easy for you i would love to see a performance report.  I think that this would help us to talk about bottlenecks that might exist in this algorithm.  https://docs.dask.org/en/latest/diagnostics-distributed.html#capture-diagnostics
3.  I'm curious about when the `da.linalg.svd_compressed` algorithm  in dask.array might be a better fit, particularly with the `compute=True` keyword.


Again, I apologize for the delay, and for not yet giving this a thorough review

dask_ml/decomposition/incremental_pca.py
TomAugspurger: Perhaps check what we do elsewhere in the library. It's not clear to me whether the estimator should implicitly rechunk or not, since this could cause surprisingly high memory usage.

dask_ml/decomposition/incremental_pca.py
```diff
@@ -0,0 +1,340 @@
...
+        U, V = svd_flip(U, V)
+        explained_variance = S ** 2 / (n_total_samples - 1)
+        explained_variance_ratio = S ** 2 / np.sum(col_var * n_total_samples)
+
+        self.n_samples_seen_ = n_total_samples
+        self.components_ = V[:self.n_components_]
+        self.singular_values_ = S[:self.n_components_]
+        self.mean_ = col_mean
+        self.var_ = col_var
+        self.explained_variance_ = explained_variance[:self.n_components_]
```
TomAugspurger: Can you check the types of all these vs. those in `dask_ml.decomposition.PCA`? I notice some that differ.

Which do you find more useful here: a Dask Array or a concrete ndarray?

fujiisoup:
Thank you for the comments ;) and sorry for my late reply.

We tested this PR with our real data and found some problems.
Now we are fixing them.
We are also preparing some benchmarking of our PR.

I'll push and post the update (probably later this week or the next week.)
@yasahi-hpc

dask_ml/decomposition/extmath.py
TomAugspurger: The line commented above is what we would need to support in dask? I wonder if we could get that implemented in Dask.

fujiisoup: > The line commented above is what we would need to support in dask?
Yes. If I understand correctly dask-array does not support the item assignment. If it is supported, it would be great of course.

fujiisoup:
Thanks. I marked the updated part from scikit-learn by comment. 
We will make sure its performance, and report it back!

yasahi-hpc:
Hello, I have made a performance analysis with the performance_report and a naive timer. The testbed is a [Intel Skylake CPU](https://ark.intel.com/content/www/us/en/ark/products/120489/intel-xeon-gold-6148-processor-27-5m-cache-2-40-ghz.html). Using the incremental PCA to our large data ~20GB, it gave the [dask-report](https://gistcdn.githack.com/yasahi-hpc/7bc274e4cd0122ed8cf55379c1ae82f2/raw/aacf28810fefd512fae3d2aca054365ed619cf98/dask-report.html) for the fit method . For some reason, if we call the method under performance_report, it gets 50 times slower. 

A naive timer gave 184s, 148s and 63s for incremental_mean_and_var, svd, and compute(), respectively. For the naive measurement, we deliberately compute() all the quantities for synchronization. Other functions are immediately finished.

mrocklin:
> For some reason, if we call the method under performance_report, it gets 50 times slower.

Hrm, that is concerning.  Thank you for reporting this.  I don't immediately see why.  I raised https://github.com/dask/distributed/pull/3654 to make sure that we are not counting the time taken to generate the plots.  

However after I looked more closely at your performance report I see that this is not the problem.  I see that the computations are actually taking many hours, which is surprising. There are tasks like `mean_agg-aggregate` and `getitem` which take 60 seconds, which is very surprising.  Computing means and getting slices should be very fast.  I do not know why these would be slow, or why running your code with the `performance_report` context manager would make things slow.  

The performance report code doesn't run during execution, it only collects data afterwards.  

When I look at the administrative profiles, which observe Dask itself (not the numpy/xarray code) I don't see anything that would cause this.  I do however see that it is taking some time to read excess data from disk (you are maybe running out of memory?)  But this adds only a few minutes of total execution time across all workers.

In summary, I am confused :/

TomAugspurger:
@fujiisoup apologies for the CI failures. If you merge master they should be resolved.

tests/test_incremental_pca.py
```diff
@@ -0,0 +1,443 @@
+"""Tests for Incremental PCA."""
+import numpy as np
+import pytest
+from dask import array as da
+
+from sklearn.utils._testing import assert_almost_equal
+from sklearn.utils._testing import assert_array_almost_equal
+from sklearn.utils._testing import assert_allclose_dense_sparse
```
TomAugspurger, line 8: Seems to be causing an import error. Can these be avoided?

yasahi-hpc:
@mrocklin 
Thanks so much for your detailed explanation and help.
I am also sorry for my late reply. Our cluster has been unavailable for a week just until today.
I guess some of the issues stem from the Lustre file system of our cluster.
(Still not very clear, why the problem occurs only with the `performance_report`.)
The memory shortage may be a thing since there is a warning message implying a memory leak (the message appears when creating a `Client` object).

In order to focus on the performance of Dask itself, I have simplified the script with the smaller input data size (~7.5GB).
I no longer use xarray/numpy to read our actual input data, rather I use the mimic data generated by `dask.random.random`. In addition, it turned out that inserting the line `dask.config.set({'temporary_directory': './tmp'}) ` to avoid creating the scratch directory gives a significant performance improvement. 
In the end, the script runs three times slower than without `performance_report`.

The [new performance report](https://gistcdn.githack.com/yasahi-hpc/1fece3e9f9e1f6f813ddc4d171233c1d/raw/d05eab0a8fdfa07ef4293f680f318d9efbc6431b/dask-report-ipca.html) seems to show more reasonable performance. Unfortunately, I am not sure why adding the `performance_report` changes the performance.

tests/test_incremental_pca.py
fujiisoup: Just noticed that older scikit-learn does not have `utils._testing` but instead `utils.testing`.
Added a patch by `try: except ImportError`.

dask_ml/decomposition/incremental_pca.py
```diff
@@ -0,0 +1,407 @@
...
+
+        # This is the first partial_fit
+        if not hasattr(self, "n_samples_seen_"):
+            self.n_samples_seen_ = 0
+            self.mean_ = 0.0
+            self.var_ = 0.0
+
+        # Update stats - they are 0 if this is the first step
+        # The next line is equivalent with np.repeat(self.n_samples_seen_, X.shape[1]),
+        # which dask-array does not support
```
TomAugspurger, line 292: Dask does support `np.repeat`. Is it just when `repeats=0` that there's an issue? Opened https://github.com/dask/dask/issues/6076 for that.

PR #622, opened 2020-03-04
Make GridSearchCV reuse scattered data.

### LINKED ISSUE #516: GridsearchCV with prescattered data
During a recent debug session with @TomAugspurger we thought that dask-ml could take advantage of data which was pre-scattered:

> client.scatter(data, broadcast=True)

Current understanding is that dask-ml will not take advantage of pre-scatter data as the call to `KFold` will chop up data and not see the cached data on the workers.

lesteve:
Fix #516.

Maybe there is a better test? Also could use `type(x).__name__` (rather than hardcoded `ndarray`) as in `client.scatter`: https://github.com/dask/distributed/blob/9b5bf448af478a166069aefc9c0c1354a29ae482/distributed/client.py#L1930

### DIFF
--- a/dask_ml/model_selection/utils.py
+++ b/dask_ml/model_selection/utils.py
@@ -85,7 +85,7 @@ def to_keys(dsk, *args):
             yield x.key
         else:
             assert not is_dask_collection(x)
-            key = "ndarray-" + tokenize(x)
+            key = type(x).__name__ + "-" + tokenize(x)
             dsk[key] = x
             yield key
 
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks. I think you're missing a `.utils` in your test.

I think what you have (directly testing to_keys against client.scatter) is about as good as we could do. I never came up with an easy way to test it using GridSearchCV itself.

w.r.t. using `type(x).__name__`, that might be a bit better. We *might* be able to get a pandas DataFrame here.

lesteve:
Done. I also test pandas Dataframe now.

TomAugspurger:
Thanks!

Issue #607, opened 2020-02-01
Add mild type annotations


mrocklin:
I'm looking through the model selection code and I find myself curious what the various inputs and outputs are to functions.  This might be a bit of our codebase that could use with some mild type annotations.  

This might be a good activity for someone who wanted to become familiar with the Dask-ML codebase.

dhirschfeld:
I find type annotations hugely helpful to understand the intended interface/behaviour of a function

TomAugspurger:
Sounds good to me.

prasunanand:
Hi, 
I would like to work on this issue.

jakirkham:
Please feel free. My guess is people here would be happy to help review a PR 🙂

prasunanand:
How should I declare the type signature of `array-like` ?

TomAugspurger:
I'd follow pandas' definitions:
https://github.com/pandas-dev/pandas/blob/master/pandas/_typing.py

with the addition of Dask Arrays / DataFrames where appropriate.

prasunanand:
Thanks @TomAugspurger :)

PR #626, opened 2020-03-18
Annotations

prasunanand:
* Run `black` and `isort`
* Add annotations

### DIFF
--- a/benchmarks/kmeans_airline.py
+++ b/benchmarks/kmeans_airline.py
@@ -91,7 +91,7 @@ def main(args=None):
 
     logger.info("Reading data")
     X = read().pipe(transform).pipe(as_array)
-    X, = persist(X)
+    (X,) = persist(X)
 
     timings = []
 
--- a/dask_ml/_typing.py
+++ b/dask_ml/_typing.py
@@ -1,0 +1,79 @@
+# Borrowed from https://github.com/pandas-dev/pandas/blob/master/pandas/_typing.py
+
+from pathlib import Path
+from typing import (
+    IO,
+    TYPE_CHECKING,
+    Any,
+    AnyStr,
+    Callable,
+    Collection,
+    Dict,
+    Hashable,
+    List,
+    Mapping,
+    Optional,
+    TypeVar,
+    Union,
+)
+
+import numpy as np
+
+# To prevent import cycles place any internal imports in the branch below
+# and use a string literal forward reference to it in subsequent types
+# https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles
+if TYPE_CHECKING:
+    from pandas._libs import Period, Timedelta, Timestamp  # noqa: F401
+    from pandas.core.arrays.base import ExtensionArray  # noqa: F401
+    from pandas.core.dtypes.dtypes import ExtensionDtype  # noqa: F401
+    from pandas.core.indexes.base import Index  # noqa: F401
+    from pandas.core.generic import NDFrame  # noqa: F401
+    from pandas import Interval  # noqa: F401
+    from pandas.core.series import Series  # noqa: F401
+    from pandas.core.frame import DataFrame  # noqa: F401
+    from dask.array import Array
+
+# array-like
+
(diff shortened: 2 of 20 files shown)


TomAugspurger:
Thanks. I think we should start with getting this running on CI and gradually add types one module or so at a time. That'll make reviewing things much easier.

mrocklin:
Thanks for doing this @prasunanand .  it's interesting to see.

> Thanks. I think we should start with getting this running on CI and gradually add types one module or so at a time. That'll make reviewing things much easier.

@TomAugspurger are you suggesting turning on `mypy`?  My experience trying this before was that it wasn't trivial to turn on `mypy` for incremental testing.  It was pretty vocal about things failing, and, at least for me at the time, it wasn't clear how to get it to be useful with a mildly annotated codebase.

More generally, now that we see some basic typing in this codebase, how do you feel about it?  Useful?  A distraction?  Is this something that we want to encourage more of?

TomAugspurger:
It's possible to only use mypy on specific files. Pandas is going through
that now. You can see the configuration starting at
https://github.com/pandas-dev/pandas/blob/dbd7a5d3e2f1d196e8634c620fc72db1127de157/setup.cfg#L124.
I think the current effort is around enabling `check_untyped_defs`
everywhere.

I think most of the value is in better completion through your text editor,
and secondarily through static check (though the static checking can be
annoying when mypy is "wrong").
So I'd prefer to do it as part of CI, so that we can actually rely on the
annotations. Just spat checking, there are a few annotations that aren't
quite right here (e.g. `centers` in `make_blobs` should be something like
`Union[int, ndarray]`, rather than just int.

prasunanand:
Working on it @TomAugspurger

prasunanand:
Should I create new PRs wrt each module ?

mrocklin:
I think that it's ok to include everything here if that's easier for you

benchmarks/kmeans_airline.py
```diff
@@ -91,7 +91,7 @@ def main(args=None):
 
     logger.info("Reading data")
     X = read().pipe(transform).pipe(as_array)
-    X, = persist(X)
+    (X,) = persist(X)
```
TomAugspurger: Was mypy compaining about this?

dask_ml/_typing.py
```diff
@@ -0,0 +1,79 @@
+# Borrowed from https://github.com/pandas-dev/pandas/blob/master/pandas/_typing.py
```
TomAugspurger: I'd prefer to take only what we need (e.g. ArrayLike).

STuff like pandas._libs, pandas.core is private to pandas, and so might break our CI without warning.

dask_ml/cluster/k_means.py
```diff
@@ -22,6 +21,9 @@
 from ..utils import _timed, _timer, check_array, row_norms
 from ._compat import _k_init
 
+import numba  # isort:skip (see https://github.com/dask/dask-ml/pull/577)
```
TomAugspurger: The import order here was apparently important. Do you know why it was moved?

prasunanand: Yeah. I will remove this change.

dask_ml/_typing.py
prasunanand: Ok.  I will correct it.

benchmarks/kmeans_airline.py
prasunanand: `black` formatting

prasunanand:
@mrocklin  @TomAugspurger  I too think that doing this one module at a time would be easier. Will open a new PR.

stsievert:
Thanks for this PR @prasunanand! :+1:

> I too think that doing this one module at a time would be easier. Will open a new PR.

Personally, I'd like to see `dask_ml.model_selection.{_incremental, _hyperband}` to be type checked. I'd make an educated guess this is where the confusion in https://github.com/dask/dask-ml/issues/607#issue-558572736 came from; `_hyperband._get_SHAs` is especially confusing.

PR #627, opened 2020-03-22
DOC: update year

raybellwaves:
update year at the bottom of https://ml.dask.org/index.html

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -70,7 +70,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "dask-ml"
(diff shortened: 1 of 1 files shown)


stsievert:
I don't think this should be merged. The Dask developers published this document in 2017 and have a copyright claim on that material. What if someone copied our content in 2019, and Dask sued them in court?

This is the topic of an SO question: https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated/2391555#2391555

raybellwaves:
Interesting. Closing to save space.

Issue #535, opened 2019-08-09
No support for stratified split in dask_ml.model_selection.train_test_split


chauhankaranraj:
scikit-learn implementation of train test split (sklearn.model_selection.train_test_split) supports splitting data according to class labels (stratified split) by using the argument `stratify`. This is especially useful when datasets have high class imbalance. It would be really helpful to have this feature in dask_ml as well.

TomAugspurger:
Agreed. Are you interested in working on this?

chauhankaranraj:
Tempted to say yes, but I don't know the codebase/internals very well (specifically, I'm not sure how we can get a stratified split with blockwise=False not implemented for the ShuffleSplit class).

So it'd be faster if someone more knowledgeable could volunteer. If not then I'd be happy to give it a shot, but it might take some time.

TomAugspurger:
That's great if you're willing to try. Let us know if you get stuck.

tiagofassoni:
Hey, Tom, I'm thinking of picking this up. My doubt is:

Say we have a big csv file with 2 categories and two partitions of the data. 

So `file_0 has only category 0 and 1`, `file_1 has only category 1`.

My first thought was to just use the stratify parameter of scikit-learn, but in this case that wouldn't work. Another idea would be to compute all the categories beforehand and pass those to the stratify parameter, but seems overly complicated and prone to a ton of edge cases. 

I'd be glad to pick this up, as it would help in some research I'm doing.

TomAugspurger:
@tiagofassoni great! dask-ml's OneHotEncoder may be helpful here. It will use the Categorical dtype for pandas dataframes. Otherwise you can (or maybe need?) to pass the `categories` manually as a list / array. Does that make sense?

In other places that just work with arrays, like `Incremental`, we require that the `classes` (groups in this case) be specified ahead of time.

jerrytim:
is there any luck with this feature request? in the case of huge imbalanced dataset, the stratify argument in train_test_split is useful

TomAugspurger:
I’m not aware of any progress. Perhaps Tiago can share a status update.

tiagofassoni:
Hello, @TomAugspurger, @jerrytim. Got to try my hand on this just last week and... gotta say, I have no idea on how to make it. I don't know why OneHotEncoder would be helpful, if at all.

I was thinking of using something like Pandas' value_counts for the series results and then trying to make a shuffle, but I don't know if such an approach is feasible.

[...]

PR #635, opened 2020-04-03
WIP: Add stratified split feature to model_selection.train_test_split

chauhankaranraj:
I took a stab at implementing a solution for issue #535 

Adding a WIP label because currently the stratified split is not completely lazily for dask arrays (`compute_chunk_sizes` being called [here](https://github.com/chauhankaranraj/dask-ml/blob/master/dask_ml/model_selection/_split.py#L580)). Nonetheless, I think it works fine for dask series and dataframes.

Any feedback would be appreciated :)

### DIFF
--- a/dask_ml/model_selection/_split.py
+++ b/dask_ml/model_selection/_split.py
@@ -9,6 +9,7 @@ import dask
 import dask.array as da
 import dask.dataframe as dd
 import numpy as np
+import pandas as pd
 import sklearn.model_selection as ms
 from sklearn.model_selection._split import BaseCrossValidator, _validate_shuffle_split
 from sklearn.utils import check_random_state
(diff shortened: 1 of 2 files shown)


dask_ml/model_selection/_split.py
```diff
@@ -377,6 +379,15 @@ def train_test_split(
         :func:`sklearn.model_selection.train_test_split`.
     test_size : float or int, default 0.1
     train_size : float or int, optional
+    stratify : Dask Array or Series, optional (default=None)
+        If all *arrays are non-dask objects, stratify will be passed through
```
TomAugspurger, line 383: Does `*arrays` need to be inside backticks for this to render correctly?

dask_ml/model_selection/_split.py
```diff
@@ -502,12 +559,44 @@ def train_test_split(
...
-            (_blockwise_slice(arr, train_idx), _blockwise_slice(arr, test_idx))
-            for arr in arrays
-        )
+        if stratify is not None:
+            # convert to da.Array
+            if not isinstance(stratify, da.Array):
+                stratify = stratify.to_dask_array(lengths=True)
+
+            # must be 1-d for indexing
+            stratify = stratify.ravel()
```
TomAugspurger, line 569: This seems a bit dangerous. Can you at least assert that the `.size` doesn't change?

dask_ml/model_selection/_split.py
```diff
@@ -502,12 +559,44 @@ def train_test_split(
...
-        train_idx, test_idx = next(splitter.split(*arrays))
 
-        train_test_pairs = (
-            (_blockwise_slice(arr, train_idx), _blockwise_slice(arr, test_idx))
-            for arr in arrays
-        )
+        if stratify is not None:
+            # convert to da.Array
+            if not isinstance(stratify, da.Array):
+                stratify = stratify.to_dask_array(lengths=True)
```
TomAugspurger, line 566: This will make an additional pass over the data. Ideally we could combine this pass with the pass discovering the frequencies.

dask_ml/model_selection/_split.py
```diff
@@ -502,12 +559,44 @@ def train_test_split(
...
+            train_test_pairs = []
+            for arr in arrays:
+                # list of class-wise train/test split array slices
+                arr_train_slices = []
+                arr_test_slices = []
+                for ci in classes:
+                    # get subarray of data from this class
+                    ci_arr = arr[stratify == ci]
+                    # FIXME: can chunks be determined lazily?
+                    ci_arr.compute_chunk_sizes()
```
TomAugspurger, line 580: We'll want to rethink this a bit... `arrays` will likely have a common dependency structure (e.g. they may be read from disk at the same time). So we'll want to do as much as possible in the same call to compute. Right now it looks like we have a `compute` per array.

TomAugspurger (commented):
Can you say a bit about the high-level strategy here, and the challenges?

Stepping back, the idea behind `stratify` is to get approximately the same frequency of each class in the output splits as in the input? So do we absolutely require computing? At the very least, I think we'll need a full pass over the data to compute the frequencies in `stratify`, since the data may not be shuffled ahead of time. But can that pass be delayed until `.compute` time rather than when we construct the graph?

austinzh:
My two cents.
1. classes can be optional because computing classes from an out-of-core dataset, outside train test split will cost the same.
2. If we split classes by classes, does it mean the return train, test datasets are ordered by classes?

TomAugspurger:
I don't think we would want to order by class. Does scikit-learn do that?

austinzh:
Yes. But if we check `for ci in classes:` loop, we will found that we split class by class then concatenate them back.
That implies the return array, for example, the train set looks like 
[randomlized_classA, randomlized_classB, randomlized_ClassC] meaning in this PR's implementation, same class stick together.

But If we use the same parameter for scikit-learn's train_test_split, the output will be shuffled.

dask_ml/model_selection/_split.py
```diff
@@ -478,12 +511,36 @@ def train_test_split(
...
+                    )
+
+                    # add subdf's from this class to list of all subdf's
+                    arr_train_slices.append(arr_train)
+                    arr_test_slices.append(arr_test)
+
+                # concat all train subdfs as 1 train df, same for test
+                train_test_pairs.append(
+                    [dd.concat(arr_train_slices), dd.concat(arr_test_slices)]
+                )
```
austinzh, line 537: For the output is not ordered by classes. I think we needs to add some kind of shuffle here.
```suggestion
                train = dd.concat(arr_train_slices)
                test = dd.concat(arr_test_slices)
                train = train.shuffle(train.index)
                test = test.shuffle(test.index)
                # concat all train subdfs as 1 train df, same for test
                train_test_pairs.append([train, test])
```

dask_ml/model_selection/_split.py
```diff
@@ -433,6 +444,22 @@ def train_test_split(
 
     types = set(type(arr) for arr in arrays)
 
+    if stratify is not None:
```
hsteinshiromoto, line 447: Quick question: why you are not using `if stratify:` ?

hsteinshiromoto (commented):
Hi,

I am a new guy and I had a question / suggestion to improve the code.

Best,

dask_ml/model_selection/_split.py
roye10: Could an alternative be to force dask data-frames/series? This could alleviate the problem of the additional pass and avoid this additional `compute` per array

bpsut:
What is the current status of this? It's 2026 and this appears to still be open, but it would be REALLY nice to have.

chauhankaranraj:
Hi dask community,

Apologies for being MIA, had a busy stretch at my day job (we were launching a new AWS product, S3 Files). Finally got some time to work on this PR this weekend with some LLM help. Would really appreciate a re-review whenever you have cycles! :pray:

> Can you say a bit about the high-level strategy here

The high level approach is as follows
1. Count how many rows of each class live in each block
2. Sum these counts to get the global class distribution. Use `test_size` to decide how many test rows each class should contribute overall (keeping in mind sklearn's rule of at least one row of every class in both train and test).
3. For each class, split its "test rows budget" across blocks in proportion to how many rows of that class each block holds.
4. Each block picks that many rows at random per class. Slice every input array by those indices and concatenate the pieces.

Note that everything stays lazy until `.compute()`. Only the small `(n_blocks, n_classes)` shaped row count matrix is computed and brought into memory, which to me seems like a fair trade-off for the split accuracy.

Please lmk what y'all think!

dask_ml/model_selection/_split.py
chauhankaranraj: Good catch, fixed in the latest commit

dask_ml/model_selection/_split.py
chauhankaranraj: Updated code only computes class frequencies from the `stratify` array, no compute happens for the others

dask_ml/model_selection/_split.py
chauhankaranraj: Removed lengths=True path, should not be applicable anymore, but lmk if I missed something!

Issue #317, opened 2018-07-27
Samplers / pipelines for imbalanced datasets


TomAugspurger:
See if any of these http://contrib.scikit-learn.org/imbalanced-learn/stable/api.html can be scaled.

sephib:
Hi,
I think that most of the changes would be to introduce the option of `fit_resample` and `fit_sample` into the `fit_transform` method.  
I'll be happy to assist on this issue.

TomAugspurger:
@sephib do you have any examples of fit_resample and fit_sample? I'm not familiar with them.

sephib:
The core [fit_resample](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/6b3c5ae/imblearn/base.py#L54) function is from within [imblearn/base.py](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/6b3c5ae/imblearn/base.py).  
 It is incorporated throughout the `imblearn` library - for example here is the implementation within [imblearn pipeline](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/6b3c5aed61f2e5dc0e8af87d97ea92b95dcafdd0/imblearn/pipeline.py#L333)

TomAugspurger:
Thanks. The standard `sklearn.pipeline.Pipeline` works well with dask
containers. Does the one in imblearn work with Dask objects? If not, what
breaks?

sephib:
Currently when I ran `daskml` with an `imblearn pipeline`  I got an error:  
> AttributeError: 'FunctionSamplerw object has no attribute 'transform'     

This is from the [dask_ml/model_selection/method.py fit_transform](https://github.com/dask/dask-ml/blob/master/dask_ml/model_selection/methods.py#L263)  function which is  looking for a `fit_transform` or `fit` and `transform` attributes (which are in imblearn "converted" to `fit_resample`

TomAugspurger:
It'd would help to have a minimal minimal example: http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports

sephib:
Hi
[...]

PR #638, opened 2020-04-10
Add support for imblearn's Pipeline and Samplers.

xKHUNx:
I'm using dask-ml to parallelized my sklearn pipeline, which consist of imblearn pipeline. I hack the code to make it work with it.

This would be extremely useful for people that are using both libraries.

### DIFF
--- a/dask_ml/model_selection/_search.py
+++ b/dask_ml/model_selection/_search.py
@@ -749,7 +749,7 @@ def _do_fit_step(
                 n_splits,
                 error_score,
             )
-    return (fits, Xs, ys) if is_transform else (fits, None, None)
+    return (fits, Xs, ys) if is_transform else (fits, None, ys)
 
 
 def _do_pipeline(
--- a/dask_ml/model_selection/_search.py
+++ b/dask_ml/model_selection/_search.py
@@ -330,7 +330,7 @@ def do_fit_and_score(
     scorer,
     return_train_score,
 ):
-    if not issubclass(type(est), Pipeline):
+    if not isinstance(est, Pipeline):
         # Fitting and scoring can all be done as a single task
         n_and_fit_params = _get_fit_params(cv, fit_params, n_splits)
 
@@ -434,7 +434,7 @@ def do_fit(
     n_splits,
     error_score,
 ):
-    if issubclass(type(est), Pipeline) and params is not None:
+    if isinstance(est, Pipeline) and params is not None:
         return _do_pipeline(
             dsk,
             next_token,
@@ -504,7 +504,7 @@ def do_fit_transform(
     n_splits,
     error_score,
 ):
-    if issubclass(type(est), Pipeline) and params is not None:
+    if isinstance(est, Pipeline) and params is not None:
         return _do_pipeline(
             dsk,
             next_token,
--- a/dask_ml/model_selection/_search.py
+++ b/dask_ml/model_selection/_search.py
@@ -50,6 +50,7 @@ from .methods import (
     cv_split,
     feature_union,
     feature_union_concat,
+    feature_union_y,
     fit,
     fit_and_score,
     fit_best,
@@ -951,10 +952,13 @@ def _do_featureunion(
                     nsamp + (n,),
                     wl,
                 )
-                dsk[(yt_name, m, n)] = ys[0]
+                dsk[(yt_name, m, n)] = (
+                    feature_union_y, 
+                    ys[0]
+                )
             seen[steps, wt] = m
             out_append(m)
-            m += 1
+            m += 1,
     return [(fit_name, i) for i in out], [(tr_name, i) for i in out], [(yt_name, i) for i in out]
 
 
--- a/dask_ml/model_selection/_search.py
+++ b/dask_ml/model_selection/_search.py
@@ -951,12 +951,7 @@ def _do_featureunion(
                     nsamp + (n,),
                     wl,
                 )
-                dsk[(yt_name, m, n)] = (
-                    feature_union_concat,
-                    [None if y is None else y + (n,) for y in ys],
-                    nsamp + (n,),
-                    wl,
-                )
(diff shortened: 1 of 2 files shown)


dask_ml/model_selection/_search.py
```diff
@@ -326,7 +330,7 @@ def do_fit_and_score(
     scorer,
     return_train_score,
 ):
-    if not isinstance(est, Pipeline):
```
TomAugspurger, line 329: Does `imblearn.pipeline.Pipeline` not subclass `sklearn.pipeline.Pipeline`?

glemaitre:
> cc @glemaitre, if you have any thoughts on how this should be supported.

Basically, we just inherit from `Pipeline` and allow to call `fit_resample` during `fit` of the `Pipeline`.
Actually, it would be interesting to have some supported resamplers in dask: random under- and over-sampling. The rest of the sampler would not make so much sense with large dataset and not sure that you can generate sample without a full view on the dataset. However, random sampler could be handy to resample on each node maybe.

dask_ml/model_selection/_search.py
```diff
@@ -30,7 +30,11 @@
     _BaseKFold,
     _CVIterableWrapper,
 )
-from sklearn.pipeline import FeatureUnion, Pipeline
+from sklearn.pipeline import FeatureUnion
+try:
+    from imblearn.pipeline import Pipeline
+except:
+    from sklearn.pipeline import Pipeline
```
TomAugspurger, line 37: Why is this needed? Can we always use `sklearn.pipeline.Pipeline` and check for `hasattr(x, "fit_resample")`?

xKHUNx: > @xKHUNx can you summarize the high-level changes? I'm a little concerned that this is affecting so many small places? For example, why does the return type of `fit_transform` need to be changed? Can that be limited to a new `fit_reshape` method that returns the transformed `yt` as well?

I assume you meant  `fit_resample`, correct me if I am wrong. My high level idea is that `yt` needs to be propagated throughout the pipeline as component that implements `fit_resample` might change the `y` values. The code works for my personal use case, where I perform RandomizedCVSearch with imblearn component in my pipeline. I can't post my code here as it is related to my work.

> Why is this needed? Can we always use `sklearn.pipeline.Pipeline` and check for `hasattr(x, "fit_resample")`?

This is needed when the refit option is selected. If I understand correctly, when the refit option is selected, dak-ml will fit the pipeline normally, without going through each compoennt in the pipeline. In this scenario, if a normal sklearn pipeline is used, it won't handle pipeline component that implements `fit_resample` instead of `fit_transform`. On the other hand, if imblearn's pipeline is used, it would behave as it should.

stsievert:
Would this PR close #317? It seems certainly is related.

TomAugspurger:
@xKHUNx can you include a test with these changes (will need to include imblearn in the environment file)? It'd be helpful to have an example to play with.

sephib:
> Would this PR close #317? It seems certainly is related.

Yes

TomAugspurger:
@xKHUNx are you able to provide a simple test / example using this. That'd help me with reviewing.

dask_ml/model_selection/_search.py
glemaitre: We do.

TomAugspurger:
@xKHUNx are you able to add tests here?

xKHUNx:
I have not test it myself, but this should do:

```import pandas as pd
import numpy as np
from sklearn.compose import ColumnTransformer
from imblearn.pipeline import Pipeline
from sklearn.impute import SimpleImputer
from sklearn.preprocessing import StandardScaler, OneHotEncoder
from sklearn.linear_model import LogisticRegression
from sklearn.model_selection import train_test_split, GridSearchCV
from imblearn.over_sampling import SMOTE
from dask_ml.model_selection import RandomizedSearchCV

np.random.seed(0)

# Read data from Titanic dataset.
titanic_url = ('https://raw.githubusercontent.com/amueller/'
               'scipy-2017-sklearn/091d371/notebooks/datasets/titanic3.csv')
data = pd.read_csv(titanic_url)

# We create the preprocessing pipelines for both numeric and categorical data.
numeric_features = ['age', 'fare']
numeric_transformer = Pipeline(steps=[
    ('imputer', SimpleImputer(strategy='median')),
    ('scaler', StandardScaler())])

categorical_features = ['embarked', 'sex', 'pclass']
categorical_transformer = Pipeline(steps=[
    ('imputer', SimpleImputer(strategy='constant', fill_value='missing')),
    ('onehot', OneHotEncoder(handle_unknown='ignore'))])

preprocessor = ColumnTransformer(
    transformers=[
        ('num', numeric_transformer, numeric_features),
        ('cat', categorical_transformer, categorical_features)])

# Append classifier to preprocessing pipeline.
# Now we have a full prediction pipeline.
clf = Pipeline(steps=[('preprocessor', preprocessor),
                      ('smote', SMOTE(sampling_strategy='all')),
                      ('classifier', LogisticRegression(solver='lbfgs'))])

X = data.drop('survived', axis=1)
y = data['survived']

params = {
    'classifier__C': loguniform(0.001, 1)
}

X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2)

final_clf = RandomizedSearchCV(final_clf, n_iter=10, param_distributions=params, cv=3, n_jobs=-1, refit=True)
print("model score: %.3f" % clf.score(X_test, y_test))
```

TomAugspurger:
I won't have time, unfortunately. Are you interested in working on this PR still?

xKHUNx:
> Are you interested in working on this PR still?

This PR is good enough for my personal use case, I don't see the need to improve it for the near future.

TomAugspurger:
OK. We would need to add a tests, clean up a few things, and clarify a couple changes before this will be merged. Let me know if you want to pick it up again.

PR #640, opened 2020-04-12
Added annotations to preprocessing

prasunanand:
* Added annotations
* Needed to add --no-strict-optional because of  this [issue]( https://www.pythonsheets.com/notes/python-typing.html#avoid-none-access)


ci/code_checks.sh
```diff
@@ -15,6 +15,7 @@ RET=$(($RET + $?)) ; echo $MSG "DONE"
 
 MSG='Checking mypy... ' ; echo $MSG
 mypy dask_ml/metrics
+mypy --no-strict-optional dask_ml/preprocessing
```
TomAugspurger, line 18: Ideally we do these all at once. I think you can pass multiple paths.

Why do you need no-strict-optional? mypy recommends against disabling strict-optional: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#disabling-strict-optional-checking

dask_ml/preprocessing/_encoders.py
```diff
@@ -137,13 +140,13 @@ def __init__(
         super(OneHotEncoder, self).__init__(**signature)
 
     @classmethod
-    def _get_param_names(cls):
+    def _get_param_names(cls: Any):
```
TomAugspurger, line 143: Does this need to be typed? Or is it like `self` and doesn't need to be?

A return type would be helpful though.

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -60,15 +64,15 @@ class BlockTransformer(BaseEstimator, TransformerMixin):
...
     """
 
-    def __init__(self, func, *, validate=False, **kw_args):
+    def __init__(self, func: Callable, *, validate: bool = False, **kw_args: Any):
         self.func = func
         self.validate = validate
         self.kw_args = kw_args
 
-    def fit(self, X, y=None):
+    def fit(self, X: Union[ArrayLike, DataFrameType], y: SeriesType = None):
```
TomAugspurger, line 72:
```suggestion
    def fit(self, X: Union[ArrayLike, DataFrameType], y: SeriesType = None) -> 'BlockTransformer':
```

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -60,15 +64,15 @@ class BlockTransformer(BaseEstimator, TransformerMixin):
...
         return self
 
-    def transform(self, X, y=None):
+    def transform(self, X: Union[ArrayLike, DataFrameType], y: SeriesType = None):
```
TomAugspurger, line 75:
```suggestion
    def transform(self, X: Union[ArrayLike, DataFrameType], y: SeriesType = None) -> Union[ArrayLike, DataFrameType]:
```

dask_ml/preprocessing/data.py
```diff
@@ -20,12 +21,14 @@
...
 
+from .._typing import ArrayLike, DataFrameType, SeriesType
+
 _PANDAS_VERSION = LooseVersion(pd.__version__)
 _HAS_CTD = _PANDAS_VERSION >= "0.21.0"
 BOUNDS_THRESHOLD = 1e-7
 
 
-def _handle_zeros_in_scale(scale, copy=True):
+def _handle_zeros_in_scale(scale=ArrayLike, copy=True):
```
TomAugspurger, line 31: Is this ever called with Dask arrays? Or just NumPy?

ci/code_checks.sh
prasunanand: I get these errors:
```
 mypy dask_ml/preprocessing/
dask_ml/preprocessing/label.py:151: error: Item "None" of "Optional[Any]" has no attribute "categories"
dask_ml/preprocessing/label.py:167: error: Item "None" of "Optional[Any]" has no attribute "categories"
dask_ml/preprocessing/label.py:169: error: Item "None" of "Optional[Any]" has no attribute "ordered"
dask_ml/preprocessing/label.py:186: error: Item "None" of "Optional[Any]" has no attribute "categories"
dask_ml/preprocessing/label.py:187: error: Item "None" of "Optional[Any]" has no attribute "ordered"
dask_ml/preprocessing/_encoders.py:267: error: Argument 2 to "zip" has incompatible type "Optional[Any]"; expected "Iterable[Any]"
```

`self.dtype_: Optional[pd.CategoricalDtype] = None`  makes mypy infer `dtype_` as `Any

TomAugspurger: OK, thanks. It seems like we'll need a check that `isinstance(self._dtype, CategoricalDtype)`  under that branch. There's an implicit assumption that type passed to `transform` matches the type passed to `fit`.

prasunanand: Adding `if self.dtype_ is not None:` solves this issue.

prasunanand:
@TomAugspurger Please take a look at this. I think its good to go.

dask_ml/preprocessing/_block_transformer.py
```diff
@@ -60,15 +64,19 @@ class BlockTransformer(BaseEstimator, TransformerMixin):
     Dask Name: hash_pandas_object, 60 tasks
     """
 
-    def __init__(self, func, *, validate=False, **kw_args):
+    def __init__(self, func: Callable, *, validate: bool = False, **kw_args: Any):
```
TomAugspurger, line 67: Is there any way to restrict the type on `Callable`? IIUC, we only know that the first argument will be a NumPy Array or Pandas DataFrame, and that it should return one of those, correct?

dask_ml/preprocessing/_encoders.py
```diff
@@ -184,7 +189,7 @@ def _fit(self, X, handle_unknown="error"):
                 )
 
         self.categories_ = []
-        self.dtypes_ = []
+        self.dtypes_: List[Optional[pd.Categorical]] = []
```
TomAugspurger, line 192: Is this Categorical or CategoricalDtype?

dask_ml/preprocessing/data.py
```diff
@@ -47,7 +50,9 @@ class StandardScaler(sklearn.preprocessing.StandardScaler):
 
     __doc__ = sklearn.preprocessing.StandardScaler.__doc__
 
-    def fit(self, X, y=None):
+    def fit(
+        self, X: Union[ArrayLike, DataFrameType], y: Optional[SeriesType] = None
```
TomAugspurger, line 54: Why just SeriesType here (and other places) for `y`? It can also be an array I think.

dask_ml/preprocessing/label.py
```diff
@@ -222,15 +231,15 @@ def _check_and_search_block(arr, uniques, onehot_dtype=None, block_info=None):
...
+def _construct(x: np.ndarray, categories: np.ndarray) -> scipy.sparse.csr_matrix:
     """Make a sparse matrix from an encoded array.
 
     >>> construct(np.array([0, 1, 0]), np.array([0, 1])).toarray()
     array([[1., 0.],
            [0., 1.],
            [1., 0.]])
     """
-    # type: (np.ndarray, np.ndarray) -> scipy.sparse.csr_matrix
+    # type : (np.ndarray, np.ndarray) -> scipy.sparse.csr_matrix
```
TomAugspurger, line 242: You can perhaps remove this.

dask_ml/preprocessing/_block_transformer.py
prasunanand: Yeah.

I have been able to restrict the return type on `Callable`.

TomAugspurger:
Thanks!

Issue #643, opened 2020-04-14
Hyperparameter optimization benchmarking


TomAugspurger:
It'd be nice to have some benchmarks for how our different hyperparameter optimizers perform. There are a few comparisons that would be useful

1. dask_ml's drop-in replacements for GridSearchCV & RandomizedSearchCV. We're able to deconstruct Pipeline objects to avoid redundant fit calls. This benchmark would compare a `GridSearchCV(Pipeline(...))` for `dask_ml.model_selection.GridSearchCV` and `sklearn.model_selection.GridSearchCV`. We'd expect Dask-ML's to perform better the more CV splits there are and the more parameters that are explored early on in the pipeline (https://github.com/dask/dask-ml/issues/141 has some discussion).
2. Scaling of Dask's joblib backend for large problems. Internally, scikit-learn uses joblib for parallel for loops. With

```python
with joblib.parallel_backend("dask"):
    ...
```

The items in the for loop are executed on the Dask Cluster. There are some issues with the backend (https://github.com/joblib/joblib/issues/1020, https://github.com/joblib/joblib/issues/1025). Fixing those aren't in scope for this work, but we'd like to have benchmarks to understand the current performance and measure the speedup from fixing those.
3. General performance on large datasets with Incremental, Hyperband, etc. We can't really compare to scikit-learn here, since it doesn't handle larger-than-memory datasets. @stsievert may have some thoughts / benchmarks to share here.

cc @dankerrigan. This is more than enough work I think. If you're able to make progress on any of these (or other things you think are important) it'd be great.

mrocklin:
Is there a dataset or workflow on which it makes sense to perform this benchmark? 

I'm more than happy to walk through performance profile information with anyone doing this work.

mrocklin:
Also, as an output of this I'd love to see a blogpost

stsievert:
It's good to see more work on Dask-ML's model selection! I have some ideas I like to see implemented, and would love to see benchmarks on modifications (e.g., to resolve #532).

> Is there a dataset or workflow on which it makes sense to perform this benchmark?

I have a benchmark at https://github.com/stsievert/dask-hyperband-comparison. This benchmark focuses on heavy computation, not large datasets.

mrocklin:
I'm also curious if @dankerrigan has applications from within his workplace that would be both relevant and open (my guess is that this is hard, but it's worth asking :) )

mmccarty:
[...]

PR #644, opened 2020-04-15
make_classification as dask.DataFrame

mmccarty:
Hopefully useful for #643

### DIFF
--- a/dask_ml/datasets.py
+++ b/dask_ml/datasets.py
@@ -1,7 +1,9 @@
 import numbers
+from datetime import timedelta
 
 import dask
 import dask.array as da
+import dask.dataframe as dd
 import numpy as np
 import sklearn.datasets
 import sklearn.utils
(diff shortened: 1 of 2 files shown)


dask_ml/datasets.py
```diff
@@ -377,3 +379,84 @@ def make_classification(
...
+        -   A blocksize like 1000.
+    dates : tuple, optional, default is None
+        tuple of start and end date objects to use for generating
+        random dates in the date column
+    **kwargs
+        Other keyword arguments to pass to `sklearn.datasets.make_classification`
+
+    Returns
+    -------
+    dask.DataFrame
```
TomAugspurger, line 425: I have a slight preference for returning a tuple of (DataFrame, Series) here.

The Series will have the same index as the DataFrame. But we return `y` separately, rather than as a column in the DataFrame. This makes things a bit more consistent with the other dataset generators, which returns tuples for supervised problems.

mmccarty: I was wondering what was standard here. Thanks!

TomAugspurger:
Thanks!

mmccarty:
Thank you! I'll follow up with a regressor df generator.

PR #648, opened 2020-04-17
Updated estimator check link in contrib docs.

mmccarty:
I just noticed the link to the sklearn estimator check was out of date. It looks like it was moved.


TomAugspurger:
Thanks1

PR #649, opened 2020-04-20
Added annotations to cluster module

prasunanand:


prasunanand:
Added annotations to `cluster` module. Added `#type: ignore` as I was unable to solve these issues:
```
dask_ml/_partial.py:43: error: Too many arguments for "__init__" of "object"
dask_ml/_partial.py:78: error: "predict" undefined in superclass
dask_ml/_partial.py:87: error: "predict" undefined in superclass
```

Issue #84, opened 2017-11-04
LogisticRegression cannot train from Dask DataFrame


julioasotodv:
A simple example:

```python
from dask import dataframe as dd
from dask_glm.datasets import make_classification
from dask_ml.linear_model import LogisticRegression

X, y = make_classification(n_samples=10000, n_features=2)

X = dd.from_dask_array(X, columns=["a","b"])
y = dd.from_array(y)

lr = LogisticRegression()
lr.fit(X, y)
```

Returns `KeyError: (<class 'dask.dataframe.core.DataFrame'>,)`

I did not have time to try if it is also the case for other models.

TomAugspurger:
Thanks. At the moment the dask_glm based estimators just work with dask arrays, not dataframes. You can use `.values` to get the array.

I'm hoping to put in some helpers for handling all the extra DataFrame metadata sometime soon, so this will be more consistent.

julioasotodv:
Thank you so much for the quick response!

The problem is that when fitting a glm with intercept (which is usually the case), the dask array containing the features needs to have defined the chunk size, which I believe it is not possible when the array comes from a dataframe.

Anyways, I will reach out to the main dask issue page and ask there.

Thank you!

TomAugspurger:
@julioasotodv, yes I forgot about that case. Let me put something together quick.

julioasotodv:
Do you think there is a way to achieve this without making changes to dask's engine itself?

TomAugspurger:
What do you mean by "dasks's engine"?

See https://github.com/dask/dask-glm/issues/63 for a discussion on the
relationship between dask-ml and dask-glm, and
https://github.com/dask/dask-glm/compare/master...TomAugspurger:add-intercept-dd
for what the fix will look like.

On Mon, Nov 6, 2017 at 5:05 PM, Julio Antonio Soto <notifications@github.com

julioasotodv:
I see. Would it work with that fix, even if chunksize is not defined for the underlying dask array?

TomAugspurger:
Yes, that should work. The solvers only require that the shape along the second axis is known:

```python
from dask_ml.linear_model import LinearRegression
from dask_ml.datasets import make_regression

X, y = make_regression(chunks=50)

df = dd.from_dask_array(X)
X2 = df.values  # dask.array with unknown chunks along first dim

lm = LinearRegression(fit_intercept=False)
lm.fit(X2, y)
```

[...]

PR #653, opened 2020-04-26
MAINT: provide potential solutions in warning on "dataframes not accepted"

stsievert:
It modifies message of the `TypeError` in utils.py if Dask Dataframes are not accepted. It suggests code that may resolve the issue.

I only provide `ddf.to_dask_array(...)` solutions because Pandas recommends using `to_numpy` instead of `values`: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.values.html

**Reference issues/PRs**
The motivation is https://github.com/dask/dask-ml/issues/84. I tried to get the code to fail with unknown chunk sizes, but the code worked. Instead, I decided to modify the message about not accepting Dask DataFrames with a potential solution.

(note: I make sure `fit_intercept in [True, False]` because of https://github.com/dask/dask-ml/issues/84#issuecomment-342526205).

### DIFF
--- a/dask_ml/utils.py
+++ b/dask_ml/utils.py
@@ -174,8 +174,12 @@ def check_array(
 
     elif isinstance(array, dd.DataFrame):
         if not accept_dask_dataframe:
-            raise TypeError("This estimator does not support dask dataframes.")
-        # TODO: sample?
+            raise TypeError(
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks.

Issue #135, opened 2018-02-05
Blockwise ensemble meta-estimator


TomAugspurger:
Fit clones of an estimator on each block / partition, and ensemble the results somehow http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.VotingClassifier.html, http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.BaggingClassifier.html#sklearn.ensemble.BaggingClassifier, http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.BaggingRegressor.html#sklearn.ensemble.BaggingRegressor

js3711:
2nd this for tree based methods as well.  
http://scikit-learn.org/stable/modules/ensemble.html#ensemble
Namely ExtraTrees, and RandomForest based approaches

This might involve changing how the current dask_ml cross validation methods work as well.  It would be nice to add the individual "voters" to the task graph.

jakirkham:
+1 for RandomForest :)

Also would add AdaBoost and Gradient Tree Boosting.

TomAugspurger:
Closed by #657.

PR #657, opened 2020-05-04
Add blockwise ensemble meta-estimators

TomAugspurger:
This adds `BlockwiseVotingClassifier` and `BlockwiseVotingRegressor`, which are meta-estimators for

1. *Blockwise training* of the sub-estimator.
2. *Ensemble prediction* on new data.

Given an input array split into `k` partitions, `k` clones of the *sub*estimator are fit independently, one per partition. This is efficient since we don't need to move any data (assuming the partitions of X and y are co-located).

At prediction time we combine the predictions from each of the `k` fitted models. For classification we take the class with either the most votes (`voting="hard"`) or the highest total probability (`voting="soft")`. See https://scikit-learn.org/stable/modules/ensemble.html#voting-classifier. For regression we take the average.

```python
In [1]: import sklearn.linear_model
   ...: import dask_ml.datasets
   ...: import dask_ml.ensemble
   ...:
   ...: X, y = dask_ml.datasets.make_classification(n_features=20, chunks=25)
   ...:
   ...: clf = dask_ml.ensemble.BlockwiseVotingClassifier(
   ...:     sklearn.linear_model.LogisticRegression(), voting="soft",
   ...:     classes=[0, 1]
   ...: )
   ...:
   ...: clf.fit(X, y)

In [2]: clf.estimators_
Out[2]:
[LogisticRegression(C=1.0, class_weight=None, dual=False, fit_intercept=True,
                    intercept_scaling=1, l1_ratio=None, max_iter=100,
                    multi_class='auto', n_jobs=None, penalty='l2',
                    random_state=None, solver='lbfgs', tol=0.0001, verbose=0,
                    warm_start=False),
 LogisticRegression(C=1.0, class_weight=None, dual=False, fit_intercept=True,
                    intercept_scaling=1, l1_ratio=None, max_iter=100,
                    multi_class='auto', n_jobs=None, penalty='l2',
                    random_state=None, solver='lbfgs', tol=0.0001, verbose=0,
                    warm_start=False),
 LogisticRegression(C=1.0, class_weight=None, dual=False, fit_intercept=True,
                    intercept_scaling=1, l1_ratio=None, max_iter=100,
                    multi_class='auto', n_jobs=None, penalty='l2',
                    random_state=None, solver='lbfgs', tol=0.0001, verbose=0,
                    warm_start=False),
 LogisticRegression(C=1.0, class_weight=None, dual=False, fit_intercept=True,
                    intercept_scaling=1, l1_ratio=None, max_iter=100,
                    multi_class='auto', n_jobs=None, penalty='l2',
                    random_state=None, solver='lbfgs', tol=0.0001, verbose=0,
                    warm_start=False)]
```


TomAugspurger:
cc @js3711 from https://github.com/dask/dask-ml/issues/135 and @nbren12.

One thing I'm not sure about is what the output of `.transform` should be. In scikit-learn's VotingClassifier, `.transform` returns the *predictions* of each sub-estimator. We can add that later if we want to match that.

dask_ml/ensemble/_blockwise.py
```diff
@@ -0,0 +1,240 @@
...
+            dtype = "int"
+        else:
+            dtype = "float"
+
+        if isinstance(X, da.Array):
+            results = [
+                X.map_blocks(_predict, dtype=dtype, estimator=estimator, drop_axis=1)
+                for estimator in self.estimators_
+            ]
+            combined = da.vstack(results).T.rechunk({1: -1})
```
TomAugspurger, line 57: This is kinda slow for many estimators (many partitions). We end up with `X.npartitions * len(self.classifiers_)` tasks. Looking into doing a "batch" predict where each task represents the predictions from every estimator on that partition stacked into a single ndarray.

TomAugspurger: Fixed for array. For dask.dataframe.map_partitions didn't especially like returning a 3D array. Looking into it a bit, but not a blocker right now.

PR #659, opened 2020-05-05
Normalize fixup

### LINKED ISSUE #658: BaseEstimator tokenization may cause issues for fitted models
In https://github.com/dask/dask-ml/blob/master/dask_ml/model_selection/_normalize.py#L18-L24, we register a tokenizer for scikit-learn models that inherit from BaseEstimator. This can cause issues when we have multiple instances of the same model that have been fitted to different pieces of data (as in https://github.com/dask/dask-ml/pull/657).

```python
In [8]: import dask
   ...: import sklearn.datasets
   ...: import sklearn.linear_model
   ...: import dask_ml.model_selection._normalize
   ...:
   ...: m1 = sklearn.linear_model.LogisticRegression(random_state=0)
   ...: m2 = sklearn.linear_model.LogisticRegression(random_state=0)
   ...:
   ...: # Not fitted, fine for these to match
   ...: assert dask.base.tokenize(m1) == dask.base.tokenize(m2)
   ...:
   ...: X1, y1 = sklearn.datasets.make_classification()
   ...: X2, y2 = sklearn.datasets.make_classification()
   ...: m1.fit(X1, y1)
   ...: m2.fit(X2, y2)
   ...:
   ...: # After fitting, these should be different!
   ...: assert dask.base.tokenize(m1) != dask.base.tokenize(m2)
   ...:
   ...:
AssertionError                            Traceback (most recent call last)
<ipython-input-8-590215541e34> in <module>
     16
     17 # After fitting, these should be different!
---> 18 assert dask.base.tokenize(m1) != dask.base.tokenize(m2)
     19

AssertionError:
```

I think that `normalize_estimator` should also (try to) hash things like `coef_`. Since this is supposed to work for arbitrary estimators, the safest thing to do is tokenize any attribute ending in `_`. We might exclude selected attributes like `GridSearchCV.cv_results_`, since those include things that are incidental (like the time it took to train a specific split).

TomAugspurger:
Closes #658.

### DIFF
--- a/dask_ml/__init__.py
+++ b/dask_ml/__init__.py
@@ -1,7 +1,18 @@
 from pkg_resources import DistributionNotFound, get_distribution
 
+# Ensure we always register tokenizers
+from dask_ml.model_selection import _normalize
+
+__all__ = []
+
 try:
     __version__ = get_distribution(__name__).version
+    __all__.append("__version__")
 except DistributionNotFound:
     # package is not installed
     pass
+
+
+del DistributionNotFound
+del get_distribution
+del _normalize
--- a/dask_ml/model_selection/_normalize.py
+++ b/dask_ml/model_selection/_normalize.py
@@ -32,7 +32,7 @@ def normalize_estimator(est):
             continue
         try:
             val = getattr(est, attr)
-        except sklearn.exceptions.NotFittedError:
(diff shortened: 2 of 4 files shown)


TomAugspurger:
Only failure is the known hang on incremental `test_basic` .

PR #669, opened 2020-05-21
CI: Added 3.8

### LINKED ISSUE #663: Add Python 3.8 CI / support
Numba supports Python 3.8 now so we can support it too.

TomAugspurger:
Closes #663


TomAugspurger:
Haven't looked too closely, but it seems like the `LoggingContext.__exit__` is running before you get to `_incrementalf.fit` where the logger is used (set a `breakpoint` in the `__exit__`). Are you using it in the right place?

PR #670, opened 2020-05-23
BUG: clearly show error message while initing models

stsievert:
**What does this PR implement?**
In https://github.com/dask/dask-examples/pull/149/, @ToddMorrill had an error while initializing a model. However, he couldn't see the traceback. This PR resolves that.

**Reference issues/PRs**
Motivation is https://github.com/dask/dask-examples/pull/149/

PR #671, opened 2020-05-23
MAINT: show errors while scoring models in model selection

stsievert:
In #670 I more clearly showed errors while initializing models. This PR does the same thing, but while scoring models.

**Reference issues/PRs**
This PR is a follow up to #670 (ideally this commit would be in that PR).


stsievert:
- [ ] make clear that each `partial_fit` call sees one chunk of the data

PR #682, opened 2020-06-17
Update pca.py

sosofun:
to fix bug: when X is dask.dataframe,  the pca.explained_variance_ratio_ is  will be  nan

### DIFF
--- a/dask_ml/decomposition/pca.py
+++ b/dask_ml/decomposition/pca.py
@@ -250,7 +250,9 @@ class PCA(sklearn.decomposition.PCA):
         return solver
 
     def _fit(self, X):
+        n_samples, n_features = X.shape
         if isinstance(X, dd.DataFrame):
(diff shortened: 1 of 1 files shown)


dask_ml/decomposition/pca.py
```diff
@@ -250,7 +250,9 @@ def _get_solver(self, X, n_components):
         return solver
 
     def _fit(self, X):
+        n_samples, n_features = X.shape
         if isinstance(X, dd.DataFrame):
+            n_samples = len(X)
```
TomAugspurger, line 255: This will trigger a computation of the task graph backing the DataFrame. We'd like to avoid that until the main `compute` down around line 330.

TomAugspurger:
Closing this to clear the backlog. Let us know if you're interested in working on it @sosofun.

PR #688, opened 2020-06-30
Log amount of found clusters in kmeans init

JulianWgs:
- Put the changes behind the old information to not destroy old processing pipelines (if there are any)

### DIFF
--- a/dask_ml/cluster/k_means.py
+++ b/dask_ml/cluster/k_means.py
@@ -447,7 +447,10 @@ def init_scalable(
     # to do that.
 
     if len(centers) < n_clusters:
-        logger.warning("Found fewer than %d clusters in init.", n_clusters)
+        logger.warning(
+            f"Found fewer than {n_clusters} clusters in init "
+            f"(found {len(centers)})."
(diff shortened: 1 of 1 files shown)


dask_ml/cluster/k_means.py
```diff
@@ -447,7 +447,10 @@ def init_scalable(
     # to do that.
 
     if len(centers) < n_clusters:
-        logger.warning("Found fewer than %d clusters in init.", n_clusters)
+        logger.warning(
```
TomAugspurger, line 450: Can we keep the `%d` style formatting? It avoids the cost of formatting when the logger isn't enabled.

JulianWgs: You mean that the f-string syntax is slower?

TomAugspurger: Can be, so I prefer to consistently use `%`-style for logging.

TomAugspurger:
Thanks. The sklearn dev CI issue is unrelated. I'll debug that separately.

PR #695, opened 2020-07-07
remove xpassing tests

TomAugspurger:


TomAugspurger:
I'll track down the timeouterror on windows separately.

PR #698, opened 2020-07-15
datasets.make_regression as Dask dataframe and series

mmccarty:
- Added `return_dataframes` wrapper to repackage Xs and ys
- Create `make_regression_df` using wrapper
- Refactored `make_classification_df` to reuse wrapper

### DIFF
--- a/dask_ml/datasets.py
+++ b/dask_ml/datasets.py
@@ -390,7 +390,7 @@ def random_date(start, end):
 
 
 def return_dataframes(func):
-    """Wraps a function that returns a tuple of Dask arrays (X, y) to return 
+    """Wraps a function that returns a tuple of Dask arrays (X, y) to return
     a tuple of Dask dataframe and series (X_df, y_series).
     """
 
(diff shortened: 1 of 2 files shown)


mmccarty:
cc @dhavide

mmccarty:
cc @TomAugspurger I promised this a while back. Do you think I should do anything fancy for the docstrings? This just specifies the new return values.

dask_ml/datasets.py
```diff
@@ -388,75 +389,50 @@ def random_date(start, end):
...
-                    columns=["date"],
-                ),
-            ],
-            axis=1,
-        )
 
-    return X_df, y_series
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        """
```
TomAugspurger, line 399: Is this the eventual docstring for `make_classification_df`? If so, can you add the additional parameters to the docstring (dates, response_rate, etc.)

mmccarty: It is. Yes, I'll do that and add a note to see the array versions.

Issue #696, opened 2020-07-12
Provide wrappers for popular ML libraries


stsievert:
It'd be convenient to provide support for use of Keras or PyTorch models in model selection. There are two issues:

1. Keras/PyTorch models don't conform to the Scikit-learn API.
2. Keras models are not pickle-able.

stsievert:
I think this is possible with these wrappers:

* [skorch], a library that brings the Scikit-learn API to PyTorch. This is what's used in the "[Better and faster hyperparameter optimization with Dask ][1]."
* There's an implementation of `__reduce_ex__` for  Keras models in https://github.com/tensorflow/tensorflow/pull/39609/

[1]:https://blog.dask.org/2019/09/30/dask-hyperparam-opt
[skorch]:https://skorch.readthedocs.io/en/stable/

adriangb:
If you're looking for a way to make Keras models conform to the scikit-learn API, check out [SciKeras](https://github.com/adriangb/scikeras) (full disclosure: I'm the author)

TomAugspurger:
Thanks for the link Adrian. To minimize our maintenance burden, I'd aim for the goal that our `model_selection` estimators work with any model implementing the scikit-learn interface, and encourage the development / use of wrappers like skorch and SciKeras.

On top of that, we have the additional burden of these models needing to work well with distributed's serialization. To the extent possible, that functionality should be in the projects themselves (making Keras models picklable) or in distributed.

mrocklin:
How hard is it to support PyTorch/Keras fit/predict APIs?  If this is as simple as making a function like the following, then I would be in favor

```python
def fit(estimator, X, y=None):
    if hasattr(estimator, "fit"):
        return estimator.fit(X, y)
    elif hasattr(estimator, ...): #  pytorch-like
        return ...
    elif hasattr(estimator, ...): # keras-like
        return ...
```

mrocklin:
For serialization I think that we have a decent Pytorch serializer in distributed (early work from @stsievert if I recall correctly).  I don't think that we have anything for Keras today.

mrocklin:
Serialization is also maybe something that we could ask for help from the RAPIDS folks like @quasiben @jakirkham @pentschev  .  It's not RAPIDS obviously, but these are often GPU related and that team is familiar with these sorts of issues.

adriangb:
[...]

PR #699, opened 2020-07-15
DOC, TST: Wrapping of PyTorch models

stsievert:
It provides wrappers for Keras/PyTorch models, primarily aimed at model selection. This PR wraps PyTorch/Keras by relying on SciKeras/Skorch.

Of course, this is a very thin wrapper. I think it's warranted for the following reasons:

* PyTorch and Keras/Tensorflow are very popular. No other deep learning library comes close (i.e, Chainer/MXNet), and both PyTorch and Keras are about 15x more popular than Scikit-learn on Google Trends.
* Skorch is not suited for model selection (too much printing, has validation split by default, too many epochs for `partial_fit`).

**References issues/PRs**
* #696: "Provide a wrapper for popular ML libraries"
* https://github.com/adriangb/scikeras/pull/17: "ENH: Implement partial_fit for SciKeras models"
* https://github.com/adriangb/scikeras/issues/18: "Default parameters of build_fn are not returned by get_params."

This PR will be a WIP until https://github.com/adriangb/scikeras/issues/19 is resolved.


dask_ml/wrappers.py
```diff
@@ -551,3 +551,141 @@ def _predict_proba(part, estimator):
...
     return estimator.transform(part)
+
+
+try:
+    from scikeras.wrappers import KerasRegressor, KerasClassifier
+except:
+
+    class _KerasError:
+        def __init__(self, *args, **kwargs):
+            raise ModuleNotFoundError(
```
TomAugspurger, line 562: I'll need to check the intended usage, but what's the advantage to delaying the error until class creating time rather than import time? As a user I think it'd be best to know immediately, and I think the code is simpler.

The potential issue I see: what if I just want to use torch and not tensforflow? In that case we'd need to do something fancier.

Perhaps we require Python 3.7 and use https://www.python.org/dev/peps/pep-0562/.

dask_ml/wrappers.py
```diff
@@ -551,3 +551,141 @@ def _predict_proba(part, estimator):
...
+                "SciKeras not installed. Please install Tensorflow, Keras "
+                "and SciKeras. This code will work:\n\n"
+                "    $ pip install tensorflow keras scikeras\n"
+            )
+
+    class KerasClassifier(_KerasError):
```
TomAugspurger, line 568: Why the subclassing?

mrocklin:
What would be necessary to support PyTorch/Keras natively?  Could we provide a function that called fit/partial_fit on Scikit-Learn estimators and something else on torch/keras models?

TomAugspurger:
I'm not sure, but I don't think that'll suffice. It would also need things
like an sklearn-compatible `.score`, `get_params` & `set_params` to support
the search. And at this point it feels like we'd be implementing (a poorer
version of) scikeras / skorch. But again, that's just a guess. Maybe things
are easier than I expect.

mrocklin:
Ah, fair point

stsievert:
> Can you expand a bit more on the value added by our wrappers? In particular, to what extent can the changes here be pushed upstream? ...  having the tests, and ideally examples at dask-examples will be valuable

That's part of why I opened this PR. I think having tests, an example at dask-examples and another Dask-ML documentation page would suffice to meet my goals. I aim for these implementations to be easy to discover and work well with model selection. I think those needs are met without adding an implementation in `dask_ml.wrappers`.

I'll rework this PR to be focused on documentation/testing, and open a PR in dask-examples.

tests/model_selection/test_wrappers.py
```diff
@@ -48,14 +54,30 @@ def _keras_build_fn(lr=0.01):
...
+
     X, y = mnist()
     assert X.ndim == 2 and X.shape[-1] == 784
     assert y.ndim == 1 and len(X) == len(y)
+    assert isinstance(X, np.ndarray) and isinstance(y, np.ndarray)
 
     model = KerasClassifier(build_fn=_keras_build_fn, lr=0.1)
     params = {"lr": loguniform(1e-3, 1e-1)}
 
+    with pytest.warns(DataConversionWarning):
```
TomAugspurger, line 71: Can you post this warning and a bit about what's happening?

ci/environment-3.8.yaml
```diff
@@ -30,3 +30,7 @@ dependencies:
   - pip
   - pip:
     - pytest-azurepipelines
+    - tensorflow
```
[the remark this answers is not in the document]
stsievert: Would it be okay to only run on the master branch?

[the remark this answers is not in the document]
TomAugspurger: Yeah I think that's fine.

tests/model_selection/test_wrappers.py
stsievert: I think this is happening because SciKeras plays with the shapes: https://github.com/adriangb/scikeras/issues/19#issuecomment-658549923

adriangb: Yep. Keras and ScikitLearn don't always agree on the shape of things (ex: `(n_points, )` vs `(n_points, 1)`). SciKeras *tries* to convert back and forth to make everyone happy, but I'm not sure that it can make everyone happy and avoid modifying the user's data 100% of the time. I opened an issue to try and improve this behavior (and hopefully reduce the occurrence of that warning): https://github.com/adriangb/scikeras/issues/20

stsievert:
> Could we provide a function that called fit/partial_fit on Scikit-Learn estimators and something else on torch/keras models?

For model selection not really – @TomAugspurger is right. I suppose it would be possible to hack a solution to use arbitrary training functions using distributed variables:

<details>

``` python
from keras import fit
from distributed import Variable

def train_model(hidden=4, epochs=10, data: Variable = None, stop: Variable = None):
    model = create_model(hidden=hidden)
    X_train, y_train, X_test, y_test = ...
    for epoch in range(epochs):
        keras.fit(model, X_train, y_train, epochs=1, workers=4)
        score = model.score(X_test, y_test)
        datum = {"score": score, "pf_calls": epoch}
        d = data.get()
        data.set(d + [datum])
        if stop.get():
            break

class FunctionTrainer(BaseEstimator):
    def __init__(self, fn, **kwargs):
        vars(self).update(kwargs)
        self.fn = fn
        self._data = Variable("_data")
        self._stop = Variable("_stop")
        self._pf_calls = 0

    def _wait_for_training_to_complete(self) -> Dict[str, Any]:
        while True:
            data = self._data.get()
            pf_calls = {d["pf_calls"] for d in data}
            if self._pf_calls in pf_calls:
                break
            await asyncio.sleep(0.1)
        return data[-1]

    def _initialize(self):
        self.data.set([])
        self._stop.set(False)
        kwargs = self.get_params()
        client.submit(self.fn, kwargs, data=self._data, stop=self._stop)

    def _stop_training(self):
        self._stop.set(True)

    def partial_fit(self, X, y):
        datum = self._wait_for_training_to_complete()
        self._pf_calls += 1
        return self

    def score(self, X, y):
        datum = self._wait_for_training_to_complete()
        return datum["score"]
```

We could make this work with Dask-ML's model selection; it'd have to call `FunctionTrainer._stop_training` when it kills a model.

</details>

It's easier for training a single model. The PyTorch/Keras implementation would be similar to [dask-glm]: the model and optimization would be reside client side, and the Dask workers would be tasked with computing the gradient.

[dask-glm]:https://github.com/dask/dask-glm

docs/source/keras.rst
```diff
@@ -0,0 +1,45 @@
+Keras
```
TomAugspurger, line 1: Perahaps title this Keras and Tensorflow, to make it easier for people just familiar with tensorflow to find this?

docs/source/keras.rst
```diff
@@ -0,0 +1,45 @@
...
+
+.. _Keras Sequential model: https://keras.io/api/models/sequential/
+
+.. code-block:: python
+
+   import tensorflow as tf
+   from tensorflow.keras.layers import Dense, Activation, Dropout
+   from tensorflow.keras.models import Sequential
+
+   def _keras_build_fn(lr=0.01):
```
TomAugspurger, line 21:
```suggestion
   def build_model(lr=0.01):
```

docs/source/keras.rst
```diff
@@ -0,0 +1,45 @@
...
+       opt = tf.keras.optimizers.SGD(learning_rate=lr)
+       model.compile(loss="categorical_crossentropy", optimizer=opt, metrics=["accuracy"])
+       return model
+
+Now, we can use the SciKeras to create a Scikit-learn compatible model:
+
+.. code-block:: python
+
+   from scikeras.wrappers import KerasClassifier, KerasRegressor
+   model = KerasClassifier(build_fn=_keras_build_fn, lr=0.1)
```
TomAugspurger, line 35:
```suggestion
   model = KerasClassifier(build_fn=build_model, lr=0.1)
```

ci/environment-3.8.yaml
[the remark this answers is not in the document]
TomAugspurger: Will probably need to remove these here, and add a secondary conda / pip install that only runs on certain jobs.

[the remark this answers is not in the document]
stsievert: Thanks. That's in `posix.yaml`. This PR isn't ready for merge; SciKeras should have a new release soon that incorporates the recent changes.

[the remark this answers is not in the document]
stsievert: Azure pipelines are installing Scikeras from master right now

tests/model_selection/test_wrappers.py
```diff
@@ -0,0 +1,128 @@
+import pickle
```
TomAugspurger, line 1: Can you split this into a `test_keras.py` and a `test_pytorch.py`?

Then you can use a `pytest.importorskip` at the top of the file and the tests won't have to worry about missing imports.

ci/posix.yaml
```diff
@@ -33,6 +34,16 @@ jobs:
   - bash: conda env create --quiet --file=$(envFile) --name=dask-ml-test && conda list -n dask-ml-test
     displayName: "install"
 
+  - bash: |
+      conda install pytorch torchvision -c pytorch
+      pip install skorch
+      pip install tensorflow scikeras keras
```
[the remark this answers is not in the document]
adriangb: SciKeras is currently not on conda-forge. Would it help if it was?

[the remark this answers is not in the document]
TomAugspurger: It's not a huge deal, especially for pure-python libraries (like I assume SciKeras is). I'm more concerned about tensorflow.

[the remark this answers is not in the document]
adriangb: Yeah SciKeras is pure Python, although it depends on ScikitLearn and TF. I've never installed TF from conda-forge, I'm not sure if it's published there.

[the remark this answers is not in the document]
stsievert: It's not on conda-forge, and the docs recommend a pip install: https://www.tensorflow.org/install/

They have a separate page making sure GPU support is enabled: https://www.tensorflow.org/install/gpu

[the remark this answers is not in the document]
TomAugspurger: Is https://github.com/conda-forge/tensorflow-feedstock not what we want? GPU shouldn't be relevant since they CI system won't have GPUs.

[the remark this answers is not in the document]
TomAugspurger: Seems that conda-forge's tensorflow isn't updated for 2.x yet, so it'll have to come from PyPI.

dask_ml/model_selection/_incremental.py
```diff
@@ -522,7 +522,10 @@ def _validate_parameters(self, X, y):
         kwargs = dict(accept_unknown_chunks=False, accept_dask_dataframe=False)
         X = self._check_array(X, **kwargs)
         y = self._check_array(y, ensure_2d=False, **kwargs)
-        scorer = check_scoring(self.estimator, scoring=self.scoring)
+        estimator = self.estimator
+        if isinstance(estimator, Future):
+            estimator = await estimator.result()
```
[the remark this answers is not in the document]
stsievert: PyTorch and Keras models are often large.* When this model is put into the DAG, a warning is printed saying "Use data_future = client.scatter(data) and pass that to client.submit." This code allows `client.scatter` to take a model future instead of a (large) model.

I don't know if it's necessary to call `check_scoring`. In the case of PyTorch/Keras models it's probably not necessary. If it is, it won't be so bad because `check_scoring` only checks for attributes of `estimator`. It might not be necessary to bring the entire model to the client; I think we could use the implementation below:

<details>
<summary>Check for attrs only</summary>

``` python
from types import SimpleNamespace
def getattrs(cls):
    return [x for x in dir(cls) if x[0] != "_" and callable(getattr(cls, x))]

def _validate_params(self, X, y):
    ...
    attrs = await client.compute(getattrs, self)
    kwargs = {a: lambda x: for a in attrs}
    dummy = SimpleNamespace(**kwargs)
    scorer = check_scoring(estimator, scoring=self.scoring)
    return X, y, scorer
```

</details>

On model size: very small neural nets have at least 100,000 parameters, or 360KB. Small neural nets have about 1 million parameters, about 3.8MB. The smallest deep network has 11 million parameters, or 44MB:

``` python
>>> import torchvision.models as m
>>> sum([x.nelement() for x in m.resnet18.parameters()]) / 1e6
11.689512
```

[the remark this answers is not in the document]
mrocklin: Would it be possible to keep the estimator remote in this case and send the `check_scoring` function up?

```python
scorer = await client.submit(check_scoring, estimator, scoring=self.scoring)
```

I've been playing with cloud systems from my laptop a lot recently (coiled stuff) and I've started to become very sensitive to sending things back to the client.  My use case may be uncommon today, so we don't need to pay too much attention here, but it seems like it might be easy and slightly more efficient.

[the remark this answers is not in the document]
stsievert: D'oh. Thanks for the suggestion @mrocklin!

ci/posix.yaml
```diff
@@ -33,6 +34,16 @@ jobs:
   - bash: conda env create --quiet --file=$(envFile) --name=dask-ml-test && conda list -n dask-ml-test
     displayName: "install"
 
+  - bash: |
+      conda install pytorch torchvision -c pytorch
```
TomAugspurger, line 38:
```suggestion
      conda install -y pytorch torchvision -c pytorch
```

ci/posix.yaml
```diff
@@ -33,6 +34,16 @@ jobs:
...
+      pip install skorch
+      pip install tensorflow scikeras keras
+      pip install -U git+https://github.com/adriangb/scikeras.git
+    displayName: "install Tensorflow/PyTorch"
+    # condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
```
TomAugspurger, line 43: Just a reminder: comment this before merging.

stsievert:
I've put Keras in a separate PR, #713. I think it needs some more work, and I don't think it should block this PR. For more detail on the issues, see https://github.com/dask/dask-ml/pull/713#issuecomment-663924182.

Now this PR focuses on documenting a PyTorch wrapper and reorganizing the documentation.

dask_ml/model_selection/_incremental.py
```diff
@@ -515,21 +515,25 @@ def __init__(
...
                 )
             )
 
-        # Make sure dask arrays are passed so error on unknown chunk size is raised
         kwargs = dict(accept_unknown_chunks=True, accept_dask_dataframe=True)
         if not isinstance(X, dd.DataFrame):
             X = self._check_array(X, **kwargs)
-        if not isinstance(y, dd.Series):
+        if not isinstance(y, dd.DataFrame):
             y = self._check_array(y, ensure_2d=False, **kwargs)
```
mrocklin, line 530: Why this change?  I would expect that `y` would more often be a Series than a DataFrame.  Should we include either type?

stsievert: Thanks for the catch, I think this is a bad merge. I've fixed it.

stsievert:
Are we particularly tied to isort 4.3.21? I can not get the `pytest.importorskip` and the following torch imports to work with isort 4.3.21. It works under isort >= 5 because isort >=5.0.0 supports action comments ([source][1]) like "isort: skip" or "isort: split."

[1]:https://timothycrosley.github.io/isort/docs/major_releases/introducing_isort_5/#action-comments

TomAugspurger:
> Are we particularly tied to isort 4.3.21?

No, but I don't think we'd want those anyway. 5b6e20c277443d3b765068aa0f45fbd788d3e838 will hopefully work.

ci/posix.yaml
```diff
@@ -33,6 +33,12 @@ jobs:
   - bash: conda env create --quiet --file=$(envFile) --name=dask-ml-test && conda list -n dask-ml-test
     displayName: "install"
 
+  - bash: |
+      conda install -y pytorch -c pytorch
+      pip install skorch
```
TomAugspurger, line 38: This reinstalled a bunch of packages already present

```
Installing collected packages: scipy, tabulate, joblib, threadpoolctl, scikit-learn, skorch
Successfully installed joblib-0.16.0 scikit-learn-0.23.1 scipy-1.5.2 skorch-0.8.0 tabulate-0.8.7 threadpoolctl-2.1.0
```

Can you `pip install --no-deps skorch` and explicitly include any of skorch's other deps we don't already have?  Maybe just tabulate and tqdm (if they aren't already present).

TomAugspurger:
Ah @stsievert I think that torch & skorch are being installed into the `base` conda environment. I think the tests are run in the `dask-ml-test` env.

stsievert:
The Windows CI is failing on the model selection tests named `test_small` and `test_warns_scores_per_fit`. The traceback reports `TimeoutError` or `CancelledError`, and they pass on my non-Windows machine.

ci/posix.yaml
TomAugspurger:
```suggestion
    condition: eq(variables['Build.SourceBranch'], 'refs/heads/master')
```

TomAugspurger:
Thanks @stsievert!

PR #701, opened 2020-07-18
Support Dask Dataframes in Hyperband

mrocklin:
It doesn't look like we actually use exact chunk sizes anywhere, so this should be ok?

I may be wrong though.  Regardless there is probably still some cleanup to do here.  I thought I'd push up something early for feedback though.

cc @stsievert

### DIFF
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -642,9 +642,7 @@ class BaseIncrementalSearchCV(ParallelPostFit):
 
         X, y, scorer = self._validate_parameters(X, y)
 
-        X_train, X_test, y_train, y_test = self._get_train_test_split(
-            X, y, shuffle=True
-        )
+        X_train, X_test, y_train, y_test = self._get_train_test_split()
 
         with context:
             results = await fit(
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -642,7 +642,7 @@ class BaseIncrementalSearchCV(ParallelPostFit):
 
         X, y, scorer = self._validate_parameters(X, y)
 
-        X_train, X_test, y_train, y_test = self._get_train_test_split()
+        X_train, X_test, y_train, y_test = self._get_train_test_split(X, y)
 
         with context:
             results = await fit(
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -568,8 +568,7 @@ class BaseIncrementalSearchCV(ParallelPostFit):
         else:
             test_size = self.test_size
         X_train, X_test, y_train, y_test = train_test_split(
-            X, y, test_size=test_size, random_state=self.random_state,
-            shuffle=True
+            X, y, test_size=test_size, random_state=self.random_state, shuffle=True
         )
         return X_train, X_test, y_train, y_test
 
@@ -645,8 +644,9 @@ class BaseIncrementalSearchCV(ParallelPostFit):
 
         X, y, scorer = self._validate_parameters(X, y)
 
-        X_train, X_test, y_train, y_test = self._get_train_test_split(X, y,
-                shuffle=True)
+        X_train, X_test, y_train, y_test = self._get_train_test_split(
+            X, y, shuffle=True
+        )
 
         with context:
             results = await fit(
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -179,6 +179,15 @@ async def _fit(
 
     # Convert testing data into a single element on the cluster
     # This assumes that it fits into memory on a single worker
+    if isinstance(X_train, (dd.DataFrame, dd.Series)):
+        X_train = X_train.to_dask_array()
+    if isinstance(X_test, (dd.DataFrame, dd.Series)):
+        X_test = X_test.to_dask_array()
+    if isinstance(y_train, dd.Series):
+        y_train = y_train.to_dask_array()
+    if isinstance(y_test, dd.Series):
+        y_test = y_test.to_dask_array()
+
     if isinstance(X_test, da.Array):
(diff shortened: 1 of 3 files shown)


dask_ml/model_selection/_incremental.py
```diff
@@ -635,7 +643,10 @@ async def _fit(self, X, y, **fit_params):
             context = dummy_context()
 
         X, y, scorer = self._validate_parameters(X, y)
-        X_train, X_test, y_train, y_test = self._get_train_test_split(X, y)
+
+        X_train, X_test, y_train, y_test = self._get_train_test_split(
+            X, y, shuffle=True
```
stsievert, line 648: Is this `shuffle=True` needed? There's made a modification above where `shuffle=True` is hard-coded.

mrocklin: This method warns when given a dataframe and `shuffle=` is not specified.  We need to choose some default.  Would `False` be a better choice?

mrocklin:
> I don't have any issues or qualms with this PR, especially if the tests pass. The failing test looks relevant, but only because we were performing an unnecessary check.

Ah indeed.  Fixed I think.

Interestingly.  Tests also pass if I remove the to_dask_array lines.  I'm curious do we know what support is like for passing Pandas dataframes around to models is?  Is this something that we would want to leave for the user?

dask_ml/model_selection/_incremental.py
stsievert: Ah, I see. I prefer `shuffle=True` because it's the right choice from an optimization perspective. For example, what if the Dask Dataframe are CSVs from each state? Calling `partial_fit` on any one CSV is not the right answer if each CSV is very different.

mrocklin: I think that in this function shuffle is only per-partitions, not between partitions.  I think that we would leave inter-partition shuffling to the user ahead of sending to Hyperband

stsievert: > I think that in this function shuffle is only per-partitions, not between partitions.

That's not true; it looks like for default value of `blockwise` depends on whether it's a Dask Array or DataFrame: https://github.com/dask/dask-ml/blob/fc6a04aec4517abca318bc7e44e99f8e540b6fe6/dask_ml/model_selection/_split.py#L392-L395

But that's unrelated to this PR. I think `shuffle=True` is a good choice in case there's order. Maybe we should make `_get_train_test_split` public?

TomAugspurger: > Maybe we should make _get_train_test_split public?

What's the motivation for that?

---

I think

TomAugspurger: Sorry, my last comment is misleading. DataFrame's shuffle_split is definately blockwise-only. So I think we have everything we want here? `shuffle=True, blockwise=True` was the default behavior on master, and that's still what we have?

TomAugspurger:
The sklearn dev failure can be ignore. I need to look into it today or tomorrow.

Planning to merge later today, assuming the discussion in https://github.com/dask/dask-ml/pull/701/files/485ff530bca48eb6051591c75dbe6a5965d42333#diff-0a54ded74c8013caf0588e9a5c879e99 has been resolved.

dask_ml/model_selection/_incremental.py
stsievert: >> Maybe we should make _get_train_test_split public?
>
> What's the motivation for that?

In case someone wants a specific train/test split. Maybe they have time series and don't want `shuffle=True` or `blockwise=True`? Or maybe they want the test set to be a specific chunk from the Dask Array?

stsievert: > shuffle=True, blockwise=True was the default behavior on master, and that's still what we have?

`shuffle=True` is definitely the behavior on master; `shuffle=False` isn't supported ([_split.py#L493][1], [_split.py#L477][2]), and `blockwise` is False for DataFrames, and True for Arrays. I think that's the right behavior.

I need to investigate `blockwise` some; I'm not sure if there's a mixup in the documentation or implementation.

[1]:https://github.com/dask/dask-ml/blob/fc6a04aec4517abca318bc7e44e99f8e540b6fe6/dask_ml/model_selection/_split.py#L493
[2]:https://github.com/dask/dask-ml/blob/fc6a04aec4517abca318bc7e44e99f8e540b6fe6/dask_ml/model_selection/_split.py#L477

TomAugspurger: I think blockwise is backwards for DataFrame in the documentation & implementation. It only supports the equivalent of `blockwise=True` (within-block shuffling).

dask_ml/model_selection/_incremental.py
```diff
@@ -635,7 +641,10 @@ async def _fit(self, X, y, **fit_params):
...
+        )
```
TomAugspurger, line 647:
```suggestion

        X_train, X_test, y_train, y_test = self._get_train_test_split()
```

TomAugspurger: `_get_train_test_split` wasn't using `shuffle`. It's OK to just use the hard-coded value.

dask_ml/model_selection/_incremental.py
stsievert: > I think blockwise is backwards for DataFrame in the documentation & implementation. It only supports the equivalent of blockwise=True (within-block shuffling).

I'm a little confused. Should the implementation support shuffling between chunks or within chunks? I'd like to see `train_test_split` support shuffling between chunks (`blockwise=False`). Currently, `train_test_split` only supports shuffling within chunks.

TomAugspurger: > Should the implementation support shuffling between chunks or within chunks? 

Ideally yes. But that's distinct from this PR.

TomAugspurger:
Thanks! Apologies for the incorrect suggestion.

Issue #700, opened 2020-07-15
Add datasets to API documentation


TomAugspurger:
Right now `dask_ml.datasets` isn't included in the API docs at https://ml.dask.org/modules/api.html. Need to add them to `docs/source/modules/api.rst`.

raybellwaves:
Closed via #702 right?

TomAugspurger:
Yep, thanks.

PR #702, opened 2020-07-20
Docs: [Issue/700]: Add datasets to API documentation

angulartist:
Issue reference : https://github.com/dask/dask-ml/issues/700

This PR:
- includes dask_ml.datasets module to the API documentation

Comment:
I see a make_classification() function that is not documented. I didn't include it in this PR for this reason. Can you confirm?

### DIFF
--- a/docs/source/modules/api.rst
+++ b/docs/source/modules/api.rst
@@ -264,3 +264,20 @@ Classification Metrics
 
    train
    predict
+
+:mod:`dask_ml.datasets`: Datasets
+======================================================
+
+dask-ml provides some utilities for generating toy datasets.
+
+.. automodule:: dask_ml.datasets
+
+.. currentmodule:: dask_ml.datasets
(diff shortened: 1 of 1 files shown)


docs/source/modules/api.rst
```diff
@@ -264,3 +264,20 @@ Classification Metrics
...
+
+.. currentmodule:: dask_ml.datasets
+
+.. autosummary::
+   :toctree: generated/
+
+   make_counts
+   make_blobs
+   make_regression
+   make_classification_df
```
TomAugspurger, line 283: Can you add `make_classification` here? It doesn't have a docstring yet (would be a good addition) but we'll want to make sure that it's included when one is added.

angulartist: Sure!

TomAugspurger:
Thanks!

Issue #689, opened 2020-07-01
CountVectorizer for text preprocessing


jrdzha:
I understand why hash seems to be a better solution for distributed text preprocessing, but I also need a way to make my features human-readable. It seems like spark has a [CountVectorizer](https://spark.apache.org/docs/latest/ml-features#countvectorizer). Would it be possible to implement one for dask-ml?

TomAugspurger:
Yeah, that should be relatively straightforward, especially if you provide a `vocabulary`: https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html, which will tell us the output shape without having to look at the data.

If a `vocabulary` isn't provided then I think we'd need to make two passes over the data

1. Determine all the unique terms in the documents
2. Transform the values into the appropriate place in the discovered vocuabulary

Are you interested in working on this?

mrocklin:
@lesteve you might find this issue interesting.

jrdzha:
@TomAugspurger Yes this is something I'd be interested in contributing. I'm not familiar with dask internals, so I'll need to see...

TomAugspurger:
Thanks. Let me know if you run into any stumbling blocks.

My guess is that the end result will

1. Somehow determine the `vocabulary` (the trickiest part I think)
2. Use input.map_partitions(est.transform, ...) with an
sklearn.feature_extreaction.text.CountVectorizer with that vocabulary

stsievert:
> Somehow determine the `vocabulary` (the trickiest part I think)

Why won't the naive solution work?

``` python
>>> import dask.array as da
>>> def get_vocab(dask_array):
...     return da.unique(y)
>>> x = [["this", "is"], ["a", "sentence"]]
>>> y = da.from_array(x, chunks=(1, -1))
>>> vocab = get_vocab(y)
>>> vocab.compute()
array(['a', 'is', 'sentence', 'this'], dtype='<U8')
```

Doesn't `compute` have to be called to use Scikit-learn's CountVectorizer?

jrdzha:
@stsievert I believe that unique is not good enough, since that just counts unique strings. CountVectorizer is to break strings into n-gram words. For an example, "this is a sentence" can be broken into "this", "is", "a", "sentence" or "this is" "a sentence". So there is some non-arbitrary computation that needs to be done on each string record in a array.

stsievert:
@jrdzha  I think this solution would work to compute the vocabulary:

``` python
# input
[...]

PR #705, opened 2020-07-22
Count vectorizer (with Actors)

TomAugspurger:
This has an implementation of CountVectorizer.

The primary difficulty is learning the vocabulary from the data. The basic idea is to just use scikit-learn's CountVectorizer on each partition of a bag. That'll give us a list of vocabularies (dictionaries) that we need to merge.

A naive implementation would just pass the `vocabularies` around as a dictionaries in the task graph. This isn't great since these can get pretty large. For the news dataset, scikit-learn learns a vocabulary that's 18 MB.

I've prototyped something using distributed's Actors (cc @mrocklin if you're interested in seeing them in use). Timings:

Actors? | fit | compute |
------- | --- | ------- |
Yes | 2.3 | 3.4     |
No | 5.2 | 4.2

I'll probably spend a bit more time to make this work without actors again. Regardless, I'll be writing up an example / blog post.

Closes https://github.com/dask/dask-ml/issues/689

### DIFF
--- a/dask_ml/feature_extraction/text.py
+++ b/dask_ml/feature_extraction/text.py
@@ -66,7 +66,6 @@ class _BaseHasher(sklearn.base.BaseEstimator):
 
 
 
-
 class HashingVectorizer(_BaseHasher, sklearn.feature_extraction.text.HashingVectorizer):
     # explicit doc for Sphinx
     __doc__ = sklearn.feature_extraction.text.HashingVectorizer.__doc__
@@ -116,6 +115,25 @@ class FeatureHasher(_BaseHasher, sklearn.feature_extraction.text.FeatureHasher):
         return sklearn.feature_extraction.text.FeatureHasher
 
 
+class Vocabulary:
+    def __init__(self, vocabulary=None):
+        if vocabulary is None:
+            self.vocabulary = set()
+            self.fixed_vocabulary = True
+        else:
+            self.vocabulary = vocabulary
+
+    def update(self, obj):
+        assert not self.fixed_vocabulary
+        self.vocabulary |= obj
+
+    def finalize(self):
+        if self.fixed_vocabulary:
+            return self.vocabulary
+        else:
+            return {key: i for i, key in enumerate(sorted(self.vocabulary))}
+
+
 class CountVectorizer(sklearn.feature_extraction.text.CountVectorizer):
     # Potential issues:
     # 1. The vocabularies might be somewhat large.
@@ -129,24 +147,28 @@ class CountVectorizer(sklearn.feature_extraction.text.CountVectorizer):
         params = self.get_params()
         vocabulary = params.pop("vocabulary")
 
-        client = 
+        client = get_client()
 
         if self.vocabulary is not None:
             # Case 1: Just map transform.
             fixed_vocabulary = True
+            vocabulary_ = client.submit(Vocabulary, vocabulary=vocabulary, actor=True)
+            vocabulary_actor = vocabulary_.result()
+
             result = raw_documents.map_partitions(_count_vectorizer_transform,
-                                                  dask.delayed(vocabulary), params)
+                                                  vocabulary_actor, params)
         else:
             fixed_vocabulary = False
             # Case 2: learn vocabulary from the data.
-            vocabularies = raw_documents.map_partitions(_build_vocabulary, params)
+            vocabularies = raw_documents.map_partitions(_build_vocabulary, vocabulary, params)
             # compute and merge.
-            vocabularies = dask.compute(vocabularies)  # List[Set[str]]
+            dask.compute(vocabularies)  # List[Set[str]]
+            vocabulary = 
             # maybe the merge should be done as a task in the cluster.
             vocabulary = {key: i for i, key in enumerate(set(itertools.chain.from_iterable(vocabularies)))}
             # same as before
             result = raw_documents.map_partitions(_count_vectorizer_transform,
-                                                  dask.delayed(vocabulary), params)
+                                                  vocabulary, params)
 
         meta = scipy.sparse.eye(0, format="csr", dtype=self.dtype)
         result = build_array(result, len(vocabulary), meta)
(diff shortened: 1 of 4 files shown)


tests/feature_extraction/test_text.py
```diff
@@ -107,3 +107,17 @@ def test_correct_meta():
     assert scipy.sparse.issparse(result._meta)
     assert result._meta.dtype == "float64"
     assert result._meta.shape == (0, 0)
+
+
+def test_count_vectorizer():
+    # TODO: gen_cluster, pickle futures, issue.
+    from distributed import Client
+
+    with Client():
```
mrocklin, line 116: Either gen_cluster (recommended) or the `c, s, a, b` fixtures in utils_test which support the sync API

dask_ml/feature_extraction/text.py
```diff
@@ -108,3 +110,114 @@ class FeatureHasher(_BaseHasher, sklearn.feature_extraction.text.FeatureHasher):
...
+            # Case 2: learn vocabulary from the data.
+            vocabularies = raw_documents.map_partitions(
+                _build_vocabulary, vocabulary_actor, params
+            )
+            # compute and merge.
+            dask.compute(vocabularies)  # List[Set[str]]
+            # same as before
+            result = raw_documents.map_partitions(
+                _count_vectorizer_transform, vocabulary_actor, params
+            )
```
mrocklin, line 176: So if we're doing two passes anyway, why use actors here?  Why not do something like a tree reduction?  

I see in your original post there is some performance difference.  Can I ask what approach you tried previously?

TomAugspurger: I pushed a commit making actors a toggle. https://github.com/dask/dask-ml/pull/705/files#diff-67ef0189a95e65d0765c18596857a748R269-R281 has the non-actor version. Roughly:

1. Build a vocabulary for each partition.
2. Collect all the vocabularies on the client.
3. Merge the vocabularies.

I don't immediately see where a tree reduction is helpful (but I don't think I've ever implemented one before). I think that our "tree" structure is very flat here: A `List[Set[str]]` that we need to merge into a single `Set[str]` with a set union.

I'll poke around the dask repo. I know I've seen references to them there.

dask_ml/feature_extraction/text.py
```diff
@@ -108,3 +112,253 @@ class FeatureHasher(_BaseHasher, sklearn.feature_extraction.text.FeatureHasher):
...
+        else:
+            fixed_vocabulary = False
+            # Case 2: learn vocabulary from the data.
+            vocabularies = raw_documents.map_partitions(
+                _build_vocabulary_no_actor, params
+            )
+
+            # compute
+            vocabulary = _merge_vocabulary(*vocabularies.to_delayed()).compute()
```
mrocklin, line 272: Do we need to bring this local and call compute, or can we delay the `_merge_vocabulary` call?  It looks like we might need to bring this locally eventually, but hopefully we can do all of this computation at once, rather than pull things locally, and then push them back and start computation again.

TomAugspurger: Still futzing with this, but I think it's unavoidable.

1. We need the length to get the output shape right (number of columns)
2. We need the concrete result eventually to provide to the user as vocabulary_

mrocklin: Right, but we don't need to pull down and then push back up.  Rather, I recommend that we keep everything delayed, and then when we're done we pull things down rather than re-scatter them back up.  This makes it so that this can be run locally without the distributed scheduler, and it reduces communication back and forth.

```suggestion
            remote_vocabulary = _merge_vocabulary(*vocabularies.to_delayed())
```

mrocklin: Ah, I see we don't call compute after this.

Maybe then we do

```python
remote_vocabulary = remote_vocabulary.persist()  # compute, and keep in remote memory
local_vocabulary = remote_vocabulary.compute()
```

dask_ml/feature_extraction/text.py
```diff
@@ -108,3 +112,253 @@ class FeatureHasher(_BaseHasher, sklearn.feature_extraction.text.FeatureHasher):
...
+            fixed_vocabulary = True
+            result = raw_documents.map_partitions(
+                _count_vectorizer_transform_no_actor, dask.delayed(vocabulary), params
+            )
+        else:
+            fixed_vocabulary = False
+            # Case 2: learn vocabulary from the data.
+            vocabularies = raw_documents.map_partitions(
+                _build_vocabulary_no_actor, params
+            )
```
mrocklin, line 269: For tree reductions I would expect that `apply_concat_apply` probably has a method that lets you supply functions on how to combine the results of intermediate aggregated chunks.  

Alternatively, this thing might already be a `bag`, in which case I would hope that the bag.unique method would work ok.

mrocklin: Ah, Bag.distinct

dask_ml/feature_extraction/text.py
```diff
@@ -108,3 +112,253 @@ class FeatureHasher(_BaseHasher, sklearn.feature_extraction.text.FeatureHasher):
...
+        return result
+
+    def _transform_no_actor(self, raw_documents):
+        params = self.get_params()
+        vocabulary = params.pop("vocabulary")
+        del params["use_actors"]
+
+        if vocabulary is None:
+            check_is_fitted(self, "vocabulary_")
+            vocabulary = self.vocabulary_
```
mrocklin, line 301: References like these could be `self.remote_vocabulary_` if we wanted.

dask_ml/feature_extraction/text.py
```diff
@@ -108,3 +112,253 @@ class FeatureHasher(_BaseHasher, sklearn.feature_extraction.text.FeatureHasher):
...
+            else:
+                (remote_vocabulary,) = client.scatter([vocabulary], broadcast=True)
+
+            # same as before
+            result = raw_documents.map_partitions(
+                _count_vectorizer_transform_no_actor, remote_vocabulary, params
+            )
+
+        meta = scipy.sparse.eye(0, format="csr", dtype=self.dtype)
+        result = build_array(result, len(vocabulary), meta)
```
mrocklin, line 286: we could use `np.nan` rather than `len(vocabulary)` here if we wanted to keep everything lazy.  

All of this might be too much though.  If the vocabulary is small then it probably makes sense to do a separate pass and a full compute call.  I think that what you have here is probably better than what I'm suggesting, except for the `Bag.distinct` call.

TomAugspurger: Going to pass on this. Getting the number of columns right is pretty important / necessary for using this in a pipeline.

TomAugspurger:
Ok, I've removed the use of actors. They might provide slightly faster performance in some cases, but once I was using regular old dask correctly (not moving data around unnecessarily), the difference is negligible for my test case.

I'll probably write this up in a blog post, just a reminder of how thinking about where your data is relative to your compuation can be important for performance.

mrocklin:
Cool.  I'm glad to hear that you were able to find a high performing
solution that was also simple.

Issue #718, opened 2020-08-05
Long tests


mrocklin:
I was looking at a recent test run and saw this

```python
1000.12s call     tests/model_selection/test_incremental.py::test_basic
17.06s call     tests/model_selection/dask_searchcv/test_model_selection.py::test_as_completed_distributed
10.88s call     tests/test_kmeans.py::test_check_estimator
8.63s call     tests/model_selection/test_split.py::test_20_newsgroups
8.35s call     tests/model_selection/test_incremental.py::test_search_basic
6.90s call     tests/model_selection/dask_searchcv/test_model_selection.py::test_scheduler_param[multiprocessing-4]
6.30s call     tests/model_selection/test_incremental.py::test_warns_scores_per_fit
5.72s call     tests/test_incremental.py::test_score
5.48s call     tests/model_selection/test_incremental.py::test_small
4.98s call     tests/model_selection/dask_searchcv/test_model_selection.py::test_scheduler_param_distributed
```

It looks like we have a `timeout=1000` in one of our tests.  

Running this locally test_basic does seem to hang for me for quite a while

mrocklin:
Ah, this is bad on master, but not this bad.  I think that this hang is particularly an issue in @stsievert 's recent PR

PR #713, opened 2020-07-26
DOC, TST: Wrapping of Keras models

stsievert:
**What does this PR implement?**
It documents and tests wrapping Keras models.

**Reference issues/PRs**
This inherits from #699.

PR #720, opened 2020-08-05
TST MAINT: reduce timeout

stsievert:
While writing Dask-ML model selection tests, I blindly copy/pasted old code with `timeout=1000` or `timeout=5000`. This PR removes those statements.

**Reference issues/PR**
This will close (most of) #718


TomAugspurger:
Thanks. Closing out #718 would require finding the cause of the hangs?

stsievert:
I think so. Having long timeouts is unrelated to the issue I think.

#713 is also hanging.

PR #725, opened 2020-08-12
ENH: add mean_squared_log_error

raybellwaves:
Added mean squared log error to the regression metrics.

Used https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/metrics/_regression.py#L354 as a template

### DIFF
--- a/dask_ml/metrics/regression.py
+++ b/dask_ml/metrics/regression.py
@@ -122,8 +122,8 @@ def mean_squared_log_error(
 ) -> ArrayLike:
 
     result = mean_squared_error(
-        da.log1p(y_true),
-        da.log1p(y_pred),
+        np.log1p(y_true),
+        np.log1p(y_pred),
         sample_weight=sample_weight,
         multioutput=multioutput,
         compute=compute,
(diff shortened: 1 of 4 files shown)


dask_ml/metrics/regression.py
```diff
@@ -110,3 +110,22 @@ def r2_score(
...
+def mean_squared_log_error(
+    y_true: ArrayLike,
+    y_pred: ArrayLike,
+    sample_weight: Optional[ArrayLike] = None,
+    multioutput: Optional[str] = "uniform_average",
+    compute: bool = True,
+) -> ArrayLike:
+
+    result = mean_squared_error(
+        da.log1p(y_true),
```
TomAugspurger, line 125: Slight preference for using `np.log1p` here.

```suggestion
        np.log1p(y_true),
```

dask_ml/metrics/regression.py
```diff
@@ -110,3 +110,22 @@ def r2_score(
...
+        da.log1p(y_pred),
```
TomAugspurger, line 126:
```suggestion
        np.log1p(y_pred),
```

TomAugspurger:
Thanks!

PR #728, opened 2020-08-19
Add a 'decision_function()' method to the 'LogisticRegression' class.

wfondrie:
I noticed that the `dask_ml.linear_model.LogisticRegression` class lacked a `decision_function()` method like is implemented in the corresponding [scikit-learn API](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#sklearn.linear_model.LogisticRegression.decision_function).

This PR adds a `decision_function()` method and updates one corresponding test.


TomAugspurger:
Thanks. The CI failures are known and unrelated.

Issue #736, opened 2020-09-14
Update svd tests for svd_flip in dask


eric-czech:
Hey @TomAugspurger, I'm trying to figure out the best way to integrate https://github.com/dask/dask/pull/6613.

The svd, pca, and incremental pca tests are all broken for any dask >= 2.26.0 install because the svd_flip in dask is now different from scikit-learn.

Ideally I could tweak the tests a little to deal with different signs in results but I didn't realize how often they are behind a few levels of abstraction, e.g. `assert_estimator_equal`: https://github.com/dask/dask-ml/blob/b94c587abae3f5667eff131b0616ad8f91966e7f/tests/test_pca.py#L25-L30

My first thought to fix this was to do something like this instead:

```python
def test_basic(): 
     # This would also need to exist in TruncatedSVD and IncrementalPCA
     a = dd.PCA(sklearn_compatible=True)  # This will preserve the original behavior, but default to False?
     b = sd.PCA() 
     a.fit(dX) 
     b.fit(X) 
     assert_estimator_equal(a, b) 
```

I dislike muddling the signatures with that parameter, but I can't think of a much better solution.  Another possibility would be to modify all the tests such that any checks that compare SVD results (or something derived from them) to scikit-learn are sign-independent.  That would make the tests a bit uglier/more complicated but is probably ideal.

Has this come up before where you have some functionality that diverges from scikit-learn for performance reasons but still want to preserve compatible behavior in unit tests?

eric-czech:
FYI https://github.com/dask/dask-ml/pull/737 is a fix for TruncatedSVD that takes this approach from above:

> Another possibility would be to modify all the tests such that any checks that compare SVD results (or something derived from them) to scikit-learn are sign-independent

I'll keep looking at how to do the same in PCA/IncrementalPCA tests.

eric-czech:
Closing since the changes in https://github.com/dask/dask/pull/6658 ensure sign stability after small changes to the input data, and the incremental PCA tests now work without changes.

Issue #681, opened 2020-06-17
PCA with dataframe can't  real explained_variance_ratio_ except nan


sosofun:
**What happened**:

when use dask_ml.decomposition.PCA fit the dask.dataframe， then will get the nan of pca Attributes  pca.explained_variance_ratio_/explained_variance_  ; maybe can't get  the dataframe shape lead to the error.

**What you expected to happen**:
[0.99244289 0.00755711]

**Minimal Complete Verifiable Example**:
[nan nan]

```python
# Put your MCVE code here
import dask.dataframe as dd
import dask.array as da
import numpy as np
from dask_ml.decomposition import PCA

X = np.array([[-1, -1], [-2, -1], [-3, -2], [1, 1], [2, 1], [3, 2]])
dX = da.from_array(X, chunks=X.shape)
pca = PCA(n_components=2)
pca.fit(dX)
print(pca.explained_variance_ratio_)
#the result is [0.99244289 0.00755711]


df = dd.from_dask_array(dX, columns=['a','b'])
pca = PCA(n_components=2, svd_solver='full')
pca.fit(df)
print(pca.explained_variance_ratio_)
#the result is [nan nan]
```

**Anything else we need to know?**:

**Environment**:
python 3.6
dask                      2.18.1
dask-glm                  0.2.0
dask-ml                   1.5.0
dask-xgboost              0.1.10

- Dask version:
- Python version:
- Operating System:
- Install method (conda, pip, source):

sosofun:
have fix it  with :https://github.com/dask/dask-ml/pull/682/commits/ff66a196b862d569c6895db44b21dbdcf71ace5f

zexuan-zhou:
I encountered the similar issue. Looking forward to the solution to be merged

TomAugspurger:
The PR at https://github.com/dask/dask-ml/pull/682 has stalled. Perhaps we give @sosofun a few days to respond, and if they haven't then you make a new PR @zexuan-zhou?

zexuan-zhou:
> The PR at #682 has stalled. Perhaps we give @sosofun a few days to respond, and if they haven't then you make a new PR @zexuan-zhou?

Sounds great. Thank you!

TomAugspurger:
@zexuan-zhou I just closed #682, if you're interested in working on it.

zexuan-zhou:
> @zexuan-zhou I just closed #682, if you're interested in working on it.

[...]

PR #737, opened 2020-09-15
Update svd tests for svd_flip

eric-czech:
This is a partial fix for https://github.com/dask/dask-ml/issues/736.

I think this is probably the best approach: https://github.com/dask/dask-ml/compare/master...eric-czech:svd_flip?expand=1#diff-151bb55aba85377c700e75a189fa2256

That only fixes TruncatedSVD at the moment, but it seems like a reasonable way to account for the differences now.  A function like `flip_vector_signs` could be applied to PCA components, loadings or scores to point all the vectors in the same direction before comparing them.  I'll go forward with that and try to fix as much as I can in the current build unless you would prefer something else @TomAugspurger.

FYI this PR also removes several xfail cases that now work for short-fat arrays.

### DIFF
--- a/ci/environment-3.6.yaml
+++ b/ci/environment-3.6.yaml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - black
+  - black==19.10b0
   - coverage
   - dask ==2.4.0
   - dask-glm >=0.2.0
--- a/ci/environment-3.7.yaml
+++ b/ci/environment-3.7.yaml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - black
+  - black==19.10b0
   - coverage
   - codecov
   - dask
--- a/ci/environment-3.8.yaml
+++ b/ci/environment-3.8.yaml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - black
+  - black==19.10b0
   - coverage
   - codecov
   - dask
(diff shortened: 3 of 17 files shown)


eric-czech:
This now includes the required changes in PCA and it's tests as well as a fix for https://github.com/dask/dask-ml/issues/681.  I needed to unpack what was going on there to make some of these tests work.

TomAugspurger:
Thanks, this looks great. I'll get the unrelated CI failure sorted out.

eric-czech:
Ok thanks @TomAugspurger!

IncrementalPCA, PCA, and TruncatedSVD should be ok now.  The biggest wrinkle I hit was that this new svd_flip isn't stable when not fit on the same data (but it does at least make results the same for different chunkings / operating systems).  I left IncrementalPCA as-is because of that.  It still uses the sklearn svd_flip.

TomAugspurger:
Great, thanks.

To avoid creating merge conflicts, I'm going to push the other CI fixups to this branch.

TomAugspurger:
The Windows timeout is a known flaky test that IIRC Scott has been investigating.

Thanks @eric-czech.

TomAugspurger:
@eric-czech let me know if / when you'd benefit from a Dask-ML release. I'm happy to do one at any time.

eric-czech:
Ok thanks @TomAugspurger!  I'm out for the rest of the week but I wanted to investigate the wrinkle I mentioned in https://github.com/dask/dask-ml/pull/737#issuecomment-692959949 a bit more when I get back.  Also https://github.com/dask/dask-ml/issues/740.  I think that would likely be enough for a v1 in our project.

PR #743, opened 2020-09-21
Add compute flag to TruncatedSVD

### LINKED ISSUE #740: Allow lazily evaluated TruncateSVD results
I don't see any reason why the pca-specific parameters like these need to be computed immediately rather than leaving them as dask arrays: https://github.com/dask/dask-ml/blob/b94c587abae3f5667eff131b0616ad8f91966e7f/dask_ml/decomposition/truncated_svd.py#L181-L187

Is that true @TomAugspurger?  I'd like to add a `compute` flag if that's ok.  I'm guessing there is nothing in scikit-learn that would ever rely on those specific attributes.

eric-czech:
Closes #740

### DIFF
--- a/dask_ml/decomposition/truncated_svd.py
+++ b/dask_ml/decomposition/truncated_svd.py
@@ -8,7 +8,13 @@ from ..utils import svd_flip
 
 class TruncatedSVD(BaseEstimator, TransformerMixin):
     def __init__(
-        self, n_components=2, algorithm="tsqr", n_iter=5, random_state=None, tol=0.0
+        self,
+        n_components=2,
+        algorithm="tsqr",
+        n_iter=5,
+        random_state=None,
+        tol=0.0,
+        compute=True,
     ):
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks. Could you add a small sanity check for this? Just something like

```python
svd = TruncatedSVD(comptue=False)
svd.fit(X)
assert isinstance(self.components_, da.Array)
```

eric-czech:
> Could you add a small sanity check for this? Just something like

Ah whoops, yes added one at https://github.com/dask/dask-ml/pull/743/commits/78b69fa855e7d0191c3c4905006aafa49109167b#diff-151bb55aba85377c700e75a189fa2256R72-R83.

eric-czech:
Hey @TomAugspurger, this is ready if you get a chance.  Between this and https://github.com/dask/dask-ml/pull/742, we'd also be ready to use a new dask-ml release if you find a convenient time to cut one.

TomAugspurger:
Thanks!

Issue #746, opened 2020-09-30
Train Test Split Unexpected Behavior


skirmer:
**What happened**:
`_validate_shuffle_split` is not setting the size of test correctly when only `train_size` is included. The reverse is also true, train size will be wrong if you only set `test_size`.

I passed a dask array to `train_test_split` from `dask-ml`. The `_validate_shuffle_split` method in this function is actually imported from sklearn. (It's possible that this may mean the issue is better posed to sklearn but as it's affecting dask users you might be interested.) 

The error you should see when running the reprex is: `"ValueError: With n_samples=1, test_size=0.30000000000000004 and train_size=0.7, the resulting train set will be empty. Adjust any of the aforementioned parameters."` This error would be entirely accurate if it said `test_size = 0.3` but it's clearly not right as it is. If you manually set `train_size` and `test_size` both, then the error is correct and behavior is fine.

The smallest specified value that shows an issue as far as I can tell is .66.

Given the incredibly small decimal error on the split default, if you run correct data the error doesn't show itself. However, at scale it eventually results in the test set being microscopically larger than it should be, so this error surfaced in my dask use case because the data was large enough for the test size to calculate out to one extra observation. 

**What you expected to happen**:
dask-ml version of `train_test_split` would default the size of test to 0.3 exactly, not 0.30000000000000004.

**Minimal Complete Verifiable Example**:

```python
from dask_ml.model_selection import train_test_split
import dask.dataframe as dd
from sklearn import datasets
import pandas as pd
import numpy as np

# Grab sample dataset
iris = datasets.load_iris()
iris_pd = pd.DataFrame(data= np.c_[iris['data'], iris['target']],
                     columns= iris['feature_names'] + ['target'])

iris_pd['col1_madeup'] = iris_pd['sepal length (cm)'].astype(str)
iris_pd['col2_madeup'] = iris_pd['sepal width (cm)'].astype(str)

iris_dd = dd.from_pandas(iris_pd, npartitions=10)

def preprocess(dns: dd.DataFrame) -> dd.DataFrame:
    dns['target'] = 0
    dns = dns.groupby(["target", "col2_madeup"]).col1_madeup.apply(list)
    dns = dns.reset_index()
    return dns

iris_dd = preprocess(iris_dd)
iris_da = iris_dd.to_dask_array().compute_chunk_sizes()

x_train, x_test, y_train, y_test = train_test_split(
    iris_da[:,1],
    iris_da[:,2],
    train_size=0.7,
[...]

PR #747, opened 2020-09-30
Bugfix: adding rounding to calculation of unspecified sample size

skirmer:
A proposed solution to #746. I ran tests like this to confirm it works across expected values:

```
for i in np.arange(0.01, 1.0, 0.0001):
    print(round(1-i, 6))
```

### DIFF
--- a/dask_ml/model_selection/_split.py
+++ b/dask_ml/model_selection/_split.py
@@ -424,9 +424,9 @@ def train_test_split(
         test_size = 0.1
 
     if train_size is None and test_size is not None:
-        train_size = 1 - test_size
+        train_size = round(1 - test_size, 6)
     if test_size is None and train_size is not None:
-        test_size = 1 - train_size
+        test_size = round(1 - train_size, 6)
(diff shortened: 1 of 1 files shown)


jsignell:
This seems like a fine and concise solution to me.

dask_ml/model_selection/_split.py
```diff
@@ -424,9 +424,9 @@ def train_test_split(
         test_size = 0.1
 
     if train_size is None and test_size is not None:
-        train_size = 1 - test_size
+        train_size = round(1 - test_size, 6)
```
hristog, line 427: Perhaps, the number 6 could be promoted to a named variable to avoid copy-and-paste bugs upon possible updates to this.

Also, I think for some users, the value 6 wouldn't be as effective, as it would for others. Perhaps, this warrants a consideration for introducing a configurable parameter. If such parameter is absent, some default value (eg, 6, as it has been the choice here) could be used as a fallback.

Issue #748, opened 2020-10-16
Allowing HyperbandSearchCV to also work with non dask arrays


gioxc88:
I have a custom packages which uses a custom data-structure similar to `pandas DataFrame` but that also carry additional metadata. Let's call it `CustomFrame`.

I wrapped all estimator which share the `sklearn` API such as `StandardScaler`, `Lasso`, `XGBRegressor` etc.. to make them able to handle this `CustomFrame`.

This works well with `dask` implementation of `GridSearchCV` or `RandomizedSearchCV` as they  just pass the `X` and `y` to the underlying estimators that, as I said, can correctly handle my `CustomFrame`.

I wanted to do the same with HyperbandSearchCV but I found this much more difficult, if not impossible without rewriting most of it, because the implementation is very different as it relies on Future objects and async await. 

To be more precise at the moment it doesn't even work with `pandas DataFrames`. If you don't specify the `test_size`, the earliest time it fails is in `dask_ml.model_selection._incremental` at line 569 `test_size = min(0.2, 1 / X.npartitions)` because `DataFrame` does not have the attribute `npartitions`.

I would like to ask if you could consider adding the possibility to also handle also other `ArrayLike` objects the same way is already possible using `GridSearchCV` or `RandomizedSearchCV`

Many thanks
Gio

stsievert:
I think specifying `HyperbandSearchCV(..., test_size=0.2)` will resolve this problem in the short term. The method underlying HyperbandSearchCV works with Dask Dataframes, so if your `CustomFrame` inherits from `dask.dataframe.Dataframe` it should work:

https://github.com/dask/dask-ml/blob/7e11ce1505a485104e02d49a3620c8264e63e12e/dask_ml/model_selection/_incremental.py#L181-L182

stsievert:
What updates do you have on this @gioxc88? Did my solution work, or is there some other issue?

gioxc88:
> What updates do you have on this @gioxc88? Did my solution work, or is there some other issue?

Sorry for my late reply. I believe the problem is more complex than that and I didn't have time to test it because I have been working on something else for the past week but I will try to test it and get back to you.

Ideally I don't want to inherit from dask dataframe if possible. In general I believe that this hyperband search should have the same api as GridSearch. There is no need to force the user to use Dask Arrays

[...]

PR #751, opened 2020-10-28
support for non dask arrays for HyperbandSearchCV

gioxc88:
The main changes in this PR are:

1. change to  the async def _fit function in `dask_ml.model_selection._incremental.py` to allow Hyperband to work with non dask arrays

2. fixed default `test_size` which didn't work with pandas dataframe

### DIFF
--- a/dask_ml/model_selection/_incremental.py
+++ b/dask_ml/model_selection/_incremental.py
@@ -198,33 +198,39 @@ async def _fit(
     else:
         y_test = await client.scatter(y_test)
 
-    # Convert to batches of delayed objects of numpy arrays
-    X_train = sorted(futures_of(X_train), key=lambda f: f.key)
-    y_train = sorted(futures_of(y_train), key=lambda f: f.key)
-    assert len(X_train) == len(y_train)
+    if hasattr(X_train, 'npartitions'):
+        # Convert to batches of delayed objects of numpy arrays
+        X_train = sorted(futures_of(X_train), key=lambda f: f.key)
+        y_train = sorted(futures_of(y_train), key=lambda f: f.key)
+        assert len(X_train) == len(y_train)
 
-    train_eg = await client.gather(client.map(len, y_train))
-    msg = "[CV%s] For training there are between %d and %d examples in each chunk"
-    logger.info(msg, prefix, min(train_eg), max(train_eg))
+        train_eg = await client.gather(client.map(len, y_train))
+        msg = "[CV%s] For training there are between %d and %d examples in each chunk"
+        logger.info(msg, prefix, min(train_eg), max(train_eg))
+
+        def get_futures(partial_fit_calls):
+            """Policy to get training data futures
+
+            Currently we compute once, and then keep in memory.
+            Presumably in the future we'll want to let data drop and recompute.
+            This function handles that policy internally, and also controls random
+            access to training data.
+            """
+            # Shuffle blocks going forward to get uniform-but-random access
+            while partial_fit_calls >= len(order):
+                L = list(range(len(X_train)))
+                rng.shuffle(L)
+                order.extend(L)
+            j = order[partial_fit_calls]
+            return X_train[j], y_train[j]
+    # __addition__ start
+    else:
+        def get_futures(partial_fit_calls):
+            return X_train, y_train
+    # __addition__ end
 
     # Order by which we process training data futures
     order = []
-
-    def get_futures(partial_fit_calls):
-        """Policy to get training data futures
-
-        Currently we compute once, and then keep in memory.
-        Presumably in the future we'll want to let data drop and recompute.
-        This function handles that policy internally, and also controls random
-        access to training data.
-        """
-        # Shuffle blocks going forward to get uniform-but-random access
-        while partial_fit_calls >= len(order):
-            L = list(range(len(X_train)))
-            rng.shuffle(L)
-            order.extend(L)
-        j = order[partial_fit_calls]
-        return X_train[j], y_train[j]
 
     # Submit initial partial_fit and score computations on first batch of data
     X_future, y_future = get_futures(0)
@@ -566,7 +572,8 @@ class BaseIncrementalSearchCV(ParallelPostFit):
         X, y : dask.array.Array
         """
         if self.test_size is None:
-            test_size = min(0.2, 1 / X.npartitions)
+            npartitions = getattr(X, 'npartitions', 1)
+            test_size = min(0.2, 1 / npartitions)
         else:
             test_size = self.test_size
         X_train, X_test, y_train, y_test = train_test_split(


dask_ml/model_selection/_incremental.py
```diff
@@ -198,34 +198,40 @@ async def _fit(
...
-    logger.info(msg, prefix, min(train_eg), max(train_eg))
+    if hasattr(X_train, 'npartitions'):
+        # Convert to batches of delayed objects of numpy arrays
+        X_train = sorted(futures_of(X_train), key=lambda f: f.key)
+        y_train = sorted(futures_of(y_train), key=lambda f: f.key)
+        assert len(X_train) == len(y_train)
+
+        train_eg = await client.gather(client.map(len, y_train))
+        msg = "[CV%s] For training there are between %d and %d examples in each chunk"
+        logger.info(msg, prefix, min(train_eg), max(train_eg))
```
stsievert, line 209: Why won't this code work?

``` python
train_eg =  ...
msg = ...
if len(train_eg):
    logger.info(msg, prefix, min(train_eg), max(train_eg))
```

This avoids `min([])` as mentioned in https://github.com/dask/dask-ml/issues/748#issuecomment-717814887.

Then, I think the if-statement could be moved inside `get_futures`:

``` python
def get_futures(partial_fit_calls):
    if not isinstance(X_train, da.Array):
        return X_train, y_train
    ...  # existing implementation
```

dask_ml/model_selection/_incremental.py
```diff
@@ -198,34 +198,40 @@ async def _fit(
...
 
-    # Convert to batches of delayed objects of numpy arrays
-    X_train = sorted(futures_of(X_train), key=lambda f: f.key)
-    y_train = sorted(futures_of(y_train), key=lambda f: f.key)
-    assert len(X_train) == len(y_train)
-
-    train_eg = await client.gather(client.map(len, y_train))
-    msg = "[CV%s] For training there are between %d and %d examples in each chunk"
-    logger.info(msg, prefix, min(train_eg), max(train_eg))
+    if hasattr(X_train, 'npartitions'):
```
stsievert, line 201: Can this if-statement be changed to `if isinstance(X_train, da.Array)`?

stsievert (commented):
Thanks for the PR!

Could you also implement a short test for this? The custom data/model doesn't have to be complex, something like

``` python

class CustomDataLoader:
    def __len__(self):
        return 100
    ...

class CustomModel(BaseEstimator):
    def _partial_fit(self, X, y=None, **kwargs):
        assert isinstance(X, CustomDataLoader)
        return self

    fit = partial_fit = _partial_fit

    def score(self, X, y=None):
        assert isinstance(X, CustomDataLoader)
        return np.random.uniform()
```

dask_ml/model_selection/_incremental.py
gioxc88: you are right, I was not careful when I red your comment, this would work.

The only reason why I did not implement it that way is because saying `if len(train_eg):` or even shorter `if train_eg` is the same as checking `isinstance(X_train, da.Array)` because `train_eg` will be empty only if `X_train` is not a `da.Array`. This means that in your implementation you are still executing these lines, even if not needed:

```
X_train = sorted(futures_of(X_train), key=lambda f: f.key)
y_train = sorted(futures_of(y_train), key=lambda f: f.key)
assert len(X_train) == len(y_train)

train_eg = await client.gather(client.map(len, y_train))
```

dask_ml/model_selection/_incremental.py
gioxc88: the reason I don't want to restrict the if statement to be `isinstance(X_train, da.Array)` is because, potentially any  custom data structure like the hypotetical `CustomFrame` could have a `dask-like` API by implementing `npartitions` and `__dask_graph__`. 

In this case, for example, in the future I could further extend my `CustomFrame` (I or any other user) to work with data larger than memory and this would still work with `hyperband` even if is not a `da.Array`, because the API is compatible.

With this in mind maybe a good compromise could be:

`if dask.is_dask_collection(X_train)` ?

finally another reason why it might be better not to incorporate the if statement inside the `get_futures` function is because this will force you to check the condition every time you call the function which will not be ideal for `dask.collections` with a large number of partitions

stsievert: > if dask.is_dask_collection(X_train)

I think that's a better choice (though @TomAugspurger might have more input).

dask_ml/model_selection/_incremental.py
stsievert: I think the logging message should be issued if the dataset object has an implementation of `__len__`. The computation of `train_eg` only happens once (`train_eg` is a list of ints).

I don't like two separate definitions of `get_futures`, especially if the both return training data.

dask_ml/model_selection/_incremental.py
TomAugspurger: Yeah, I think we `is_dask_collection` is better.

TomAugspurger (commented):
Even just testing with a NumPy array or pandas DataFrame would be good.

gioxc88:
@stsievert I implemented the changes we discussed.
sorry it took so long but I have been busy with other projects.

Cheers 
Gio

TomAugspurger:
@stsievert do you have another chance to look at this?

dask_ml/model_selection/_incremental.py
```diff
@@ -218,13 +220,20 @@ def get_futures(partial_fit_calls):
...
+            return X_train, y_train
+
         # Shuffle blocks going forward to get uniform-but-random access
         while partial_fit_calls >= len(order):
             L = list(range(len(X_train)))
             rng.shuffle(L)
             order.extend(L)
         j = order[partial_fit_calls]
         return X_train[j], y_train[j]
+    ### end addition ###
```
stsievert, line 233: Style nit: could this comment be removed?

dask_ml/model_selection/_incremental.py
```diff
@@ -218,13 +220,20 @@ def get_futures(partial_fit_calls):
         This function handles that policy internally, and also controls random
         access to training data.
         """
+        if dask.is_dask_collection(y_train):
```
stsievert, line 223: I think this should only check `X_train` because `y_train` is an optional argument.

dask_ml/model_selection/_incremental.py
```diff
@@ -204,11 +204,13 @@ async def _fit(
     assert len(X_train) == len(y_train)
 
     train_eg = await client.gather(client.map(len, y_train))
-    msg = "[CV%s] For training there are between %d and %d examples in each chunk"
-    logger.info(msg, prefix, min(train_eg), max(train_eg))
 
-    # Order by which we process training data futures
-    order = []
+    ### start addition ###
```
stsievert, line 208: Style nit: can this comment be removed?

dask_ml/model_selection/_incremental.py
stsievert: I've done some tracing and it appears that passing `y=None` into `search.fit` isn't supported? That's unrelated to this PR.

stsievert (commented):
I'm surprised this Hyperband/Incremental didn't support Pandas Dataframes already. That shouldn't be surprising – Dask DataFrames weren't supported until #701, and I have verified that the test in this PR fails on v1.7.0.

This PR LGTM, past a couple style nits and one implementation detail.

gioxc88:
Hi @stsievert I agree with you in principle the check should be done on `X_train`.
And sorry for the comments, it was just for me to remember where I made changes.
Of course I can remove them.

tests/model_selection/test_hyperband_non_daskarray.py
```diff
@@ -0,0 +1,22 @@
...
+from dask_ml.model_selection import HyperbandSearchCV
+from dask_ml.datasets import make_classification
+from distributed.utils_test import gen_cluster
+from sklearn.linear_model import SGDClassifier
+
+
+@gen_cluster(client=True)
+def test_pandas(c, s, a, b):
+    X, y = make_classification(chunks=100)
+    X, y = pd.DataFrame(X.compute()), pd.Series(y.compute())
```
stsievert, line 13: Looks like the test is failing in the `.compute` function. Why not use `from sklearn.datasets import make_classification`?

tests/model_selection/test_hyperband_non_daskarray.py
```diff
@@ -0,0 +1,22 @@
...
+import pandas as pd
+
+from dask_ml.model_selection import HyperbandSearchCV
+from dask_ml.datasets import make_classification
+from distributed.utils_test import gen_cluster
+from sklearn.linear_model import SGDClassifier
+
+
+@gen_cluster(client=True)
+def test_pandas(c, s, a, b):
```
stsievert, line 11: Could this test go in `test_hyperband.py`?

tests/model_selection/test_hyperband_non_daskarray.py
gioxc88: It's strange, it was working on my local machine.
Let me try to change it. Maybe it's something to do with how I wrote the test function. I just tried to emulate the structure I found in `test_hyperband.py`

tests/model_selection/test_hyperband_non_daskarray.py
gioxc88: yes let me move it there

gioxc88:
does anyone have any idea of what happened with the tests?
it's like it's stuck 

many thanks

stsievert:
I'm not sure, but it does look there's a linting issue (both black and isort fail):

``` 
Checking black...
black, version 19.10b0
would reformat /home/vsts/work/1/s/dask_ml/model_selection/_incremental.py
would reformat /home/vsts/work/1/s/tests/model_selection/test_hyperband.py
Oh no! 💥 💔 💥
2 files would be reformatted, 98 files would be left unchanged.
Checking black... DONE
Checking isort...
4.3.21
ERROR: /home/vsts/work/1/s/tests/model_selection/test_hyperband.py Imports are incorrectly sorted.
Checking isort... DONE
```

Try making those lint changes and pushing; that will (likely) resolve the issue.

hristog:
I've fixed the linting changes locally (and substantially reduced the computational size of the test), but the `test_pandas` still hangs. I've played around a bit, but haven't been able to identify the root cause yet.

PR #754, opened 2020-11-11
Move documentation CI to GitHub actions

jrbourbeau:
xref https://github.com/dask/dask-ml/issues/752

PR #755, opened 2020-11-11
Update pairwise check for scikit-learn 0.24

jrbourbeau:
Over in https://github.com/dask/dask-ml/pull/754 I ran across a `FutureWarning` in the scikit-learn dev build which caused CI to fail. The `estimator._pairwise` attribute was deprecated in https://github.com/scikit-learn/scikit-learn/pull/18143 in favor of using the `"pairwise"` key in the estimator's tags dictionary.


jrbourbeau:
There are some additional test failure to fix here and some that are related to upstream changes (xref https://github.com/scikit-learn/scikit-learn/issues/18815)

TomAugspurger:
Thanks James. Let's just merge this as is, and check again when https://github.com/scikit-learn/scikit-learn/issues/18815 is merged.

Issue #386, opened 2018-10-05
LogisticRegression's predict_proba(X) returns (n,) instead of (n,2) for binary class


vincentvdp:
Just like [sklearn's precit_proba(X)](http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#sklearn.linear_model.LogisticRegression.predict_proba), [Dask's documentation](https://ml.dask.org/modules/generated/dask_ml.linear_model.LogisticRegression.html#dask_ml.linear_model.LogisticRegression.predict_proba) says it returns an "array-like, shape = [n_samples, n_classes]".
However, it returns an "array-like, shape = [n_samples, 1]", for a binary classifier.
E.g. run this:
```
import numpy as np
from sklearn.linear_model import LogisticRegression
from dask_ml.linear_model import LogisticRegression as DaskLogit

# Data
X = np.random.rand(200,20)
y = np.random.rand(200).round(0) # random binary labels

# Dask 
lrdask = DaskLogit()
lrdask.fit(X, y)
print("Logit Reg 'predict_proba' shape Dask\t", lrdask.predict_proba(X).shape)

# SKLearn
lrsklr = LogisticRegression()
lrsklr.fit(X, y)
print("Logit Reg 'predict_proba' shape SKLearn\t", lrsklr.predict_proba(X).shape)
```

Of course, no information is lost, but this *does* break stuff when used in other code.
E.g. in GridSearchCV's scorer I get:
```
C:\ProgramData\Anaconda3\lib\site-packages\sklearn\metrics\scorer.py in __call__(self, clf, X, y, sample_weight)
    184 
    185                 if y_type == "binary":
--> 186                     y_pred = y_pred[:, 1]
    187                 elif isinstance(y_pred, list):
    188                     y_pred = np.vstack([p[:, -1] for p in y_pred]).T

IndexError: too many indices for array
```

predict_proba(X)'s return value is indeed implemented differently in sklearn and dask-ml:
[sklearn](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/linear_model/base.py):
```
def _predict_proba_lr(self, X):
	...
	return np.vstack([1 - prob, prob]).T
```

[dask-ml](https://github.com/dask/dask-ml/blob/master/dask_ml/linear_model/glm.py):
```
def predict_proba(self, X):
	...
	return sigmoid(dot(X_, self._coef))
```

[...]

PR #760, opened 2020-11-25
#386 LogisticRegression.predict_proba should return (n, 2) for binary

rikturr:
This PR addresses #386 by making sure `LogisticRegression.predict_proba()` returns an array of shape `(n, 2)` to match what scikit-learn does. There are some things currently hard-coded for binary classification only, but it is clear that multi-class is currently not supported so that can be changed when multi-class is implemented.

There is some risk here that if users have downstream code that depends on the current functionality (return shape of `(n,)`), the changes from this PR could break their code.

### DIFF
--- a/dask_ml/linear_model/utils.py
+++ b/dask_ml/linear_model/utils.py
@@ -59,3 +59,13 @@ def add_intercept(X):  # noqa: F811
     if "intercept" in columns:
         raise ValueError("'intercept' column already in 'X'")
     return X.assign(intercept=1)[["intercept"] + list(columns)]
+
+
+@dispatch(np.ndarray)  # noqa: F811
+def lr_prob_stack(prob):  # noqa: F811
+    return np.vstack([1 - prob, prob]).T
+
+
+@dispatch(da.Array)  # noqa: F811
+def lr_prob_stack(prob):  # noqa: F811
+    return da.vstack([1 - prob, prob]).T
--- a/dask_ml/linear_model/glm.py
+++ b/dask_ml/linear_model/glm.py
@@ -15,6 +15,7 @@ from sklearn.base import BaseEstimator
 
 from ..metrics import r2_score
 from ..utils import check_array
+from .utils import lr_prob_stack
 
 _base_doc = textwrap.dedent(
     """\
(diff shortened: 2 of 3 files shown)


rikturr:
I'm not quite sure why some of the checks are failing. It seems [one of them](https://dev.azure.com/dask-dev/dask/_build/results?buildId=1748&view=logs&j=d699f7f7-fbb1-5fe7-7f49-16a980a27dcf&t=47702d4b-eb38-52c1-7644-7bfb10716412&l=18) is due to:

`##[warning]numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject`

but I'm not quite sure what I can change to make the check pass?

TomAugspurger:
That wouldn’t be related to this PR, but indicates that NumPy may be being installed twice? If you’re able to debug that it’d be appreciated.

rikturr:
Looks like test failures have been addressed, besides linting (thanks @jsignell 🎉 ). I think this should be ready to merge if the implementation looks good

jsignell:
> Looks like test failures have been addressed

Might not be quite done yet, but I'm chipping away at it.

jsignell:
Seems like only the keras test is broken and that'll be fixed by #794, so good to merge?

rikturr:
Triggered the tests to run again, so I think all is good now! Can this be merged?

dask_ml/linear_model/utils.py
```diff
@@ -59,3 +59,13 @@ def add_intercept(X):  # noqa: F811
     if "intercept" in columns:
         raise ValueError("'intercept' column already in 'X'")
     return X.assign(intercept=1)[["intercept"] + list(columns)]
+
+
+@dispatch(np.ndarray)  # noqa: F811
```
TomAugspurger, line 64: We can actually just use `np.vstack` now, since Dask implemented [NEP 18](https://numpy.org/neps/nep-0018-array-function-protocol.html).

TomAugspurger (approved):
Thanks! This has dragged on long enough (sorry) so I'm going to merge. But if you want to make a cleanup PR you could simplify things to just call `np.vstack` on the array. Thanks to NEP-18 it works on Dask Arrays too.

PR #767, opened 2020-12-22
Fix pca and svd tests running with Dask 2020.12.0

tomwhite:
https://github.com/dask/dask/pull/6813 introduced incompatible changes - this PR fixes the affected dask-ml tests.

### DIFF
--- a/dask_ml/decomposition/pca.py
+++ b/dask_ml/decomposition/pca.py
@@ -276,8 +276,9 @@ class PCA(sklearn.decomposition.PCA):
             random_state = check_random_state(self.random_state)
             seed = draw_seed(random_state, np.iinfo("int32").max)
             n_power_iter = self.iterated_power
+            iterator = "none" if n_power_iter == 0 else "power"
             U, S, V = da.linalg.svd_compressed(
-                X, n_components, n_power_iter=n_power_iter, seed=seed
+                X, n_components, iterator=iterator, n_power_iter=n_power_iter, seed=seed
             )
         if not DASK_2_26_0:
             U, V = svd_flip(U, V)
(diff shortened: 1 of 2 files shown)


dask_ml/decomposition/pca.py
```diff
@@ -276,8 +276,9 @@ def _fit(self, X):
             random_state = check_random_state(self.random_state)
             seed = draw_seed(random_state, np.iinfo("int32").max)
             n_power_iter = self.iterated_power
+            iterator = "none" if n_power_iter == 0 else "power"
```
eric-czech, line 279: Any reason not to add this param to the `PCA` signature to be consistent with `svd_compressed`?

tomwhite: I wrote this as a minimal bugfix so that clients and tests work again.

What would the `iterator` param default be? Should it be added at the end of the parameter list (for backwards compatibility), or before `n_iter` (for consistency with `svd_compressed`)?

eric-czech: Ok fair enough.

I think the default should 'power' iteration as it was before.  I logged https://github.com/dask/dask/issues/7003 and I'm working on a PR for it right now to make it the same in dask (i.e. like it was before).

For the parameter location, I would vote before `n_iter` even if it breaks positional argument usage.  Maybe @TomAugspurger would have an opinion though?  Tom A, do you try to keep signatures reverse compatible by position?  Is it ok if we add a new kwarg somewhere in the middle?

tomwhite:
Closing, since this is redundant with [dask/dask#7004](https://github.com/dask/dask/pull/7004).

PR #772, opened 2021-01-06
Use `_is_pairwise` with SK024

### LINKED ISSUE #771: grid search with a precomputed kernel matrix fails in SK024
**What happened**:
With scikit-learn=0.24.0, dask_ml grid search fails when the input features are given in the form of a precomputed kernel matrix. 

check `test_grid_search_precomputed_kernel` in: https://github.com/dask/dask-ml/blob/976761ffcd89a6d370b5a29e61d7e80dd5bdd317/tests/model_selection/dask_searchcv/test_model_selection_sklearn.py#L455
```
ValueError: Precomputed matrix must be a square matrix. Input is a 120x180 matrix.
```
This happens with Scikit-learn version: 0.24.0
**Environment**:

- Dask version: 2020.12.0
- Python version: 3.7.6
- Operating System: MacOS

abduhbm:
Fixes #771

### DIFF
--- a/dask_ml/model_selection/_search.py
+++ b/dask_ml/model_selection/_search.py
@@ -55,6 +55,9 @@ from .methods import (
 )
 from .utils import DeprecationDict, is_dask_collection, to_indexable, to_keys, unzip
 
+if SK_024:
+    from sklearn.base import _is_pairwise
+
 logger = logging.getLogger(__name__)
 
 try:
@@ -198,7 +201,7 @@ def build_cv_graph(
     cv = check_cv(cv, y, is_classifier(estimator))
     # "pairwise" estimators require a different graph for CV splitting
     if SK_024:
-        is_pairwise = estimator._get_tags().get("_pairwise", False)
+        is_pairwise = _is_pairwise(estimator)
     else:
         is_pairwise = getattr(estimator, "_pairwise", False)
 


dask_ml/model_selection/_search.py
```diff
@@ -55,6 +55,9 @@
 )
 from .utils import DeprecationDict, is_dask_collection, to_indexable, to_keys, unzip
 
+if SK_024:
+    from sklearn.base import _is_pairwise
```
TomAugspurger, line 59: Do you know if there's a non-private way to check this? cc @thomasjpfan in case you know off the top of your head.

Regardless, I think this would need an `else` branch defining it as the old way (`estimator._get_tags().get("_pairwise", False)`.

abduhbm: @TomAugspurger not sure if there is a non-private way to check it.

I have the `else` branch here: 
https://github.com/abduhbm/dask-ml/blob/62951ab909a0128eda0a67db728ee4062f6fbedf/dask_ml/model_selection/_search.py#L203-L206

thomasjpfan: If `estimator._get_tags().get("_pairwise", False)` has been working for you, then you can use `"pairwise"` as the key instead.

`_is_pairwise` is used to help with the migration from the attribute `_pairwise` to the tag `pairwise`. It's behavior is defined in its docstring: https://github.com/scikit-learn/scikit-learn/blob/2218ec46227c92301ac6837c4a8ae9b8dc5d3960/sklearn/base.py#L819-L828

abduhbm: In `test_grid_search_precomputed_kernel` https://github.com/dask/dask-ml/blob/976761ffcd89a6d370b5a29e61d7e80dd5bdd317/tests/model_selection/dask_searchcv/test_model_selection_sklearn.py#L455 only the `_pairwise` attribute is present and it is not False which leads to raising a deprecation warning while running the test.

TomAugspurger:
OK, thanks. It'd be nice to have a public way to get at this info, but this will do for now.

PR #774, opened 2021-01-11
Fix output dtype of PCA ParallelPost transform

### LINKED ISSUE #773: PCA ParallelPostFit.transform fails with scikit-learn=0.24.0
**What happened**:
PCA.transform with ParallelPost fails when running with scikit-learn=0.24.0

```python
X, y = make_classification(n_samples=1_000, n_features=10, 
                           random_state=0, chunks=100)
clf = ParallelPostFit(estimator=PCA(random_state=0))
clf.fit(*dask.compute(X, y))
clf.transform(X)
ValueError('X has 1 features, but PCA is expecting 10 features as input.')
```

It looks like the `reset` flag in `_validate_data` is causing this error.  
https://github.com/scikit-learn/scikit-learn/blob/cbfe0ede80beca86750ab8113f17c18c8c8042ce/sklearn/decomposition/_base.py#L126

**Environment**:

- Dask version: 2020.12.0
- Python version: 3.7.6
- Operating System: MacOS

abduhbm:
Fix #773

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -204,7 +204,9 @@ class ParallelPostFit(sklearn.base.BaseEstimator, sklearn.base.MetaEstimatorMixi
         X = self._check_array(X)
 
         if isinstance(X, da.Array):
-            return X.map_blocks(_transform, estimator=self._postfit_estimator)
+            xx = np.zeros((1, X.shape[1]), dtype=X.dtype)
+            dt = _transform(xx, self._postfit_estimator).dtype
+            return X.map_blocks(_transform, estimator=self._postfit_estimator, dtype=dt)
         elif isinstance(X, dd._Frame):
             return X.map_partitions(_transform, estimator=self._postfit_estimator)
         else:


dask_ml/wrappers.py
```diff
@@ -204,7 +204,9 @@ def transform(self, X):
         X = self._check_array(X)
 
         if isinstance(X, da.Array):
-            return X.map_blocks(_transform, estimator=self._postfit_estimator)
+            xx = np.zeros((1, X.shape[1]), dtype=X.dtype)
```
TomAugspurger, line 207: Can you use a `np.zeros((0, X.shape[1]), dtype=X.dtype)` here? Or does scikit-learn not like that?

abduhbm: Scikit-learn doesn't like it. 

```
ValueError: Found array with 0 sample(s) (shape=(0, 20)) while a minimum of 1 is required.
```

TomAugspurger:
Thanks.

PR #777, opened 2021-01-25
Fix passing `n_power_iter` for `da.svd_compressed`

abduhbm:
Dask 2020.12.0 introduced `iteration` parameter for `da.array.linalg.svd_compressed`. This PR fixes calling this function in `truncated_svd`


TomAugspurger:
Thanks!

PR #778, opened 2021-01-25
Fix passing `force_all_finite` keyword for `OneHotEncoder.fit` in SK_024

abduhbm:


dask_ml/preprocessing/_encoders.py
```diff
@@ -154,12 +154,19 @@ def fit(
...
+        force_all_finite: bool = True,
+    ):
         X = check_array(
             X, accept_dask_dataframe=True, dtype=None, preserve_pandas_dataframe=True
         )
         if isinstance(X, np.ndarray):
-            return super(OneHotEncoder, self)._fit(X, handle_unknown=handle_unknown)
+            return super(OneHotEncoder, self)._fit(
+                X, handle_unknown=handle_unknown, force_all_finite=force_all_finite
+            )
```
TomAugspurger, line 169: Should this be conditional on the version of scikit-learn? i.e. only pass it for SK_024?

```suggestion
            if SK_024:
                kwargs = dict(force_all_finite=force_all_finite)
            else:
                kwargs = {}
            return super(OneHotEncoder, self)._fit(
                X, handle_unknown=handle_unknown, **kwargs
            )
```

TomAugspurger:
OK, thanks. It'd be nice to have a public way to get at this info, but this will do for now.

TomAugspurger:
Ah, whoops, wrong PR :) I'll fix stuff up.

PR #782, opened 2021-02-01
Change default branch from master to main

jsignell:

PR #784, opened 2021-02-02
Numpy 1.20 compatibility fixes

jsignell:
closes #784 
Hmmm. Maybe I should have based this off of #782 so that the ci would run....

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -28,7 +28,7 @@ jobs:
           make html
 
       - name: Deploy docs
-        if: ${{ github.ref == 'refs/heads/master' && github.event_name != 'pull_request'}}
+        if: ${{ github.ref == 'refs/heads/main' && github.event_name != 'pull_request'}}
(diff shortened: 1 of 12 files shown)


jsignell:
I rebased.

jsignell:
If this passes can probably just merge this on rather than #782

TomAugspurger:
Sounds good. I'll merge when it's green.

jsignell:
Hmm looks like linting is also broken. Would you rather merge this and do the lint fix as a follow up?

TomAugspurger:
Here is fine if you're interested in working on it. If so, I'll leave the mypy stuff to you. I'm totally fine with just disabling mypy for now, we aren't getting a ton of value out of it.

jsignell:
I took a look and I think this should merge first so that CI runs on PRs.

jsignell:
The tests are passing other than lint

TomAugspurger:
Great, thanks for this and for opening the issue!

PR #779, opened 2021-01-26
Add skip for pandas 1.2.0

TomAugspurger:
Haven't researched yet.

PR #788, opened 2021-02-04
Fix simple impute

### LINKED ISSUE #787: `SimpleImputer.fit` strange behavior with `median` and `most_frequent` strategies
**What happened**:
`SimpleImputer.fit` with `median` and `most_frequent` strategies on frames compute different results comparing to `scikit-learn`. 

They should have consistent results with `sklearn.impute.SimpleImputer`.

```python
df = pd.DataFrame({"A": [1, 1, np.nan, np.nan, 2, 2]})
# This should return the smallest value
b = dask_ml.impute.SimpleImputer(strategy="most_frequent", fill_value=None)
b.fit(df)
b.statistics_
>>> A    2.0
>>> dtype: float64

c = sklearn.impute.SimpleImputer(strategy="most_frequent", fill_value=None)
c.fit(df)
c.statistics_
>>> array([1.])
```

With `median`:
```python
df = pd.DataFrame({"A": [1, 1, np.nan, np.nan, 2, 2]})
df = dd.from_pandas(df, 2)
# This should return the smallest value
b = dask_ml.impute.SimpleImputer(strategy="median", fill_value=None)
b.fit(df)
b.statistics_
>>> A    1.0
>>> dtype: float64

c = sklearn.impute.SimpleImputer(strategy="median", fill_value=None)
c.fit(df)
c.statistics_
>>> array([1.5])
```

**Environment**:

- Dask version: 2021.01.1
- Python version: 3.7.6
- Operating System: MacOS

abduhbm:
Fix #787. 
Also related to #779

### DIFF
--- a/dask_ml/impute.py
+++ b/dask_ml/impute.py
@@ -70,12 +70,11 @@ class SimpleImputer(sklearn.impute.SimpleImputer):
         if self.strategy == "mean":
             avg = X.mean(axis=0).values
         elif self.strategy == "median":
-            avg = X.quantile().values
+            avg = [np.median(X[col].dropna()) for col in X.columns]
         elif self.strategy == "constant":
             avg = np.full(len(X.columns), self.fill_value)
         else:
-            avg = [X[col].value_counts().nlargest(1).index for col in X.columns]
-            avg = np.concatenate(*dask.compute(avg))
+            avg = [X[col].mode().min() for col in X.columns]
 
         self.statistics_ = pd.Series(dask.compute(avg)[0], index=X.columns)
 
--- a/dask_ml/cluster/k_means.py
+++ b/dask_ml/cluster/k_means.py
@@ -20,12 +20,12 @@ from ..metrics import (
 )
 from ..utils import _timed, _timer, check_array, row_norms
 
+import numba  # isort:skip (see https://github.com/dask/dask-ml/pull/577)
+
 if SK_024:
     from ._compat import _kmeans_plusplus
 else:
     from ._compat import _k_init as _kmeans_plusplus
-
-import numba  # isort:skip (see https://github.com/dask/dask-ml/pull/577)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


abduhbm:
CI failing for linting issues

dask_ml/impute.py
```diff
@@ -70,12 +70,18 @@ def _fit_frame(self, X):
         if self.strategy == "mean":
             avg = X.mean(axis=0).values
         elif self.strategy == "median":
-            avg = X.quantile().values
+            avg = [np.median(X[col].dropna()) for col in X.columns]
```
TomAugspurger, line 73: I believe this will eagerly compute the values, thanks to `np.median`. Since that's done in a list comprehension, we'd end up executing the graph for `X` once per column. We want to delay computation till the end.

I also think this will end up pulling all the data for a column into a single ndarray, to do the median, which we also want to avoid.

dask_ml/impute.py
```diff
@@ -70,12 +70,18 @@ def _fit_frame(self, X):
...
         elif self.strategy == "constant":
             avg = np.full(len(X.columns), self.fill_value)
         else:
-            avg = [X[col].value_counts().nlargest(1).index for col in X.columns]
-            avg = np.concatenate(*dask.compute(avg))
+            avg = []
+            for col in X.columns:
+                val_counts = X[col].value_counts().reset_index()
+                if isinstance(X, dd.DataFrame):
+                    x = val_counts.to_dask_array(lengths=True)
```
TomAugspurger, line 81: Do we need lengths here? This also triggers a computation.

TomAugspurger (commented):
Thanks. A few questions / comments.

dask_ml/impute.py
abduhbm: How about using `delayed` here? 
``` python
avg = [dask.delayed(np.median(X[col].dropna())) for col in X.columns]
```

dask_ml/impute.py
abduhbm: This is needed to compute chunk sizes ... any suggestion on how to avoid it? Thanks,

PR #786, opened 2021-02-03
Mypy fixes

jsignell:
Closes #785

PR #793, opened 2021-02-07
Dask annotations compatibility

jrbourbeau:
The 2021.02.0 Dask / Distributed release updated how `priority=`, `resources=`, etc. keywords are handled (xref https://github.com/dask/distributed/pull/4406). This PR adds compatibility code to ensure that we use the new annotation system when using newer versions of Dask and Distributed. 

For reference, it looks like the CI failures here are due to unrelated `mypy` issuse which are being handled in https://github.com/dask/dask-ml/pull/786


TomAugspurger:
Thanks!

PR #795, opened 2021-02-13
CI Update

TomAugspurger:
Removes mypy checks. Moves to pre-commit for running.


TomAugspurger:
Not needed now, thanks to Julia's work.

PR #799, opened 2021-02-24
[WIP] Coreset for Kmeans and GaussianMixture clustering

remiadon:
Introduce a Coreset Meta Estimator, that samples a subset of the original data (keeping the original shape) and pass this sample to a scikit-learn estimator

TODO
- [ ] unit tests
- [ ] documentation
- [ ] .fit() returns the scikit-learn estimator, make it returning `self` (the dask meta estimator)
- [ ] validate with plot and comparisons with other clustering algorithms, as in [this page from scikit-learn](https://scikit-learn.org/stable/auto_examples/cluster/plot_cluster_comparison.html#sphx-glr-auto-examples-cluster-plot-cluster-comparison-py))
- [ ] compare Coreset + sklearn.KMeans with dask_ml.KMeans

### DIFF
--- a/dask_ml/cluster/coreset.py
+++ b/dask_ml/cluster/coreset.py
@@ -1,0 +1,75 @@
+from sklearn.base import BaseEstimator, TransformerMixin
+
+import logging
+import dask.array as da
+import numpy as np
+from ..utils import _timer
+from .._utils import copy_learned_attributes
+import inspect
+
+logger = logging.getLogger(__name__)
+
+def lightweight_coresets(X, m):
+    """
+    Parameters
+    ----------
+    X : dask.array, shape = [n_samples, n_features]
+        input dask arrat to be sampled
+    m : int
+        number of samples to pick from `X`
+    """
+    dists = ((X - X.mean(axis=0)) ** 2).sum(axis=1)
+    q = 0.5 / X.shape[0] + 0.5 * (dists / dists.sum())
+    idxs = da.random.choice(X.shape[0], size=m, p=q)
+    weights = 1.0 / (m * q[idxs])
+    return X[idxs, :], weights
+
+
+class Coreset(BaseEstimator, TransformerMixin):
+    def __init__(self, estimator, m=None, k=None, eps=0.05):
+        self.m = m
+        self.eps = eps
+        self.estimator = estimator
+
+        if hasattr(estimator, "n_clusters"):  # eg. KMeans()
+            self.k = estimator.n_clusters
+            logger.info(f"k set to {self.k}")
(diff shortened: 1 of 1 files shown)


remiadon:
Currently I made this comparison, taking [this sklearn doc](https://scikit-learn.org/stable/auto_examples/cluster/plot_cluster_comparison.html#sphx-glr-auto-examples-cluster-plot-cluster-comparison-py) as a baseline


Unknown-3

The original data has 150k rows and 2 features, 200 points are sampled with the Coreset class

*Notes*
I experienced very poor running time performances with small chunks=(100, 2) when building dask.array 
The current graph show running time with a chunk size of (10_000, 2)

remiadon:
Here is a global view of the learning process

dask-coreset

TomAugspurger:
Thanks for continuing to work on this.

> I experienced very poor running time performances with small chunks=(100, 2) when building dask.array

Depending on the scheduler there's a roughly 10-200 microsecond per task overhead (https://docs.dask.org/en/latest/institutional-faq.html?highlight=overhead#how-well-does-dask-scale-what-are-dask-s-limitations).

Can you think of anyone from the scikit-learn side who would be able to review this? I'm not too familiar with the coreset idea / Gaussian mixing. When you're ready, I can take a look at things from the Dask side.

remiadon:
@TomAugspurger my pleasure. Concerning the review from the sklearn side, I think that Jérémie de Boisberranger worked on [a set of runtime/memory benchmarks for KMeans](https://github.com/jeremiedbb/K-means), whose results have been presented at the 2019 scikit-learn consortium.

Concerning interactions with the sklearn.GaussianMixture, I guess we can ask @gmaze, who initially raised the issue on adding Gaussian Mixture Models. He certainly is more informed than I am on how we should validate that our Coreset class can be used (with care ?) with GaussianMixture. My only intuition is that the sklearn.GaussianMixture uses KMeans internally to init the cluster centers (see the `init_params` parameter). The plot at the top of this page visually confirms that the `Coreset(GaussianMixture)` instance succeeds in clustering the flattened forms in the middle, where KMeans fail. 

Also, I would need help on how to properly compare runtime with sklearn. My current benchmark takes the sklearn.KMeans class as an opponent, but sklearn has a [MiniBatchKMeans](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.MiniBatchKMeans.html) which is meant to be faster.
This is to be sure that
 - `Coreset(Kmeans()).fit()` runs as fast as `MiniBatchKmeans` for datasets that hold in memory
 - `Coreset(Kmeans()).fit()` can scale to out-of-memory datasets and still deliver good quality clustering

PR #794, opened 2021-02-09
MNT Fixes failing test_keras

thomasjpfan:
Fixes failing test on the main branch

PR #800, opened 2021-02-24
TST Fixes random seed for flaky test_keras

thomasjpfan:
Follow up to #794

I suspect the failing test on `main` is due to sampling of the parameter space that may lead to scores that are constant.

CC @stsievert

### DIFF
--- a/tests/model_selection/test_keras.py
+++ b/tests/model_selection/test_keras.py
@@ -46,7 +46,9 @@ def _keras_build_fn(lr=0.01):
 @gen_cluster(client=True, Worker=Nanny, timeout=20)
 def test_keras(c, s, a, b):
     # Mirror the mnist dataset
-    X, y = make_classification(n_classes=10, n_features=784, n_informative=100)
+    X, y = make_classification(
+        n_classes=10, n_features=784, n_informative=100, random_state=0
+    )
     X = X.astype("float32")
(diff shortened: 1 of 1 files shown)


tests/model_selection/test_keras.py
```diff
@@ -56,7 +58,12 @@ def test_keras(c, s, a, b):
     params = {"lr": loguniform(1e-3, 1e-1)}
```
stsievert, line 58:
```suggestion
    params = {"lr": loguniform(1e-3, 1e-1), "random_state": [1]}
```

If we're fixing the random state, why not in the model too? This fixes the initialization (not the index ordering of the optimizer).

stsievert (commented):
Yeah, `random_state` is producing the errors. I'm not seeing why else the test would be non-deterministic.

stsievert:
Hm.... looks like the error has to do with timeout on windows. I'm not seeing how that's induced by this PR.

thomasjpfan:
The same test is also failing on `main`: https://dev.azure.com/dask-dev/dask/_build/results?buildId=1961&view=logs&jobId=2eab4704-62a3-55d0-a524-46b534a6e811&j=2eab4704-62a3-55d0-a524-46b534a6e811&t=24a2918a-3d47-54f3-09aa-2db1b58035d7

stsievert (approved):

PR #805, opened 2021-03-18
fix typos

jameslamb:
I've been trying to get familiar with `dask-ml` recently and find little ways to contribute. As part of that, I spent a little time tonight looking for typos. Typos in the repo can lead to false negatives in search, which might make it a little more difficult to navigate the code / docs.

I searched for these using `aspell`, like:

```shell
find . -type f -name "*.py" -exec cat {} \; \
    | aspell list \
    | sort -u 
```

That leads to a lot of things that aren't really typos, but a few looked like legitimate mistakes that should be fixed. This PR proposes fixing them.

Thanks for your time and consideration.

### DIFF
--- a/tests/preprocessing/test_data.py
+++ b/tests/preprocessing/test_data.py
@@ -121,7 +121,7 @@ class TestMinMaxScaler:
         mask = ["3", "4"]
         a = dpp.MinMaxScaler(columns=mask)
         result = a.inverse_transform(a.fit_transform(df2))
-        assert dask.is_dask_colelction(result)
+        assert dask.is_dask_collection(result)
         assert_eq_df(result, df2)
 
     def test_df_values(self):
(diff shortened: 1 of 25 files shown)


tests/preprocessing/test_data.py
```diff
@@ -121,7 +121,7 @@ def test_df_inverse_transform(self):
         mask = ["3", "4"]
         a = dpp.MinMaxScaler(columns=mask)
         result = a.inverse_transform(a.fit_transform(df2))
-        assert dask.is_dask_colelction(result)
+        assert dask.is_dask_collection(result)
```
jrbourbeau, line 124: Whoops

jrbourbeau (approved):
Thanks for taking the time to do this @jameslamb!

PR #806, opened 2021-03-18
Make all references to sklearn.utils.validation.check_is_fitted uniform

hristog:
**Associated GitHub Issue(s):** N/A
**Tests/`black`/`pyflakes` passing:** Yes
**Tests added:** No

I've noticed that `check_is_fitted` is referenced differently from a couple of files in the `dask-ml` codebase.

The majority (8 vs 2) of the references are `from .._compat import check_is_fitted`.

This PR proposes unifying the references (i.e., updating the remaining 2) for the sake of future-proofing, with respect to any potential updates in the behavior of `dask_ml/compat.py`.

### DIFF
--- a/dask_ml/_compat.py
+++ b/dask_ml/_compat.py
@@ -1,7 +1,7 @@
 import contextlib
 import os
 from collections.abc import Mapping  # noqa
-from typing import Any, List, Optional, Union
+from typing import Any
 
 import dask
 import dask.array as da
--- a/dask_ml/_compat.py
+++ b/dask_ml/_compat.py
@@ -43,12 +43,6 @@ annotate = dask.annotate if DASK_2021_02_0 else dummy_context
 blockwise = da.blockwise
 
 
-def check_is_fitted(est, attributes: Optional[Union[str, List[str]]] = None):
-    args: Any = ()
-
-    return sklearn.utils.validation.check_is_fitted(est, *args)
-
-
 def _check_multimetric_scoring(estimator, scoring=None):
     # TODO: See if scikit-learn 0.24 solves the need for using
     # a private method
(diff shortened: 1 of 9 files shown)


jrbourbeau (commented):
Thanks for the PR @hristog! 

After looking at `check_is_fitted` in `_compat.py`

https://github.com/dask/dask-ml/blob/027c978c253710a32ca1bd1fdd1b688aa32527a7/dask_ml/_compat.py#L46-L49

there doesn't appear to be any special logic for handling specific versions of packages (which is what the compatibility module is for). Digging a bit deeper it looks like there used to be some logic around the scikit-learn version being used, but it was removed in https://github.com/dask/dask-ml/pull/669. So I _think_ we can probably remove `check_is_fitted` from `_compat.py` and just import it directly from `sklearn` directly.

hristog:
> Digging a bit deeper it looks like there used to be some logic around the scikit-learn version being used, but it was removed in #669. So I _think_ we can probably remove `check_is_fitted` from `_compat.py` and just import it directly from `sklearn` directly.

Thanks for looking into this, @jrbourbeau!

Yes, I found it a bit odd that the `compat.py` version isn't doing anything specialized, but thought it has been planned to be extended in the future.

Fair enough, I'll update the references to point to the `sklearn` version instead.

jrbourbeau (commented):
Thanks for the updates @hristog! I merged the current `main` branch into this PR to resolve some unrelated test failures. 

It looks like there are some linting failures due to us removing `check_if_fitted` from `_compat`

```
./dask_ml/_compat.py:4:1: F401 'typing.List' imported but unused
./dask_ml/_compat.py:4:1: F401 'typing.Optional' imported but unused
./dask_ml/_compat.py:4:1: F401 'typing.Union' imported but unused
```

Other than those, this PR looks good to merge

hristog:
Thanks, @jrbourbeau! I'll get those sorted.

jrbourbeau (approved):
Thanks @hristog! This is in

PR #807, opened 2021-03-18
Temporarily restrict dask version in CI

jrbourbeau:
It looks like the `dask=2021.3.0` introduced a regression which is causing tests here to fail (see [a recent CI build on the `main` branch](https://dev.azure.com/dask-dev/dask/_build/results?buildId=1983&view=logs&j=ec4c5d99-1a39-52ae-ac68-aabdfb55b8a0&t=fc02acdf-e905-55f3-d38f-f279841d296b)). Locally I tested against the latest `main` branch of `dask` and the issue seems to be resolved.

This PR temporarily applies a version constraint to `dask` in CI to get tests to pass. We should revert the changes here after the next `dask` release.

### DIFF
--- a/ci/environment-3.7.yaml
+++ b/ci/environment-3.7.yaml
@@ -6,9 +6,12 @@ dependencies:
   - black==19.10b0
   - coverage
   - codecov
-  - dask
+  # dask 2021.3.0 introduced a regression which causes tests to fail.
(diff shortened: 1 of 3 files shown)


jrbourbeau:
A couple of tests are still failing on Windows for different reasons. I'm going to call this a step in the right direction.

Issue #734, opened 2020-09-03
Flag to skip centering in PCA


eric-czech:
Hey @TomAugspurger, how opposed would you be to introducing small deviations from the scikit-learn interface to PCA like adding a flag in dask-ml.PCA to avoid requiring this centering?

https://github.com/dask/dask-ml/blob/b94c587abae3f5667eff131b0616ad8f91966e7f/dask_ml/decomposition/pca.py#L270-L271

That would certainly be a bad default, but if I know that is going to happen downstream from a scaling operation in a pipeline that already does the centering, I would like to have the option to avoid the extra operation.  Is that reasonable or are you trying maintain something closer to perfect consistency?

TomAugspurger:
Seems reasonable to me!

PR #808, opened 2021-03-20
Introduce a `center` param to PCA (#734)

hristog:
- **Associated GitHub Issue(s):** https://github.com/dask/dask-ml/issues/734.
- **Tests/`black`/`pyflakes` passing:** Yes.
- **Tests added to the following files:**
  - `tests/test_pca.py`
  - `tests/test_incremental_pca.py`
- **tl;dr of design choices made:**
  - `IncrementalPCA` doesn't get affected by the new parameter and `IncrementalPCA(..., center=False ,...)` is unsupported.
  - `whiten=True` does not imply, nor enforce `center=True`. Instead it warns about possible unexpected behavior, if the underlying data hasn't been centered already.


hristog:
Once the maintainers confirm approval of the proposed changes, I'll be happy to update the `.rst` docs accordingly, to document the updated API, in a separate commit (as part of this PR).

dask_ml/decomposition/incremental_pca.py
```diff
@@ -127,14 +127,21 @@ def __init__(
...
         n_components=None,
         whiten=False,
+        center=True,
         copy=True,
         batch_size=None,
         svd_solver="auto",
         iterated_power=0,
         random_state=None,
     ):
+        if center is False:
```
TomAugspurger, line 137: I *think* that the scikit-learn API wants all validation of parameters to be deferred until `.fit` time. So this check could go at the top of `.fit`.

dask_ml/decomposition/pca.py
```diff
@@ -190,11 +216,19 @@ def __init__(
...
         self.svd_solver = svd_solver
         self.tol = tol
         self.iterated_power = iterated_power
         self.random_state = random_state
 
+        if whiten and not center:
+            warnings.warn(
+                "Whitening requires centering. Please, ensure that your data "
+                "is already centered, in order to avoid unexpected behavior.",
+                RuntimeWarning,
```
TomAugspurger, line 229: Users should have some way to silence this warning. I'm inclined to just document this requirement, and trust that the user has done it (no warning).

dask_ml/decomposition/pca.py
```diff
@@ -431,18 +472,25 @@ def inverse_transform(self, X):
         If whitening is enabled, inverse_transform does not compute the
         exact inverse operation of transform.
         """
-        check_is_fitted(self, "mean_")
+        check_is_fitted(self, "components_")
+
+        if self.center:
+            check_is_fitted(self, "mean_")
```
TomAugspurger, line 478: It seems that we always set `self.mean_`. Is this necessary in this if block?

dask_ml/decomposition/pca.py
hristog: Sure, I'll remove the warning (the new param has been already added to the doc-string). Please, let me know if any other documentation-related changes are necessary. I've checked `api.rst` and `compose.rst` but it doesn't seem that we need any updates to either.

dask_ml/decomposition/pca.py
hristog: Yes, indeed - `self.mean_` gets always set.
My implementation choice has been for the check to be inside the `if` block here, for the purpose of demonstrating intent (i.e., we utilize the `mean_` parameter here, only if we're performing the centering step).
The `mean_` parameter is still computed at `fit` time, regardless.

dask_ml/decomposition/incremental_pca.py
hristog: Yes, this makes sense. I've had a read over the scikit-learn codebase and it seems they've got a mixture of checks - some upon `__init__`, but _mostly_ upon `fit`.
For reference, some of the exceptions where checks are performed upon `__init__`:
- [decomposition/_kernel_pca.py](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/decomposition/_kernel_pca.py#L151) (commit from 2012-02-17, 9c6e0445ae)
- [decomposition/_fastica.py](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/decomposition/_fastica.py#L400) (commit from 2018-04-04, 1a3201cc30)
- [feature_extraction/text.py](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/feature_extraction/text.py#L1027) (commit from 2013-03-14, 3444bfbe88)
- [svm/_base.py](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/svm/_base.py#L81) (commit from 2020-02-16, 9b39c4c4d2)

Seems it's not necessarily been a recent guideline in their API, but I confirm I'm happy to update this PR accordingly.

P.S.: It seems the the scikit-learn codebase is not very consistent between `_check_params` and `_check_parameters` either. Sometimes either of these is utilized for validation, and in other instances - validation is performed directly as part of `_fit`/`fit` (I appreciate that that's saving one extra method call, but still...).

dask_ml/decomposition/incremental_pca.py
```diff
@@ -135,12 +136,18 @@ def __init__(
...
         self.n_components = n_components
         self.whiten = whiten
+        self.center = center
         self.copy = copy
         self.batch_size = batch_size
         self.svd_solver = svd_solver
         self.iterated_power = iterated_power
         self.random_state = random_state
 
+    def _check_params(self):
```
hristog, line 146: I've opted for `_check_params` (as opposed to `_check_parameters`), because that's what the majority choice has been:
```bash
$ grep -rnHwc --color=never "def _check_params" . --include=\*.py | grep -v ":0"                                                                                                                                                                                                                                
./sklearn/feature_selection/_univariate_selection.py:4
./sklearn/feature_extraction/text.py:1
./sklearn/cluster/_kmeans.py:2
./sklearn/neighbors/_graph.py:1
./sklearn/decomposition/_lda.py:1
./sklearn/decomposition/_nmf.py:1
./sklearn/ensemble/_gb.py:1

$ grep -rnHwc --color=never "def _check_parameters" . --include=\*.py | grep -v ":0"                                                                                                                                                                                                                            
./sklearn/mixture/_bayesian_mixture.py:1
./sklearn/mixture/_base.py:1
./sklearn/mixture/_gaussian_mixture.py:1
./sklearn/cluster/_bicluster.py:2
```

dask_ml/decomposition/pca.py
hristog: If you meant, whether it's _at all_ necessary to check if `mean_` has been set - yes, probably. Because this is in `transform()`, and - as it can be seen from the diff - the same check existed beforehand, it was just moved to indicate logically in what circumstance we need to utilize the `mean_` value.

hristog:
```bash
would reformat /home/vsts/work/1/s/tests/test_pca.py
Oh no! 💥 💔 💥
1 file would be reformatted, 99 files would be left unchanged.
```
Oops - pretty sure that I did this check locally. Will re-run `black` to confirm.

hristog:
Hi @TomAugspurger  - I'm writing to check if you've had a chance to look into the further changes that I pushed since [your review](https://github.com/dask/dask-ml/pull/808#pullrequestreview-619138162), and whether you'd like any additional refinements to be done on top of that.
Thanks!

TomAugspurger:
It'll be a week or so before I can take a look. Thanks for your patience!

In the meantime, perhaps @jameslamb has a chance to glance through this (no worries if you don't though)?

hristog:
> It'll be a week or so before I can take a look. Thanks for your patience!

It's all good - just wanted to check if this is still on your radar. I'm happy to wait as long as necessary, and incorporate potential further feedback.
Thanks for your confirmation, @TomAugspurger!

dask_ml/decomposition/incremental_pca.py
jameslamb: `scikit-learn`'s documentation is very c learn about this: validation should not happen at initialization time.

From [the "initialization" section in "Developing scikit-learn estimators"](https://scikit-learn.org/stable/developers/develop.html#instantiation):

> There should be no logic, not even input validation, and the parameters should not be changed.

And from [the "parameters and init" section](https://scikit-learn.org/stable/developers/develop.html#parameters-and-init)

> The easiest and recommended way to accomplish this is to not do any parameter validation in `__init__`. All logic behind estimator parameters, like translating string arguments into functions, should be done in `fit`.

dask_ml/decomposition/pca.py
```diff
@@ -60,6 +60,21 @@ class PCA(sklearn.decomposition.PCA):
...
+        without an intercept. As such, its pricipal components originate at the
+        origin of the transformed space.
+
+        `center` set to False may be employed when performing PCA on already
+        centered data.
+
+        Since centering is a required step as part of whitening, `center` set
+        to False and `whiten` set to True is a combination which may result in
+        unexpected behavior, if performed on not previously centered data.
+
```
jameslamb, line 77:
```suggestion
    center : bool, optional (default True)
        When True (the default), the underlying data gets centered at zero
        by subtracting the mean of the data from the data itself.

        PCA is performed on centered data due to its being a regression model,
        without an intercept. As such, its principal components originate at the
        origin of the transformed space.

        ``center=False`` may be employed when performing PCA on already
        centered data.

        Since centering is a required step as part of whitening, ``center`` set
        to False and ``whiten`` set to True is a combination which may result in
        unexpected behavior, if performed on not previously centered data.

```

I believe there are a few mistakes in this documentation as it's currently written.

* two backticks should be used for inline code formatting, since these docstrings are interpreted as [reStructuredText](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup) by Sphinx
* the docstring should say that the default behavior (`center=True`) is to automatically center the data (it currently says the opposite)

dask_ml/decomposition/pca.py
```diff
@@ -149,21 +164,28 @@ class PCA(sklearn.decomposition.PCA):
     >>> dX = da.from_array(X, chunks=X.shape)
     >>> pca = PCA(n_components=2)
     >>> pca.fit(dX)
-    PCA(copy=True, iterated_power='auto', n_components=2, random_state=None,
-      svd_solver='auto', tol=0.0, whiten=False)
+    PCA(n_components=2)
```
jameslamb, line 167: why were these changes necessary? They seem unrelated to the current pull request. Could you please revert them or explain why they're necessary?

dask_ml/decomposition/pca.py
```diff
@@ -175,13 +197,16 @@ class PCA(sklearn.decomposition.PCA):
       ``dask.linalg.svd_compressed``.
     * n_components : ``n_components='mle'`` is not allowed.
       Fractional ``n_components`` between 0 and 1 is not allowed.
+    * center : defaults to ``True`` and enables control over whether centering
+      gets implicitly performed as part of the PCA model steps.
```
jameslamb, line 201:
```suggestion
    * center : if ``True`` (the default), automatically center input data before performing PCA.
          Set this parameter to ``False`` if the input data have already been centered before running ``fit()``.
```

I don't think this language is quite correct. This parameter isn't about whether centering is done "implicitly" or "explicitly", but about whether it's done automatically or this class assumes that it's being given already-centered today. Please consider my suggested change to the language.

dask_ml/decomposition/pca.py
```diff
@@ -175,13 +197,16 @@ class PCA(sklearn.decomposition.PCA):
...
     """
 
     def __init__(
         self,
         n_components=None,
         copy=True,
         whiten=False,
+        center=True,
```
jameslamb, line 209: similar comment to `IncrementalPCA`...putting a new argument in the middle of the signature is a breaking change for anyone code constructing instance of this class with positional arguments. Please move this to the end of the signature.

dask_ml/decomposition/incremental_pca.py
```diff
@@ -135,12 +136,18 @@ def __init__(
...
+        super()._check_params()
```
jameslamb, line 147: What is the purpose of defining an empty `_check_params()` on the parent class and then calling it here? In my opinion, that is adding a bit of additional indirection for no functional benefit. Could you please remove this line and the parent class's empty `_check_params()`?

jameslamb (commented):
I had some time to go through and review today. Overall, I'm in favor of the change and I can understand how it could be a nice optimization for experienced users.

Please consider some of the review comments I left.

I also think it would improve understanding for people who find this pull request from search or see it in a changelog if you change the PR title to something like "Make automatic centering in PCA methods optional". By itself, the fact that there is now a parameter called `center` doesn't tell you much about the functional benefit of the changes in this pull request.

hristog:
> I had some time to go through and review today. Overall, I'm in favor of the change and I can understand how it could be a nice optimization for experienced users.
> 
> Please consider some of the review comments I left.
> 
> I also think it would improve understanding for people who find this pull request from search or see it in a changelog if you change the PR title to something like "Make automatic centering in PCA methods optional". By itself, the fact that there is now a parameter called `center` doesn't tell you much about the functional benefit of the changes in this pull request.

Thanks for your review, @jameslamb! I'll try to address them and make changes, wherever applicable, ASAP.

dask_ml/decomposition/pca.py
hristog: It really depends whether those are considered to be demonstrative examples (in doctest style) or actual doctests which are meant to pass.

If the case is the latter, then the proposed changes should probably still hold.

For example, in their original state, these doctests fail when run with:
```bash
py.test --verbose --doctest-modules dask_ml/
```

If, on the other hand, they're just meant to serve for demonstration only ([section 15](https://numpydoc.readthedocs.io/en/latest/format.html) of the `numpydoc` standard), then I'll be definitely happy to revert back these changes.

dask_ml/decomposition/pca.py
hristog: > the docstring should say that the default behavior (center=True) is to automatically center the data (it currently says the opposite)

Well spotted, thanks!
Updating as per the double backtick suggestion. I'll update the rest of the affected docstrings, to conform to the same guideline (unless we'd like to do that in a separate, dedicated PR?).

dask_ml/decomposition/incremental_pca.py
hristog: The purpose is to future-proof this by ensuring any checks that get eventually introduced in the superclass, get considered at this level as well.

dask_ml/decomposition/pca.py
jameslamb: This PR is adding one new parameter with a default value set to the current behavior. I wouldn't expect any tests to fail, and if they are I'd like to know more about what failed (a link to logs, if you have it).

hristog: I don't think the doctests get run as part of the CI steps, and I imagine the doctests haven't been updated for quite some time (haven't checked with `git blame`) or the intention has never been that they get run as actual first-class-citizen test (but, instead, serve for demonstration).

E.g., when running against the latest `main`:
```python
$ py.test --verbose --doctest-modules dask_ml/
...

dask-ml/dask_ml/wrappers.py:92: DocTestFailure
___________________ [doctest] dask_ml.decomposition.pca.PCA ____________________
142 
143     Examples
144     --------
145     >>> import numpy as np
146     >>> import dask.array as da
147     >>> from dask_ml.decomposition import PCA
148     >>> X = np.array([[-1, -1], [-2, -1], [-3, -2], [1, 1], [2, 1], [3, 2]])
149     >>> dX = da.from_array(X, chunks=X.shape)
150     >>> pca = PCA(n_components=2)
151     >>> pca.fit(dX)
Expected:
    PCA(copy=True, iterated_power='auto', n_components=2, random_state=None,
      svd_solver='auto', tol=0.0, whiten=False)
Got:
    PCA(n_components=2)
...
```

dask_ml/decomposition/incremental_pca.py
hristog: Noted, thanks!

dask_ml/decomposition/pca.py
jameslamb: If I was a maintainer here, I'd prefer that a docstring-only change be made in a separate pull request, and that this comment only apply to the lines you are already adding / changing as of this pull request

dask_ml/decomposition/pca.py
jameslamb: huh, ok I see. I'm not too familiar with that type of test, apologies. But it does seem unrelated to the change in this pull request, so I think it would be better to revert those changes here and open a separate issue / PR about the general topic of "doctest tests are failing / not run"

TomAugspurger: I don't recall if we run doctests here or not. I wouldn't worry about that failing one. Scikit-Learn changed the repr of their objects to only show the parameters that are not the default, so the docstring changed.

dask_ml/decomposition/incremental_pca.py
```diff
@@ -127,6 +127,7 @@ def __init__(
         self,
         n_components=None,
         whiten=False,
+        center=True,
```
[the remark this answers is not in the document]
TomAugspurger: @hristog you marked this as resolved, but I don't think it is yet. Could you put `copy` back between `whiten ` and `batch_size` and make `center` the last keyword argument?

TomAugspurger (approved):
I think that https://github.com/dask/dask-ml/pull/808/files#r615299186 is the only outstanding issue. Good to go once that's fixed.

Issue #802, opened 2021-03-05
KMeans fit TypeError expected dtype object, got 'numpy.dtype[float64]'


Maia97:
**What happened**:
when fit X, get error:
```
TypeError                                 Traceback (most recent call last)
<ipython-input-11-7d3bc4c475f7> in <module>
      1 km = dask_ml.cluster.KMeans(n_clusters=3, init_max_iter=2, oversampling_factor=10)
----> 2 km.fit(X)
 ...
 ...
 TypeError: expected dtype object, got 'numpy.dtype[float64]'
```

**What you expected to happen**:
output:
` KMeans(init_max_iter=2, n_clusters=3, oversamplinh_factor=10)`

**Minimal Complete Verifiable Example**:

```python
import dask_ml.cluster

X, y = dask_ml.datasets.make_blobs(n_samples=1000000,
                                   chunks=100000,
                                   random_state=0,
                                   centers=3)
km = dask_ml.cluster.KMeans(n_clusters=3, init_max_iter=2, oversampling_factor=10)
km.fit(X)
```

**Environment**:

- Dask version: dask_ml 1.8.0, dask 2021.02.0
- Python version: 3.7
- Operating System: MacOS Catalina 10.15.5
- Install method (conda, pip, source): conda
- Other packages: numpy 1.20.1, pandas 1.2.2

TomAugspurger:
Can you search on the Dask / distributed issue trackers for similar issues? I vaguely recall similar issues but can't find them right now.

Mimran0715:
I'm also having a similar issue. I think #803 is closed but not sure how to get rid of TypeError.

hristog:
Hi @Maia97,

I _think_ the series of issues and PRs that @TomAugspurger has referred to is as follows:
- https://github.com/dask/dask/pull/6430 / https://github.com/dask/dask/issues/6330
- https://github.com/numpy/numpy/pull/14882
- https://github.com/pydata/sparse/issues/383
- https://github.com/numba/numba/issues/6041 / https://github.com/numba/numba/pull/6078

I've tried a few combinations of different versions of `dask`, `numpy` and `numba`, _but_ on Python 3.8, and I haven't been able to reproduce the issue.

Could you, please, try to update to the latest version of `numba` and then attempt to run your example again?
If this doesn't help, could you provide the version number for `numba` and also the entire `Traceback`, please?

I think this issue can be labeled as "Needs Info" in the meantime.

hristog:
[...]

PR #809, opened 2021-03-21
Require numba>=0.51.0 (#802)

hristog:
- **Associated GitHub Issue(s):** https://github.com/dask/dask-ml/issues/802.
- **Tests/`black`/`pyflakes` passing:** Yes.
- **Tests added:** No.
- **Description:** We saw over at #802 that due to a discrepancy in the underlying `numba` version, some unexpected behavior was observed. This PR proposes a minimum version requirement for `numba`, namely `0.51.0`, which is the earliest one, not enabling `dask-ml` to exhibit correct, expected behavior.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ with open(os.path.join(here, "README.rst"), encoding="utf-8") as f:
 install_requires = [
     "dask[array,dataframe]>=2.4.0",
     "distributed>=2.4.0",
-    "numba",
+    "numba>=0.51.0",
     "numpy>=1.17.3",
     "pandas>=0.24.2",
     "scikit-learn>=0.23",


hristog:
Please, note that `numba 0.51.0` was released on 12 Aug 2020, which means that this PR is _not_ in line with the Dask organization-wide proposal for an 18-month "sliding window" for support of associated dependency versions (https://github.com/dask/community/issues/136).

However, it seems that a strong argument can be made for requiring minimum version numbers, which address regressions, or breaking changes in the corresponding dependencies (the latter being the case, with respect to `numba`).

jrbourbeau (commented):
Thanks for tracking this down @hristog! This seems reasonable to me, though it'd be good to get a +1 from @TomAugspurger 

> Please, note that numba 0.51.0 was released on 12 Aug 2020, which means that this PR is not in line with the Dask organization-wide proposal for an 18-month "sliding window" for support of associated dependency versions 

This is okay -- in general `dask-ml` moves faster than `dask` and `distributed`

TomAugspurger:
Yep, thanks! CI failures look unrelated.

hristog:
Thanks for confirming, @jrbourbeau @TomAugspurger!

Is there a protocol for closing associated issues (e.g., subject to confirmation from Issue authors etc)?

jrbourbeau:
GitHub has a few keywords (e.g. "closes", "fixes") you can include in a pull request's description to automatically close an issue when the corresponding PR is merged (https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). Engagement from the original issue author is good to have, but we don't have a hard policy around it

hristog:
Understood - thanks again!

PR #813, opened 2021-03-28
Pin code check versions (black==20.8b1 and flake8==3.8.3)

hristog:
Pin code checking-related versions to their [dask](https://github.com/dask/dask/blob/main/.pre-commit-config.yaml) and [distributed](https://github.com/dask/distributed/blob/main/.pre-commit-config.yaml) equivalents.

### DIFF
--- a/ci/environment-3.6.yaml
+++ b/ci/environment-3.6.yaml
@@ -3,12 +3,12 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - black==19.10b0
+  - black==20.8b1
   - coverage
   - dask ==2.4.0
   - dask-glm >=0.2.0
   - distributed ==2.4.0
-  - flake8
+  - flake8==3.8.3
   - isort==4.3.21
   - msgpack-python ==0.6.2
   - multipledispatch
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,11 @@
 repos:
 -   repo: https://github.com/python/black
-    rev: 19.10b0
+    rev: 20.8b1
     hooks:
     - id: black
       language_version: python3
 -   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.9
+    rev: 3.8.3
     hooks:
     - id: flake8
       language_version: python3
--- a/ci/environment-3.7.yaml
+++ b/ci/environment-3.7.yaml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - black==19.10b0
+  - black==20.8b1
   - coverage
   - codecov
   # dask 2021.3.0 introduced a regression which causes tests to fail.
@@ -12,7 +12,7 @@ dependencies:
   # to allow CI to pass
   - dask !=2021.3.0
   - dask-glm >=0.2.0
-  - flake8
+  - flake8==3.8.3
   - isort==4.3.21
   - multipledispatch >=0.4.9
   - mypy
--- a/ci/environment-3.8.yaml
+++ b/ci/environment-3.8.yaml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
   - defaults
 dependencies:
-  - black==19.10b0
+  - black==20.8b1
   - coverage
   - codecov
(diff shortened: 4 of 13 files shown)


hristog:
@jsignell, I've just discovered your https://github.com/dask/dask/pull/7370 (I can see that the `isort` version there is `5.7.0`).
Could you, please, have a look at this PR when you get a chance?
I'll be happy to update this PR to `isort==5.7.0` (instead of the proposed `5.8.0`), if that's your preference.

jsignell:
I imagine I just picked the latest version of isort when I opened that PR on dask, so there is no special magic to that version :)

setup.cfg
```diff
@@ -19,6 +19,7 @@ include_trailing_comma=True
 force_grid_wrap=0
 combine_as_imports=True
 line_length=88
+profile=black
```
jrbourbeau, line 22: I think this addition here means we don't have to specify `--profile=black` in other placed where `isort` is run. Could you confirm whether or not this is the case and, if so, remove `--profile=black` accordingly?

hristog: Yes, thanks, shall do! I was on the fence about this, but now that you've brought it up as well, it's probably worth confirming at this stage whether or not it can be picked up properly, and then the redundancies will be eliminated, if that ends up being the case.

hristog:
> My main suggestion is that in dask/dask, there is a github action that runs pre-commit.

Apologies - I'm not very familiar with GitHub Actions, and wanted to double-check if implementing your suggestion could be done as part of this PR, or if it requires maintainer access to the repository?

setup.cfg
```diff
@@ -20,8 +20,6 @@ force_grid_wrap=0
 combine_as_imports=True
 line_length=88
 profile=black
-skip=
```
hristog, line 22: With the new `profile=black` option, skipping `docs/source/conf.py` is no longer necessary.

jsignell:
> GitHub Actions... could be done as part of this PR, or if it requires maintainer access to the repository?

It can be done as part of this PR or as a follow-on if you prefer. Shouldn't require any special access. This is how the dask pre-commit action is set up https://github.com/dask/dask/blob/main/.github/workflows/pre-commit.yml

hristog:
> It can be done as part of this PR or as a follow-on if you prefer. Shouldn't require any special access. This is how the dask pre-commit action is set up https://github.com/dask/dask/blob/main/.github/workflows/pre-commit.yml

Understood - thank you very much, @jsignell!
I'll update this PR a bit later today, to incorporate your suggestion.

ci/environment-3.6.yaml
```diff
@@ -3,13 +3,13 @@ channels:
...
   - dask ==2.4.0
   - dask-glm >=0.2.0
   - distributed ==2.4.0
-  - flake8
-  - isort==4.3.21
+  - flake8==3.8.3
+  - isort==5.8.0
   - msgpack-python ==0.6.2
   - multipledispatch
   - mypy
```
hristog, line 15: For reproducible/predictable CI runs, the `mypy` versions in these files should probably be pinned as well.

hristog:
@jameslamb has brought up a very good point in https://github.com/dask/dask-ml/pull/822#discussion_r610737352 about considering pinning the static checker versions in https://github.com/dask/dask-ml/blob/f5e5bb4d4d21782b0e93f0f6541e6f2501b0c06c/setup.py#L28 as well (if my understanding has been correct).
Also, I've just noticed that `mypy` is missing from [setup.py](https://github.com/dask/dask-ml/blob/f5e5bb4d4d21782b0e93f0f6541e6f2501b0c06c/setup.py).

@jrbourbeau, when you get a chance, could you confirm if you've got any objections, with respect to pinning optional-dependency versions and adding `mypy` to `setup.py`?

hristog:
> pinning the static checker versions

One potential disadvantage of that (i.e., pinning minimum versions for `isort`, `flake8` and friends) might be that the local dev environment would have to be modified which, in turn, might cause inconvenience when the same underlying tools would need to be utilized in the context of other projects.

On the other hand, that's already what happens with respect to the non-testing package requirements: https://github.com/dask/dask-ml/blob/f5e5bb4d4d21782b0e93f0f6541e6f2501b0c06c/setup.py#L13

PR #816, opened 2021-04-03
Update README and documentation index page

hristog:
README preview [here](https://github.com/hristog/dask-ml/blob/update-readme-and-doc-index/README.rst).

Apparently, the Codecov report is still pointing at the `master` branch (Update: I've just raised https://github.com/dask/dask-ml/issues/817 for that to be addressed under, if needs be). The [Codecov settings](https://app.codecov.io/gh/dask/dask-ml/settings) need to be updated accordingly (requires maintainer access, unless the target brach could be set up from elsewhere).

The badges are set to point to `main`.

Update: https://github.com/dask/dask-ml/issues/817 has been closed as done.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,7 +3,7 @@ Dask-ML
 
 |Build Status| |Azure Pipelines| |Coverage| |Doc Status| |Gitter| |Version Status| |NumFOCUS|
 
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks!

PR #820, opened 2021-04-06
make regression metrics 'multioutput' behavior consistent with scikit-learn (fixes #818)

jameslamb:
Fixes #818.
Fixes #819.

## Changes in this PR

* ensures that regression metrics respect `compute` argument when `multioutput == "raw_values"`
* ensures that regression metrics' treatment of `multioutput=None` matches the behavior of their scikit-learn equivalents
* removes irrelevant comment about `np.average()` in regression metrics
* adds tests on behavior of metrics for different values of `multioutput`

## Background

See the linked issues for an explanation. I believe this PR is worth accepting because it will resolve an inconsistency with `scikit-learn` and with `dask-ml` itself, but it is technically a breaking change.

Thanks for your time and consideration.

PR #822, opened 2021-04-06
add MAPE to regression metrics (fixes #691)

### LINKED ISSUE #691: feature request: mean absolute percentage error
Add mean absolute percentage error here
https://github.com/dask/dask-ml/blob/master/dask_ml/metrics/regression.py

using https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/metrics/_regression.py#L197

jameslamb:
This PR proposes adding `mean_absolute_percentage_error()` ("MAPE"), as originally suggested in #691.

It follows the implementation from `scikit-learn` (https://github.com/scikit-learn/scikit-learn/blob/9cfacf1540a991461b91617c779c69753a1ee4c0/sklearn/metrics/_regression.py#L280), including the use of `np.finfo(np.float64).eps` in the denominator to prevent divide-by-0 errors.

## Notes for reviewers

This PR adds a bit of test coverage by adding `mean_absolute_percentage_error()` to the `metric_pairs` fixture in tests. It would automatically get more specific coverage (like for combinations of `multioutput` and `compute`) if #820 is accepted.

Thanks for your time and consideration.

### DIFF
--- a/tests/metrics/test_regression.py
+++ b/tests/metrics/test_regression.py
@@ -7,12 +7,13 @@ import sklearn.metrics
 import dask_ml.metrics
 
 
-@pytest.fixture(params=["mean_squared_error", "mean_absolute_error", "r2_score"])
+@pytest.fixture(params=["mean_squared_error", "mean_absolute_error", "mean_absolute_percentage_error", "r2_score"])
 def metric_pairs(request):
     """Pairs of (dask-ml, sklearn) regression metrics.
 
     * mean_squared_error
     * mean_absolute_error
+    * mean_absolute_percentage_error
     * r2_score
     """
     return (
--- a/tests/metrics/test_regression.py
+++ b/tests/metrics/test_regression.py
@@ -18,9 +18,7 @@ if SK_024:
     _METRICS_TO_TEST.append("mean_absolute_percentage_error")
 
 
-@pytest.fixture(
-    params=_METRICS_TO_TEST
-)
+@pytest.fixture(params=_METRICS_TO_TEST)
 def metric_pairs(request):
     """Pairs of (dask-ml, sklearn) regression metrics.
 
--- a/tests/metrics/test_regression.py
+++ b/tests/metrics/test_regression.py
@@ -7,7 +7,14 @@ import sklearn.metrics
 import dask_ml.metrics
 
 
-@pytest.fixture(params=["mean_squared_error", "mean_absolute_error", "mean_absolute_percentage_error", "r2_score"])
+@pytest.fixture(
+    params=[
+        "mean_squared_error",
+        "mean_absolute_error",
+        "mean_absolute_percentage_error",
+        "r2_score",
+    ]
+)
 def metric_pairs(request):
     """Pairs of (dask-ml, sklearn) regression metrics.
 
--- a/tests/metrics/test_regression.py
+++ b/tests/metrics/test_regression.py
@@ -5,22 +5,28 @@ import pytest
 import sklearn.metrics
 
 import dask_ml.metrics
+from dask_ml._compat import SK_024
+
+_METRICS_TO_TEST = [
+    "mean_squared_error",
+    "mean_absolute_error",
+    "r2_score",
+]
+
+# mean_absolute_percentage_error() was added in scikit-learn 0.24.0
+if SK_024:
+    _METRICS_TO_TEST.append("mean_absolute_percentage_error")
 
 
 @pytest.fixture(
-    params=[
-        "mean_squared_error",
-        "mean_absolute_error",
-        "mean_absolute_percentage_error",
-        "r2_score",
-    ]
+    params=_METRICS_TO_TEST
 )
 def metric_pairs(request):
     """Pairs of (dask-ml, sklearn) regression metrics.
 
     * mean_squared_error
     * mean_absolute_error
-    * mean_absolute_percentage_error
+    * mean_absolute_percentage_error (if scikit-learn >= 0.24.0)
     * r2_score
     """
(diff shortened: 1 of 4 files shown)


tests/metrics/test_regression.py
```diff
@@ -7,12 +7,13 @@
 import dask_ml.metrics
 
 
-@pytest.fixture(params=["mean_squared_error", "mean_absolute_error", "r2_score"])
+@pytest.fixture(params=["mean_squared_error", "mean_absolute_error", "mean_absolute_percentage_error", "r2_score"])
```
hristog, line 10:
```suggestion
@pytest.fixture(
    params=[
        "mean_squared_error",
        "mean_absolute_error",
        "mean_absolute_percentage_error",
        "r2_score",
    ]
)
```

Looks like `black==19.10b0` [isn't happy](https://dev.azure.com/dask-dev/dask/_build/results?buildId=2037&view=logs&j=73591ef5-68f9-550f-ee9a-e10f200815d9&t=28d49b26-095c-5073-6ee0-65d189ec82f7&l=21) about the line length here.

hristog (changes requested):

tests/metrics/test_regression.py
hristog: Also, perhaps, it'd be nice if the correctness of the method was sanity-checked against its sklearn counterpart, just as it's done for some of the other metrics a bit further down in the same test file.

jameslamb: > Looks like black==19.10b0 isn't happy about the line length here.

Ah ok. If `dask-ml` has chosen to pin to older versions of linters, then I think the non-conda option documented at https://ml.dask.org/contributing.html#style will be unreliable, since https://github.com/dask/dask-ml/blob/f5e5bb4d4d21782b0e93f0f6541e6f2501b0c06c/setup.py#L29 doesn't have a pin for things like `black`.

Once i switched to the conda instructions there, I got the expected diff. Updated in https://github.com/dask/dask-ml/pull/822/commits/1142fcc93a95e2af72211410959afed14116121d.

> Also, perhaps, it'd be nice if the correctness of the method was sanity-checked against its sklearn counterpart, just as it's done for some of the other metrics a bit further down in the same test file.

Can you clarify what you want me to change? As far as I can tell, that is exactly what happens by adding `mean_squared_percentage_error` to the `metric_pairs` fixture. Every metric in that fixture is tested against its scikit-learn equivalent by https://github.com/dask/dask-ml/blob/f5e5bb4d4d21782b0e93f0f6541e6f2501b0c06c/tests/metrics/test_regression.py#L37.

hristog: > Ah ok. If dask-ml has chosen to pin to older versions of linters, then I think the non-conda option documented at https://ml.dask.org/contributing.html#style will be unreliable

You're absolutely right! I've got a PR over at https://github.com/dask/dask-ml/pull/813 waiting to be reviewed (for a couple of weeks now), and subsequently merged. It should improve the static-checking situation.

> Every metric in that fixture is tested against its scikit-learn equivalent by

Indeed - ignore me about this one, please! I got confused that we could introduce extra tests like the `test_mean_squared_log_error` one.

hristog: I'll bring up the question of whether the `setup.py` versions of the linters should be pinned, too, in https://github.com/dask/dask-ml/pull/813.

hristog (approved):

jameslamb:
I'm grateful for the `linux earliest` CI job, because it caught an issue with this PR. The original changes I proposed would have made `dask-ml` incompatible with `scikit-learn < 0.24`.

I just pushed https://github.com/dask/dask-ml/pull/822/commits/b05213b52941cb73db0e8ac6f375cd117cd5683b to attempt to address it.

Basically, the use of `@derived_from(sklearn.metrics)` for the version of `dask` used in the "earliest" environment is incompatible with `scikit-learn v0.23.x`, and results in this error:

```text
ImportError while loading conftest '/home/jlamb/repos/open-source/dask-ml/tests/conftest.py'.
tests/conftest.py:9: in <module>
    from dask_ml.datasets import (
dask_ml/__init__.py:4: in <module>
    from dask_ml.model_selection import _normalize
dask_ml/model_selection/__init__.py:6: in <module>
    from ._hyperband import HyperbandSearchCV
dask_ml/model_selection/_hyperband.py:12: in <module>
    from ._incremental import BaseIncrementalSearchCV
dask_ml/model_selection/_incremental.py:32: in <module>
    from ..wrappers import ParallelPostFit
dask_ml/wrappers.py:16: in <module>
    from .metrics import check_scoring, get_scorer
dask_ml/metrics/__init__.py:7: in <module>
    from .regression import (  # noqa
dask_ml/metrics/regression.py:91: in <module>
    ) -> ArrayLike:
../../../miniconda3/envs/dask-ml-earliest/lib/python3.6/site-packages/dask/utils.py:656: in wrapper
    module_name = original_klass.__module__.split(".")[0]
E   AttributeError: module 'sklearn.metrics' has no attribute '__module__'
```

`@derived_from()` is only used to inherit documentation, not any functionality, so I copied over the relevant `scikit-learn` docstring (https://github.com/scikit-learn/scikit-learn/blob/da3c2d2a19ade5ca69adb6952ecace811ed122ff/sklearn/metrics/_regression.py#L283). I'm assuming this is ok since `scikit-learn` uses BSD 3-clause and since this project bundles the `scikit-learn` license: https://github.com/dask/dask-ml/blob/main/licenses/scikit-learn/COPYING.

TomAugspurger:
Looks good, thanks!

PR #824, opened 2021-04-09
fix typo in CI config

jameslamb:
I noticed today, looking at the config for this project's CI jobs, that there is a typo in `ci/posix.yaml`. The variable `SKLEARN_DEV` is used to control whether or not the latest nightly build of `scikit-learn` should be installed. Several matrix items use the variable `SKLARN_DEV` (missing an "E").

This doesn't affect the jobs at all as far as I can tell, since not being set at all and being set to `"no"` should result in the same outcome from `eq(variables.SKLEARN_DEV, 'yes')`. But I still think it's worth fixing, just for the sake of correctness and greppability.

### DIFF
--- a/ci/posix.yaml
+++ b/ci/posix.yaml
@@ -10,13 +10,13 @@ jobs:
     matrix:
       linux37:
         envFile: 'ci/environment-3.7.yaml'
-        SKLARN_DEV: "no"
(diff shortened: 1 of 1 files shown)


jrbourbeau (approved):
Thanks @jameslamb!

PR #825, opened 2021-04-12
test multioutput for metrics.mean_squared_log_error()

jameslamb:
This PR proposes simplifying the current tests for regression metrics and expanding the test coverage of `mean_squared_log_error()`.

As of this PR, that metric function's behavior when `multioutput=None` and `multioutput="uniform_average"` would now be tested.

### DIFF
--- a/tests/metrics/test_regression.py
+++ b/tests/metrics/test_regression.py
@@ -11,6 +11,7 @@ from dask_ml._compat import SK_024
 
 _METRICS_TO_TEST = [
     "mean_squared_error",
+    "mean_squared_log_error",
     "mean_absolute_error",
     "r2_score",
 ]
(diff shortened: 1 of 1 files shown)


tests/metrics/test_regression.py
```diff
@@ -64,18 +65,6 @@ def test_mse_squared(squared):
     assert abs(result - expected) < 1e-5
 
 
-def test_mean_squared_log_error():
```
jameslamb, line 67: adding `mean_squared_log_error()` to the `metric_pairs` fixture means that this standalone test for it can be removed. Similarity between this and the `scikit-learn` equivalent is covered by `test_ok()` in this same file.

TomAugspurger:
Thanks James!

PR #831, opened 2021-05-03
Allow subset of scorers to fail when using multiple scorers

### LINKED ISSUE #830: With multiple scorers, if a scorer fails, GridSearchCV will fail
**What happened**:
I was running GridSearchCV with multiple scoring metrics. One of them ("neg_mean_poisson_deviance") was undefined for some folds b/c it is undefined when y_hat is 0. This was handled during scoring but when `create_cv_results` was called, this raised a `TypeError: 'float' object is not subscriptable`. This is b/c `score` would normally return a dictionary when mutliple scorers are requested but in this case it returned the value I had passed as `error_score` to `GridSearchCV`, which in this case was `np.nan`. The issue is between L274 and L297 in [methods.py](https://github.com/dask/dask-ml/blob/0ea276da1d78db582f40e1c256dfca4f70e6cbc6/dask_ml/model_selection/methods.py#L274).

I expected that score to be `np.nan` for the folds in which the scorer failed, but not to raise an error

It was somewhat complex to come up with a situation in which this occurred and might take a bit to configure a MVCE so I'm going to skip this for now and just propose a PR to fix this issue directly, but if helpful I can put one together.

I think this *should* be a fairly quick fix so I'm going to give it a try

bolliger32:
Fixes #830


bolliger32:
I believe the test failures are occurring in other functions unrelated to this PR

PR #832, opened 2021-05-04
Update Keras documentation

adriangb:
- Pass `loss` to the constructor as documented in https://www.adriangb.com/scikeras/stable/migration.html
- Update package versions
- Update syntax to use SciKeras compiled models

### DIFF
--- a/docs/source/keras.rst
+++ b/docs/source/keras.rst
@@ -14,8 +14,8 @@ these packages need to be installed:
 
 .. code-block:: bash
 
-   $ pip install tensorflow>=2.3.0
-   $ pip install scikeras>=0.1.8
+   $ pip install tensorflow>=2.4.0
+   $ pip install scikeras>=0.3.2
 
 These are the minimum versions that Dask-ML requires to use Tensorflow/Keras.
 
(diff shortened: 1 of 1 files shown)


docs/source/keras.rst
```diff
@@ -14,8 +14,8 @@ these packages need to be installed:
 
 .. code-block:: bash
 
-   $ pip install tensorflow>=2.3.0
-   $ pip install scikeras>=0.1.8
+   $ pip install tensorflow>=2.4.0
+   $ pip install scikeras>=0.3.2
```
stsievert, line 18: Why is this required? Dask-ML will still work with the removed versions, right?

docs/source/keras.rst
```diff
@@ -36,24 +36,18 @@ normal way to create a `Keras Sequential model`_
    from tensorflow.keras.layers import Dense
    from tensorflow.keras.models import Sequential
 
-   def build_model(lr=0.01, momentum=0.9):
```
stsievert, line 39: :+1:

stsievert (commented):
Looks like the tests are failing because TF/Keras isn't installed on this branch. Could you comment out this line to make sure tests pass?

https://github.com/dask/dask-ml/blob/0ea276da1d78db582f40e1c256dfca4f70e6cbc6/ci/posix.yaml#L53

The example on this doc page is tested in `tests/model_selection/test_keras.py` . Can you update that test too?

docs/source/keras.rst
adriangb: This should work generally, but some of the syntax in this tutorial may not. I think we should either update the versions, or remove them altogether (since not specifying a version usually gets you the latest version anyway).

stsievert: Dask-ML will not work with SciKeras v0.1.7. I think that version didn't have serialization (?).

We should make a note about the versioning. "The example below uses X. The usage with lower versions may be different than this example."

stsievert: Wasn't there also some issue about serialization of stateful optimizers like Adam?

adriangb:
The tests in e73daa9 were failing within TF.
I cut out all of the dask stuff in 1591be8 and just did `model.fit(X, y).score(X, y)` and am getting no errors.
Any ideas why tests might be failing in e73daa9 @stsievert ([failure line](https://dev.azure.com/dask-dev/dask/_build/results?buildId=2111&view=logs&j=34469798-9dec-5482-c96c-ad1eda9ec58d&t=b5370614-32af-516a-8930-943e3caa82ec&l=941))?

docs/source/keras.rst
stsievert: That's a good reason to require SciKeras v0.3.0.

PR #843, opened 2021-07-02
Scatter estimator in ParallelPostFit

### LINKED ISSUE #842: ParallelPostFit excessive scheduler memory use and CancelledError
Notebook with MCVE and all notes: https://nbviewer.jupyter.org/gist/rikturr/43336377678018d01d4f21f19dd7ef11

When using `ParallelPostFit` to train with pandas/numpy objects then predict on dask objects, I noticed that the scheduler memory use runs extremely high. Many times I would get a `CancelledError` and scheduler dying when calling `.predict()` with pretty small data sizes (refer to notebook for full code with outputs):

```python
X_train, X_test, y_train, y_test = ...
rf = ParallelPostFit(
    RandomForestClassifier(n_estimators=100, random_state=seed, n_jobs=-1)
)
_ = rf.fit(X_train, y_train)

preds = rf.predict(X_test)
_ = preds.compute()  # failure happens on this line after ~40 minutes
```

The scheduler memory balloons to npartitions of `X_test` * size of `rf`, which can get into the multiple GBs very fast. I noticed that each time an operation would get called on `preds`, this memory exchange would happen again. I realize that this is because `ParallePostFit` uses `map_partitions` behind the scenes, but does not broadcast the model objects. This causes Dask to send the object through the scheduler to each worker _each time you do something with `preds`_ (unless of course you persisted it).

Workaround is to broadcast the model object then use `map_partitions` yourself instead of the `ParallelPostFit` wrapper:

```python
rf = RandomForestClassifier(n_estimators=100, random_state=seed, n_jobs=-1)
_ = rf.fit(X_train, y_train)

rf_fut = client.scatter(rf, broadcast=True)

def dask_predict(df, model):
    return model.predict(df)

preds = X_test.map_partitions(
    dask_predict,
    model=rf_fut,
    meta=np.array([1])
)
```

I plan to follow up with a PR to fix this in the `ParallelPostFit` class

rikturr:
closes #842 

Functionality of ParallelPostFit is the same, except behind the scenes it now scatters the scikit-learn estimator if running with a distributed cluster. Compared to existing functionality showcased in #842 where scheduler memory increases by npartitions * estimator size, the memory increase is now just estimator size for a quick blip while it gets shipped out to the workers

Screen Shot 2021-07-02 at 1 21 26 PM

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -223,7 +223,8 @@ class ParallelPostFit(sklearn.base.BaseEstimator, sklearn.base.MetaEstimatorMixi
             dt = _transform(xx, self._postfit_estimator).dtype
             return X.map_blocks(_transform, estimator=estimator_or_fut, dtype=dt)
         elif isinstance(X, dd._Frame):
-            return X.map_partitions(_transform, estimator=estimator_or_fut)
+            meta = self._postfit_estimator.transform(X.head(1, npartitions=-1))
+            return X.map_partitions(_transform, estimator=estimator_or_fut, meta=meta)
         else:
             return _transform(X, estimator=self._postfit_estimator)
 
@@ -334,7 +335,10 @@ class ParallelPostFit(sklearn.base.BaseEstimator, sklearn.base.MetaEstimatorMixi
                 chunks=(X.chunks[0], len(self._postfit_estimator.classes_)),
             )
         elif isinstance(X, dd._Frame):
-            return X.map_partitions(_predict_proba, estimator=estimator_or_fut)
+            meta = self._postfit_estimator.predict_proba(X.head(1, npartitions=-1))
+            return X.map_partitions(
+                _predict_proba, estimator=estimator_or_fut, meta=meta
+            )
         else:
             return _predict_proba(X, estimator=self._postfit_estimator)
 
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -7,6 +7,7 @@ import dask.delayed
 import numpy as np
 import sklearn.base
 import sklearn.metrics
+from distributed import get_client
 from sklearn.utils.validation import check_is_fitted
 
 from dask_ml.utils import _timer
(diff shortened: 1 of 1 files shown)


mrocklin:
Thank you for tracking this down @rikturr .  A couple of questions/comments

1.  Does this still work if things are run in async mode?  (if they worked before, that is)
2. Do you have thoughts on how to test this?
3. You might want to try wrapping the estimator in dask.delayed instead.  That should handle the deduplication.  Actually, this is already an optimization in map_partitions and map_blocks.  We wrap keywords with large sizeof values in dask.delayed in order to avoid duplication.  It might be wise to track down why that didn't occur.  That might lead you to a more general fix.

rikturr:
@mrocklin 
1. Looking into the docs for async mode and will report back
2. Broadly, I was thinking that a good way to test this and potentially other issues is to have some integration tests set up that would allow for more/larger tests with distributed clusters. Is something like that available or in the works? For this specific situation, it can be reproduced with a `LocalCluster` so should be able to cover in unit tests. I'm not sure if there's a way to get scheduler memory use programatically from dask? I think that could be a good way to verify that dups aren't being sent
3. Ah, I didn't realize that map_partitions and map_blocks already check object sizes but that makes a lot of sense. So the better solution would be to make sure that the estimator does get delayed so we don't have to change anything in dask-ml (or xgboost, or lightgbm, etc). Looks like sys.getsizeof() on scikit-learn estimators doesn't return the accurate size. Checking the size with pickle does though:
image
Do you think its reasonable to have map_partitions and map_blocks use pickle+getsizeof to get the proper size? Might introduce some latency. Perhaps there's another way to get a more accurate size

mrocklin:
I recommend that you identify where this should have triggered within
map_partitions/map_blocks, construct a simple case showing that it doesn't,
and then raise an issue in dask/dask about the problem.  That tends to get
more eyeballs on it than dask/dask-ml

rikturr:
closing in favor of fix for root issue: https://github.com/dask/dask/issues/7860

PR #850, opened 2021-08-27
Data type Changes

shivamjai17:
Float is giving warning, To avoid it I changed with np.float.

### DIFF
--- a/dask_ml/_typing.py
+++ b/dask_ml/_typing.py
@@ -16,6 +16,6 @@ ArrayLike = TypeVar("ArrayLike", Array, np.ndarray)
 FrameOrSeriesUnion = Union[DataFrame, Series, dd.Series, dd.DataFrame]
 SeriesType = Union[dd.Series, Series]
 DataFrameType = Union[DataFrame, dd.DataFrame]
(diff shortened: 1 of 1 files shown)


dask_ml/_typing.py
```diff
@@ -16,6 +16,6 @@
 FrameOrSeriesUnion = Union[DataFrame, Series, dd.Series, dd.DataFrame]
 SeriesType = Union[dd.Series, Series]
 DataFrameType = Union[DataFrame, dd.DataFrame]
-Number = Union[int, float, np.float64, np.int64, np.int32]
+Number = Union[int, np.float64, np.float64, np.int64, np.int32]
```
TomAugspurger, line 19:
```suggestion
Number = Union[int, np.float64, np.int64, np.int32]
```

np.float64 is already in this list.

TomAugspurger (commented):
Is this closing a specific issue? Can you add "closes #xxx" to your opening comment?

dask_ml/_typing.py
shivamjai17: closes #850

Issue #845, opened 2021-07-28
make_classification_df date functionality has single repeated value


ScottMGustafson:
**What happened**:
When running `dask_ml.datasets.make_classification_df` with a date range, the generated date column only has a single value, duplicated to length `chunks` (not `n_samples` as expected)

This seems to all be related to [this line here](https://github.com/dask/dask-ml/blob/3ef1c84fe02889d354247f72f476ae49a42bf321/dask_ml/datasets.py#L454) where it creates an array of length `len(X_df)` from a single value generated by [datasets.random_date](https://github.com/dask/dask-ml/blob/3ef1c84fe02889d354247f72f476ae49a42bf321/dask_ml/datasets.py#L384)

**What you expected to happen**:
we should be seeing
 - **many** randomly selected dates from within the date range `dates`
 - the date column should be of length `n_samples`, not `chunks`

**Minimal Complete Verifiable Example**:

```python
from dask_ml.datasets import make_classification_df
from datetime import date

X, y = make_classification_df(
    n_samples=100,
    n_features=5,
    random_state=123,
    chunks=10,
    dates=(date(2020, 1, 1), date(2021, 1, 1)),
)

X["date"].compute().value_counts()
```

returns

```
2020-02-27    10
Name: date, dtype: int64
```

**Anything else we need to know?**:

**Environment**:

- Dask version: 2021.7.1
- Python version: 3.7
- Operating System: linux (ubuntu)
- Install method (conda, pip, source): pip

PR #851, opened 2021-08-29
Populate make_classification_df date functionality with random dates #845

BhuvanashreeM:
This fixes the date functionality of make_classification_df as mentioned in #845 

Overview of the problem : 
- running `dask_ml.datasets.make_classification_df` with a _date range_ provided, fills the date column with **just one unique date**
- in dask.dataframe.from_array, specifying the `chunksize` to be equal to `chunks` (passed in through make_classification_df), populates the date column with **NaN values**.

Findings : 
- The helper function `random_date` works perfectly fine, generating a random date given the start and end
- [this line](https://github.com/dask/dask-ml/blob/3ef1c84fe02889d354247f72f476ae49a42bf321/dask_ml/datasets.py#L454) populates the list with the same date value, rather than calling the `random_date` function `len(X_df)` time, which is the required fix.

- [x] Run Existing Tests

- [x] Code Formatting (black, flake8, isort)

- [ ] Custom Tests: Will be added

Seeking the maintainers and @ScottMGustafson to review/provide feedback on the proposed changes.

### DIFF
--- a/dask_ml/datasets.py
+++ b/dask_ml/datasets.py
@@ -451,8 +451,8 @@ def make_classification_df(
             [
                 X_df,
                 dd.from_array(
-                    np.array([random_date(*dates)] * len(X_df)),
-                    chunksize=chunks,
+                    np.array([random_date(*dates) for i in range(len(X_df))]),
+                    chunksize=n_samples,
                     columns=["date"],
                 ),
             ],


dask_ml/datasets.py
```diff
@@ -451,8 +451,8 @@ def make_classification_df(
             [
                 X_df,
                 dd.from_array(
-                    np.array([random_date(*dates)] * len(X_df)),
-                    chunksize=chunks,
+                    np.array([random_date(*dates) for i in range(len(X_df))]),
```
stsievert, line 454: Looks like there's some test errors because `random_dates` doesn't have a random seed. Maybe this signature would help?

```
def random_dates(start, end, random_state=None):
    rng = check_random_state(random_state)
    ...
```

BhuvanashreeM: Yeah, that could be a probable reason, I'll work on it. And also, the test written for the function(`make_classification_df`) doesn't check for true randomness. I'm thinking of adding a new test/modifying the existing one for the same, will that suffice?
link to [test_make_classification_df](https://github.com/dask/dask-ml/blob/80eb0d81b22df3762706384c3ca1c0be4137e591/tests/test_datasets.py#L67)

BhuvanashreeM:
Here's what I've done in the [latest commit](https://github.com/dask/dask-ml/pull/851/commits/424fca4924e3aaab0fa070c10eb9800f7b885ae4) :
- included a seed(`random_state`) as an additional argument to the function `random_date`
- added a simple test in (`test_datasets.py`) for checking randomness given the seed, I could add more complex tests for checking true randomness in the date column. Will that be requisite? @stsievert 

Another observation I made earlier today was that: the main repository when cloned and existing tests run, I observed the same number of errors that I previously got when running the tests after making my modification in the [earlier commit](https://github.com/dask/dask-ml/pull/851/commits/4b55f9a8defb3a083e9fd3bf4a0a297734590ad7). Could the maintainers of the repository look into the same?

dask_ml/datasets.py
stsievert: Isn't #845 filed around having only one unique value in the date column? I'd make sure there's at least a couple values.

dask_ml/datasets.py
```diff
@@ -451,8 +452,13 @@ def make_classification_df(
             [
                 X_df,
                 dd.from_array(
-                    np.array([random_date(*dates)] * len(X_df)),
-                    chunksize=chunks,
+                    np.array(
+                        [
+                            random_date(*dates, random_state + i)
```
stsievert, line 457: What happens when `random_state` isn't an integer?  Scikit-learn allows for `random_state` to be an integer, `None` or an instance of `np.random.RandomState` ([source][1]).

[1]:https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html

tests/test_datasets.py
```diff
@@ -73,10 +73,21 @@ def test_make_classification_df():
...
     )
 
+    X_df1, y_series1 = dask_ml.datasets.make_classification_df(
+        n_samples=100,
+        n_features=5,
+        random_state=123,
+        chunks=100,
+        dates=(date(2014, 1, 1), date(2015, 1, 1)),
+    )
+    check_randomness = np.unique((X_df["date"] == X_df1["date"]).compute())
```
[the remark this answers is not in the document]
BhuvanashreeM: Yes, the code I've written checks for repeatability, on account of the seed. Since the the numpy's `randint` function is a deterministic pseudo random number generator, we can be sure that it will produce a random set of numbers. I've ensured in the code that the function `random_date` is called multiple times, each with a different seed. Hence I feel the line `np.unique(X["date"]).size >= threshold` would be redundant. Open to any thoughts you might have here @stsievert

dask_ml/datasets.py
BhuvanashreeM: Okay, I will have to raise a ValueError exception there, on it.

stsievert: Why not use this code?

``` python
rng = check_random_state(random_state)
dates = [random_date(*dates, rng) for i in range(len(X_df))]
...
```

tests/test_datasets.py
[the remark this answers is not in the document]
stsievert: The same `random_seed=123` is passed to both calls to `make_classification_df`; I would expect every value in `X_df["date"]` and `X_df1["date"]` to be the same.

I think `X["date"].nunique() >= threshold` would be a lot simpler.

dask_ml/datasets.py
BhuvanashreeM: The code above will produce the same random number since the seed(`rng`) remains the same in subsequent calls.

tests/test_datasets.py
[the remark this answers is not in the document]
BhuvanashreeM: Okay, that sounds good. Now I've to figure out, what would be a good threshold. Will `n_samples/2` be good?

dask_ml/datasets.py
stsievert: My main point: I think `np.random.RandomState` and `None` should be acceptable types for `random_state`. I'm fine expanding the for-loop, though I don't think that needs to happen:

<details>

``` python
[ins] In [193]: def random_dates(random_state):
           ...:     return random_state.randint(100)
           ...:

[ins] In [194]: rng = np.random.RandomState(42)

[ins] In [196]: [random_dates(rng) for _ in range(20)]
Out[196]: [51, 92, 14, 71, 60, 20, 82, 86, 74, 74, 87, 99, 23, 2, 21, 52, 1, 87, 29, 37]
```


</details>

BhuvanashreeM: I can maybe just check if random_state is one of the accepted values like this and accordingly proceed - 
`if random_state is not None or not isinstance(random_state, np.random.RandomState) or not isinstance(random_state,int):
    print("random_state is not to be accepted")`

stsievert: That runs counter to the use of `random_state` in Scikit-learn. `random_date` is public, so it should accept all types of `random_state` that Scikit-learn accepts.

If `random_date` were a private function, I wouldn't really care

tests/test_datasets.py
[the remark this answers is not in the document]
stsievert: Oh yeah, `threshold=n_samples/2` is more than good. I think `threshold=2` would suffice; that'd make sure #845 is resolved.

dask_ml/datasets.py
BhuvanashreeM: random_state : int, RandomState instance or None, optional (default=None), these are values accepted by Scikit-Learn's accepted values. I think I can check if the random_state is in neither of the accepted values(Scitkit and Numpy) and set is as the default None.

stsievert: Scikit-learn's `check_random_state` function will likely be useful: https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html

It takes those values and produces the correct random seed generator.

BhuvanashreeM: @stsievert the accepted types of `random_state` in Scikit and Numpy appear to be the same. 
Refer this: Scikit's version also accepts Numpy's accepted values. Refer this: https://scikit-learn.org/dev/glossary.html#term-random_state

dask_ml/datasets.py
```diff
@@ -381,10 +381,11 @@ def make_classification(
     return X, y
 
 
-def random_date(start, end):
+def random_date(start, end, random_state=None):
+    rng_random_date = dask_ml.utils.check_random_state(random_state)
```
stsievert, line 385: Nit:

```suggestion
    rng_random_date = sklearn.utils.check_random_state(random_state)
```

That way the `.compute()` can be avoided (especially relevant on repeated calls.).

dask_ml/datasets.py
```diff
@@ -430,6 +431,13 @@ def make_classification_df(
         The output values.
 
     """
+    if (
+        random_state is not None
+        or not isinstance(random_state, np.random.RandomState)
+        or not isinstance(random_state, int)
+    ):
+        random_state = None
```
stsievert, line 439: Why is the block necessary? `None` is already the default value for `random_state`.

BhuvanashreeM: This was to address the issue you mentioned earlier, "what if random_state is not an integer or any of the accepted values"

PR #862, opened 2021-10-15
[REVIEW]  Change `predict`, `transform`, `predict_proba` to infer metadata by default for `ParallelPostFit`  

VibhuJawa:
This PR fixes https://github.com/dask/dask-ml/issues/868


dask_ml/wrappers.py
```diff
@@ -279,10 +279,15 @@ def predict(self, X):
             return result
 
         elif isinstance(X, dd._Frame):
+            try:
+                meta = _predict(X._meta_nonempty, estimator=self._postfit_estimator)
+            except Exception:
```
quasiben, line 284: This would generally handle all cases though I'm not sure when it would come up.  @TomAugspurger do you have any better ideas on how to handle ?

VibhuJawa: Depending upon the model we can have various errors.   `IndexErrors` is one such example with  `CategoricalNB` model) (Added a test for it).

Sorry if you were all-ready aware but added a comment in detail [here](https://github.com/dask/dask-ml/pull/862#issuecomment-951341770).

TomAugspurger: I don't really see the need for this try / except

I'd prefer to handle `meta` here similarly to how we handle it everywhere in Dask: the default should be `None` (infer) and users should be able to override it. 

The wrinkle here is that `ParallelPostFit` previously assumed the output was int, so we need to deprecate that behavior.

So IMO we add a new keyword to `ParallelPostFit` called `meta` that gets passed here. By default it's `None`, which means warn and set it equal to `np.array([1])`. But in the future it'll just be passed through as is to `map_partitions`. Does that make sense?

VibhuJawa: > So IMO we add a new keyword to ParallelPostFit called meta that gets passed here. By default it's None, which means warn and set it equal to np.array([1]). But in the future it'll just be passed through as is to map_partitions. Does that make sense?

Gotcha. Thanks for the explanation.  Sorry for the back and forth. Just to be clear the plan is as follows : 

1. For the curent release, add a `meta` keyword which falls back to `np.array([1])` and raises warnings if not provided in the call. 

2. For future releases deprecate the `np.array([1])` fallback behavior by passing `meta` as is . 


Happy to update this PR to reflect `1` and raise an issue to track `2`  if we are on the same page regarding this.

jakirkham: @TomAugspurger did you have any thoughts on Vibhu's proposal above? 🙂

TomAugspurger: Yep that sounds right.

VibhuJawa: Thanks @TomAugspurger . Pushing on an update right away.

VibhuJawa: Pushed the requested changed  to the PR  commit: ( https://github.com/dask/dask-ml/pull/862/commits/f6316a978a066a6c2ebdea7dbee4da4a9b9393e8 )

Can i get a review on this. Raising an issue for 2 now.

VibhuJawa: Added issue as to track 2 [here](https://github.com/dask/dask-ml/pull/862). Please let me know if there is anything else that needs to be done to push this through.

dask_ml/wrappers.py
```diff
@@ -254,7 +255,7 @@ def score(self, X, y, compute=True):
         else:
             return self._postfit_estimator.score(X, y)
 
-    def predict(self, X):
+    def predict(self, X, meta=None):
```
TomAugspurger, line 258: This should be a parameter to the `ParallePostFit.__init__` I think not a `predict`.

And then we pass `self.meta` in this method.

docs/source/hyper-parameter-search.rst
```diff
@@ -465,6 +465,7 @@ to use post-estimation features like scoring or prediction, we recommend using
 :class:`dask_ml.wrappers.ParallelPostFit`.
 
 .. ipython:: python
+   :okwarning:
```
TomAugspurger, line 468: Can you update these to provide `meta=` whatever is appropraite?

tests/model_selection/test_hyperband.py
```diff
@@ -29,6 +29,7 @@
 pytestmark = pytest.mark.skipif(not DISTRIBUTED_2_5_0, reason="hangs")
 
 
+@pytest.mark.filterwarnings("ignore::FutureWarning")
```
TomAugspurger, line 32: These should be updated too.

tests/test_parallel_post_fit.py
```diff
@@ -49,6 +51,22 @@ def test_laziness():
     assert 0 < x.compute() < 1
 
 
+@pytest.mark.filterwarnings("ignore::FutureWarning")
+def test_predict_with_meta():
```
TomAugspurger, line 55: Can you also add a test asserting that a warning is raised when `meta` isn't provided?


```python
with pytest.warns(FutureWarning):
    ....
```

VibhuJawa: Thanks a lot for the review. Will add  that.

tests/model_selection/test_hyperband.py
VibhuJawa: Thanks a lot for the review. Will add that.

docs/source/hyper-parameter-search.rst
VibhuJawa: Thanks a lot for the review. Will add that.

dask_ml/wrappers.py
VibhuJawa: Just note that if we set `self.meta` then we cant use the same ParallePostFit estimator  on different dask-collections which may have different meta requirements say  `dask-dataframe` , `dask-array` , `dask-cudf dataframe` , `dask-cupy-array` etc.  
   

Just noting that we don't really use `self.meta` in stuff like  `transform` , ` predict_proba` function.  It felt weird having a self.meta that we only use for `predict function`  .   See below: 


### See  def transform(self, X)
https://github.com/dask/dask-ml/blob/f250432bba9046cdca3b464cb6e8eede879d3284/dask_ml/wrappers.py#L206-L213


###  def predict_proba(self, X)
https://github.com/dask/dask-ml/blob/f250432bba9046cdca3b464cb6e8eede879d3284/dask_ml/wrappers.py#L311-L322

I have the following questions: 

1.  Should we pass `self.meta` to all these functions  ?

TomAugspurger: > Just note that if we set self.meta then we cant use the same ParallePostFit estimator on different dask-collections which may have different meta requirements say dask-dataframe , dask-array , dask-cudf dataframe , dask-cupy-array etc.

I'll think about this, but I think that's OK. Users can use `.set_params(meta=my_meta)` to set this.

TomAugspurger: > Should we pass self.meta to all these functions ?

Hmm maybe... But unfortunately, I think `predict` and `predict_proba` might have different `meta` types, since some will return floats and some will return ints, right?

VibhuJawa: > Hmm maybe... But unfortunately, I think predict and predict_proba might have different meta types, since some will return floats and some will return ints, right?

Good Point, They will be different.  We can let `_predict_proba ` as is for now and just over ride `transform` and  `pedict ` and then tackle  `_predict_proba` later ?

TomAugspurger: Sure, sounds good.

docs/source/hyper-parameter-search.rst
VibhuJawa: Added with commit https://github.com/dask/dask-ml/pull/862/commits/094c31adc3352c50966f86d501739b1ea50a9c9e

tests/model_selection/test_hyperband.py
VibhuJawa: Removed with commit https://github.com/dask/dask-ml/pull/862/commits/094c31adc3352c50966f86d501739b1ea50a9c9e

tests/test_parallel_post_fit.py
VibhuJawa: Added test here: https://github.com/dask/dask-ml/blob/094c31adc3352c50966f86d501739b1ea50a9c9e/tests/test_parallel_post_fit.py#L60-L66

dask_ml/wrappers.py
VibhuJawa: Done. Addressed all reviews , should be good for another look now.

dask_ml/wrappers.py
```diff
@@ -202,13 +210,27 @@ def transform(self, X):
...
+            if meta is None:
+                xx = np.zeros((1, X.shape[1]), dtype=X.dtype)
+                dt = _transform(xx, self._postfit_estimator).dtype
+                return X.map_blocks(
+                    _transform, estimator=self._postfit_estimator, dtype=dt
+                )
+            else:
+                return X.map_blocks(
+                    _transform, estimator=self._postfit_estimator, meta=meta
+                )
```
TomAugspurger, line 225: Does this work? If I understand [the docs](https://docs.dask.org/en/stable/generated/dask.array.map_blocks.html?highlight=map_blocks) correctly, that should hopefully do the trick.

```suggestion
            if meta is None:
                xx = np.zeros((1, X.shape[1]), dtype=X.dtype)
                dtype = _transform(xx, self._postfit_estimator).dtype
                meta = type(dtype)((), dtype=dtype)
            return X.map_blocks(
                _transform, estimator=self._postfit_estimator, meta=meta
            )
```

dask_ml/wrappers.py
```diff
@@ -271,18 +293,27 @@ def predict(self, X):
         """
         self._check_method("predict")
         X = self._check_array(X)
+        meta = self.meta
+
+        if meta is None:
```
TomAugspurger, line 298: Can you put this warning in the `__init__`? I think that's where scikit-learn typically puts them.

TomAugspurger (commented):
Last question:

Right now we have `meta=None` warn, saying that the default will be to infer in the future. How does a user get that infer behavior today without a warning? I don't think they can (short of silencing the warning) since Dask uses `None` to mean infer.

So I think we need the actual default to be

```
_no_default = object()

class ParallelPostFit:
    def __init__(..., meta=_no_default):
        if meta is _no_default:
            warnings.warn(...)
        self.meta = meta
```

so that we can distinguish a user explicitly opting into inference from a user who just hasn't specified anything.


But do we even care about that case? I'm not sure, since I think the transform will essentially always fail on a length-0 input since sklearn does length checks... I'm not sure, but I think it's best to make this change.

VibhuJawa:
CI  failures  are probably unrelated to the PR. See CI issue: https://github.com/dask/dask-ml/issues/877

dask_ml/wrappers.py
VibhuJawa: Thanks again for the review. I will put that in `__init__`

dask_ml/wrappers.py
VibhuJawa: Gotcha, Let me test this and  change it to your suggestion

dask_ml/wrappers.py
VibhuJawa: Sorry , I put this in `predict` because this behavior change is only happening in `predict` .  Transform's behavior still remains the same so I think having it here makes much more sense as we are trying to warn users about the upcoming change which only impacts the `predict ` calls. 

Let me know what you think.

VibhuJawa:
> Right now we have `meta=None` warn, saying that the default will be to infer in the future. How does a user get that infer behavior today without a warning?

For dataframes they can do below: 
```python

```


 I don't think they can (short of silencing the warning) since Dask uses `None` to mean infer.

So this is not actually `True` for `dask.dataframe ` . See [call ](https://github.com/dask/dask/blob/cc0332f982b9baeb926402e4272c4bd85d62d4e1/dask/dataframe/core.py#L5808-L5811. It is true for `dask.array` though.   

```python
>>> import pandas as pd
>>> import dask.dataframe as dd
>>> df = dd.from_pandas(pd.DataFrame({'a':[1,2,3,4]}), npartitions=2)
>>> df.map_partitions(lambda df:df, meta=None)
```
```
  File "/datasets/vjawa/miniconda3/envs/dask-ml-dev-nov-8/lib/python3.8/site-packages/dask/dataframe/dispatch.py", line 127, in make_meta
    return make_meta_dispatch(x, index=index)
  File "/datasets/vjawa/miniconda3/envs/dask-ml-dev-nov-8/lib/python3.8/site-packages/dask/utils.py", line 620, in __call__
    meth = self.dispatch(type(arg))
  File "/datasets/vjawa/miniconda3/envs/dask-ml-dev-nov-8/lib/python3.8/site-packages/dask/utils.py", line 614, in dispatch
    raise TypeError(f"No dispatch for {cls}")
TypeError: No dispatch for <class 'NoneType'>

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/datasets/vjawa/miniconda3/envs/dask-ml-dev-nov-8/lib/python3.8/site-packages/dask/dataframe/core.py", line 730, in map_partitions
    return map_partitions(func, self, *args, **kwargs)
  File "/datasets/vjawa/miniconda3/envs/dask-ml-dev-nov-8/lib/python3.8/site-packages/dask/dataframe/core.py", line 5878, in map_partitions
    meta = make_meta(meta, index=meta_index, parent_meta=parent_meta)
  File "/datasets/vjawa/miniconda3/envs/dask-ml-dev-nov-8/lib/python3.8/site-packages/dask/dataframe/dispatch.py", line 131, in make_meta
    return func(x, index=index)
  File "/datasets/vjawa/miniconda3/envs/dask-ml-dev-nov-8/lib/python3.8/site-packages/dask/dataframe/backends.py", line 149, in make_meta_object
    return _nonempty_scalar(x)
  File "/datasets/vjawa/miniconda3/envs/dask-ml-dev-nov-8/lib/python3.8/site-packages/dask/dataframe/utils.py", line 320, in _nonempty_scalar
    raise TypeError(f"Can't handle meta of type '{typename(type(x))}'")
TypeError: Can't handle meta of type 'NoneType'
```

> so that we can distinguish a user explicitly opting into inference from a user who just hasn't specified anything.

I thought our ultimate aim was to default to inference and i dont really get the point of differentiating b/w a user who just hasn't specified anything.  I am not really sure what having a path where we ceate a  `_no_default = object()` do. 


Can you add an example so that i can understand this. I dont really see how this will be used .

dask_ml/wrappers.py
VibhuJawa: Thanks, Had to change the a bit but below does the trick. 
```python
  if meta is None:
       xx = np.zeros((1, X.shape[1]), dtype=X.dtype)
       x_t = _transform(xx, self._postfit_estimator)
       meta = type(x_t)((), dtype=x_t.dtype)
```
Resolving now.

VibhuJawa:
@TomAugspurger , I addressed some of the reviews you requested but did not make the changes around raising the warning in `__init__` (see [comment](https://github.com/dask/dask-ml/pull/862#discussion_r745140454))   . I also did not add the add the `_no_default = object()`  because of concerns raised in comment [here](https://github.com/dask/dask-ml/pull/862#issuecomment-963688095). 

It would be amazing if you could reply with you thoughts on both.

dask_ml/wrappers.py
TomAugspurger: Understood. The slight concern I have is that the fix for this is to actually set it in init. But agreed that having it here is better, since this won't affect you if you're just using fit.

TomAugspurger (commented):
Sorry, I think I confused myself with this `meta` stuff.

Maybe we just offer `meta` as an option to those that need it (cuml users)? That would mean we don't even need a warning. We just have a default of `None` which means "use ndarray, but override if  you need to".

Can you share an example of how cuml users will use this? Something like `ParallelPostFit(..., meta=cupy.array((), dtype=float))`

VibhuJawa:
> Can you share an example of how cuml users will use this? Something like `ParallelPostFit(..., meta=cupy.array((), dtype=float))`

This is how `cuML` users will have to use it.  `cuML` currently outputs a `cudf.Series` as default.  

```python
dask_model = ParallelPostFit(estimator=cuml_model, meta = cudf.Series([1]))
dask_model.predict(dask_df).compute()

dask_model = ParallelPostFit(estimator=cuml_model,  meta = cudf.Series([1]))
dask_model.predict(dask_cupy_ar).compute()
````


> Maybe we just offer `meta` as an option to those that need it (cuml users)? That would mean we don't even need a warning. We just have a default of `None` which means "use ndarray, but override if you need to".

I still think we should eventually `default` to `infer` rather than ndarray for `predict` like we do for `transform` .  This will streamline behavior across functions .  

More than happy to push on that once we give users enough time to update their calls.

**Comment by TomAugspurger** [comment]
1. I agree that we should be consistent between predict and transform.
2. Our long-term goal should be to infer, with the option to override.

The thing I'm not sure about is how to handle *dask-ml*'s code that helps with inference, as opposed to Dask's regular map_blocks / map_partition's inference. Because (last I checked) most sklearn estimators raise when there are zero samples, Dask's typical `map_blocks` inference fails. So we allocate an array with one sample at https://github.com/dask/dask-ml/pull/862/files#diff-20a89984412f6d21bb7a3f836c75a6f17df88ea034a8d886792dd5f102b58517R217 and infer meta on that.

I don't know how that translates back to user API though. Maybe it's not actually an issue. Any thoughts @quasiben or @VibhuJawa?

File: dask_ml/wrappers.py [hunk 8]
```diff
@@ -202,13 +210,24 @@ def transform(self, X):
         """
         self._check_method("transform")
         X = self._check_array(X)
+        meta = self.meta
```
jakirkham, line 213: Actually am curious whether we can just get `meta` from `X._meta`. We could also use [`meta_from_array`]( https://github.com/dask/dask/blob/ba78cda8167d78bed3f739b6e5d1504c2e99b46c/dask/array/utils.py#L23 )

cc @pentschev (in case you have thoughts here)

VibhuJawa:
> The thing I'm not sure about is how to handle _dask-ml_'s code that helps with inference, as opposed to Dask's regular map_blocks / map_partition's inference. Because (last I checked) most sklearn estimators raise when there are zero samples, Dask's typical `map_blocks` inference fails. 

1.   This is still the case.  Most estiimators will fail on zero shaped arrays. 

So we allocate an array with one sample at https://github.com/dask/dask-ml/pull/862/files#diff-20a89984412f6d21bb7a3f836c75a6f17df88ea034a8d886792dd5f102b58517R217 and infer meta on that.

> I don't know how that translates back to user API though. Maybe it's not actually an issue. Any thoughts @quasiben or @VibhuJawa?

**Eventually**, I think the following will be a good end behavior for users:

**For `dask-dataframes` :**

 &nbsp;   &nbsp;  **Default** 
   &nbsp;   &nbsp;   Default meta to `dd.core.no_default` because that gives us the ability to fall back to `_meta_nonempty`  thus we don't run into the zero shaped issue that  `dask-array` runs into . 
  
 &nbsp;   &nbsp;     **Reasoning:**
 &nbsp;   &nbsp;   That is `sk_model.predict(dask_df._meta_nonempty)` works but `sk_model.predict(dask_df._meta)` fails .  But as `dask-dataframe.map_partitions` falls back on [_meta_nonempty](https://github.com/dask/dask/blob/5b38dda3ff4c2432baa6af8b52ba1572a539548e/dask/dataframe/core.py#L5775-L5780) we are good. 

 &nbsp;   &nbsp;   **For Value Dependent Encoders**

 &nbsp;   &nbsp;   For encoders that are value dependent the users can over ride this with  `ParallelPostFit(..., meta=...)`  . 
 

**For `dask-arrays`:**


 &nbsp;   &nbsp;  **Default** 

 &nbsp;   &nbsp;    We can do what we do for `transform` (see [code](https://github.com/dask/dask-ml/pull/862/files#diff-20a89984412f6d21bb7a3f836c75a6f17df88ea034a8d886792dd5f102b58517R216-R219)) and run `meta` inference on `1 -sample` sized array, for `predict`  too.  
 
 &nbsp;   &nbsp;    **Reasoning:**
 
 &nbsp;   &nbsp;    ``sk_model.predict(dask_ar._meta)`` fails and we dont have  the `_meta_nonempty` fallback for `dask-arrays`. 

 
 &nbsp;   &nbsp;   **For Value Dependent Encoders**

 &nbsp;   &nbsp;   For encoders that are value dependent the users can over ride this with  `ParallelPostFit(..., meta=...)`  .

dask_ml/wrappers.py
VibhuJawa: This is a different meta than `X.meta` .  This is the meta we set here is for `predict`  and `transform` calls. 
See doc-string: 
https://github.com/dask/dask-ml/pull/862/files#diff-95e245a6cb7f05b184b66098b72f86099e09d2562205d9f22ed3f38298d89b46R157-R162

pentschev: For array-like, I would agree with @jakirkham suggestion for `meta_from_array`, but it seems like this is a more complex `meta` attribute that may combine either arrays or dataframes, then that wouldn't really apply here.

quasiben:
What do you think about creating (either in dask-ml or in dask-core) a small non-empty one container for arrays.  This would be something like:

> np.asarray(1, like=arr._meta)

Or in practice:

```python
In [10]: arr = cupy.arange(11)
In [11]: d_c = da.from_array(arr, chunks=(4,), asarray=False)
In [12]: type(np.asarray(1, like=d_c._meta))
Out[12]: cupy._core.core.ndarray
```

This would rely on NEP-35 first introduce in NumPy 1.20.1.  So we could push this to help with a default behavior of inferring type but pass the same bit of data to the `_postfit_estimator` meta arg.

> meta = np.asarray(1, like=d_c._meta) or meta = np.zeros((1,d_c.shape[1]), like=d_c._meta)

dask_ml/wrappers.py
TomAugspurger: The typical pattern to determine the new meta is to call the function on the old `.meta`. That's what `map_blocks` will do internally.

But we have an added complication that most scikit-learn functions will fail on length-0 inputs, so we'd need to generate some non-empty metadata.

VibhuJawa: Yup. This attribute should work for  either arrays or dataframes.  Resolving this comment. 

Please feel free to re-open if you think we can make this cleaner. 👍🏼

TomAugspurger:
Thanks Ben, that's helpful. I do think whatever small temporary arrays we create should use `like=`. 

OK, I've finally had a chance to sit down, install cudf, and play with this PR. I have some thoughts. Apologies for the repeated iteraations on this @VibhuJawa, but I think we're close.

I don't think that a single `meta=` argument to `ParallelPostFit.__init__` make sense. At a minimum, we would need something like a `predict_meta=..., predict_proba_meta=..., transform_meta=...`. So how about this:

1. Adopt a policy of ParallelPostFit.predict and ParallelPostFit.transform *will* infer the dtype by default. I'm OK with doing this as a bugfix. The old version was just plain wrong in assuming that it was always a specific dtype / array type.
2. Wherever possible, use the default behavior of `map_blocks` / `map_partitions`, applying the callable to an empty array / dataframe.
3. Wherever *necessary* construct a nonempty array of the right type and pass that to the subestimator. Maybe cuml already accepts NumPy arrays / pandas DataFrames in predict / transform, so using `like=...` isn't mandatory yet?
4. Add a `predict_meta`, `predict_proba_meta`, and `transform_meta` to `ParallelPostFit.__init__`. Both default to None (which means infer). I think we don't bother with warnings. But this gives users the option to override inference if necessary.

Again, apologies for going back and forth on this @VibhuJawa, but I think it's worth getting things right. I'll have some time tomorrow night to work on this and will push some commits if you don't beat me to it.

VibhuJawa:
>apologies for going back and forth on this @VibhuJawa, but I think it's worth getting things right. I'll have some time tomorrow night to work on this and will push some commits if you don't beat me to it.

No worries  on the back and forth. I agree with the idea of getting things right .   Thanks a lot for working through with me on this and the detailed suggestions. 

I agree with the approach that you suggested as well as all the suggestions.  I am gonna push some changes to this PR in accordance to  your suggestions and we can hopefully get those through. :-D

VibhuJawa:
@TomAugspurger ,  The PR should be ready for a review.  I have followed all the directions in this [comment](https://github.com/dask/dask-ml/pull/862#issuecomment-969403211).   Please let me know.

TomAugspurger:
Thanks @VibhuJawa, looks great.

jakirkham:
Thanks for helping review Tom! 😄

Issue #687, opened 2020-06-27
Should we create a `dask.ml` namespace?


mrocklin:
For dask.distributed we created a namespace in the `dask` so that people stopped calling `import distributed`.  This was also nice because it makes things feel a bit more integrated

```python
from dask.distributed import Client  # <--- actually imports from a separate distributed module
```

Should we do the same for other libraries?  Dask_ml seems like a good candidate to test this.  

This is the relevant file in the dask library: https://github.com/dask/dask/blob/master/dask/distributed.py

stsievert:
I don't have a strong opinion. I decided to look for some more context:

* Every library listed under "documentation" on https://dask.org has a `dask.foo` namespace *except* Dask-ML (array, dataframe, custom applications/delayed, real time/distributed).


<details>

<table>
<tr>
  <th>Documentation</th>
  <th>Imports</th>
  <th>URL</th>
</tr>
<tr>
  <td><img width="150" alt="Screen Shot 2020-06-27 at 3 17 44 PM" src="https://user-images.githubusercontent.com/1320475/85931361-746a4600-b8b3-11ea-916a-845c35dc0a69.png"></td>
<td>
<div class="highlight highlight-source-python">
<pre>
import dask.array as da
import dask.dataframe as dd
import dask_ml
import dask.delayed
from dask.distributed import Client
</pre>
</div>
</td>
<td>https://dask.org/</td>
</tr>
</table>

The "algorithms" heading under https://docs.dask.org adds Dask Bags and XArray, which are importable with `import dask.bag` and `import xarray as xr`. The "setup" heading mentions the following methods to setup Dask Distributed:

* Single machine: `import dask.distributed`
* SSHCluster: `from dask.distributed.deploy.ssh import SSHCluster`
* Job-Queue: `from dask_jobqueue import PBSCluster`
* Kubernetes: `from dask_kubernetes import KubeCluster`
* Yarn: `from dask_yarn import YarnCluster`

---

</details>


* The 2019 Dask user survey says this many respondents use the following APIs:
    *  Dask-ML, XArray: ~25%
    *  Futures: ~35%
    *  Array: ~55%
    *  Delayed: ~60%
    *  DataFrame: 75%
* There are 500 SO questions on "dask array," "dask dataframe" and "xarray", and 127 SO questions on "dask-ml" ([example query][q]).


[q]:https://stackoverflow.com/search?q=dask-ml

mrocklin:
[...]

PR #871, opened 2021-10-27
Update docstrings to import dask.ml

jacobtomlinson:
Now that dask/dask#6384 is in updated user examples to import from `dask.ml` instead of `dask_ml`.

xref #687

### DIFF
--- a/tests/model_selection/test_hyperband.py
+++ b/tests/model_selection/test_hyperband.py
@@ -38,93 +38,90 @@ pytestmark = pytest.mark.skipif(not DISTRIBUTED_2_5_0, reason="hangs")
         ("numpy", "ConstantFunction", 20),
     ],
 )
-def test_basic(array_type, library, max_iter):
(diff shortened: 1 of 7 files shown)


jrbourbeau (commented):
Thanks @jacobtomlinson. We'll need to wait for a `dask` release which includes https://github.com/dask/dask/pull/6384. Additionally, this will require us to bump the minimum supported version of `dask`

mmccarty:
@jacobtomlinson @jrbourbeau Are we to a point where this can be merged?

jrbourbeau:
Yes, since https://github.com/dask/dask/pull/6384 was included in the `dask=2021.11.0` release we should be good to go. However we'll need to update our minimum support `dask` version to `2021.11.0` here:

https://github.com/dask/dask-ml/blob/1e811ce48951a45a54859aca387510abdc085722/setup.py#L14

PR #874, opened 2021-11-01
Unpin `dask` in documentation environment

jrbourbeau:
This was pinned in https://github.com/dask/dask-ml/pull/807 but is no longer needed


quasiben:
Thanks @jrbourbeau

PR #879, opened 2021-11-10
Add pytest color to CI

jrbourbeau:
Similar to https://github.com/dask/dask/pull/8090 and https://github.com/dask/distributed/pull/5276. This will result in color coded output in CI (similar to what many folks see when running tests locally)


jrbourbeau:
Planning to merge later today if no further comment

Issue #870, opened 2021-10-26
[BUG] CI Broken with the latest distributed  release `2021.10.0`


VibhuJawa:
## [BUG] CI Broken with distributed `2021.10.0`

**Minimal Complete Verifiable Example**:

#### Create Environment
```bash
conda env create -f ci/environment-latest.yaml --name=dask-ml-dev
```

####  Run Tests: 

```python
pytest .
```

```python
=================================================================================================================== ERRORS ===================================================================================================================
__________________________________________________________________________________________ ERROR collecting tests/model_selection/test_hyperband.py __________________________________________________________________________________________
tests/model_selection/test_hyperband.py:163: in <module>
    def test_hyperband_patience(c, s, a, b):
/datasets/vjawa/miniconda3/envs/dask-ml-dev-oct-16/lib/python3.8/site-packages/distributed/utils_test.py:922: in _
    raise RuntimeError("gen_cluster only works for coroutine functions.")
E   RuntimeError: gen_cluster only works for coroutine functions.
_________________________________________________________________________________________ ERROR collecting tests/model_selection/test_incremental.py _________________________________________________________________________________________
tests/model_selection/test_incremental.py:338: in <module>
    def test_search_plateau_patience(c, s, a, b):
/datasets/vjawa/miniconda3/envs/dask-ml-dev-oct-16/lib/python3.8/site-packages/distributed/utils_test.py:922: in _
    raise RuntimeError("gen_cluster only works for coroutine functions.")
E   RuntimeError: gen_cluster only works for coroutine functions.
______________________________________________________________________________________ ERROR collecting tests/model_selection/test_incremental_warns.py ______________________________________________________________________________________
tests/model_selection/test_incremental_warns.py:11: in <module>
    def test_warns_decay_rate(c, s, a, b):
/datasets/vjawa/miniconda3/envs/dask-ml-dev-oct-16/lib/python3.8/site-packages/distributed/utils_test.py:922: in _
    raise RuntimeError("gen_cluster only works for coroutine functions.")
E   RuntimeError: gen_cluster only works for coroutine functions.
____________________________________________________________________________________________ ERROR collecting tests/model_selection/test_keras.py ____________________________________________________________________________________________
tests/model_selection/test_keras.py:47: in <module>
    def test_keras(c, s, a, b):
/datasets/vjawa/miniconda3/envs/dask-ml-dev-oct-16/lib/python3.8/site-packages/distributed/utils_test.py:922: in _
[...]

PR #882, opened 2021-11-15
Add upstream CI build

jrbourbeau:
There was a recent issue where a `dask` release broke `dask-ml` (xref https://github.com/dask/dask-ml/issues/870). This issue has since been fixed, but it would be nice to catch these types of breakages before an upstream dependency releases. To help increase visibility, this PR proposes we add a new `upstream` CI build, which is very similar to [the one used in Dask's CI](https://github.com/dask/dask/blob/43f33a32dd5cdf77363b6ff6348b5f1a8282d28c/.github/workflows/upstream.yml), to run the `dask-ml` test suite against the latest development version of a few key upstream dependencies.

In order to not fully block CI for `dask-ml` when an upstream CI failure occurs, this PR proposes we only run the upstream build on:

- Pushes to the `main` branch
- A nightly cron job (against the current `main` branch)
- On pull request commits where the commit message contains the phrase `test-upstream` (this allows contributors to opt-in to the upstream build for cases when it's useful)

In practice, we've found over in `dask/dask` that it's relatively easy for a CI job that only runs regularly on `main` to go unnoticed when a failure pops up. To help increase visibility in these situations, and still not have all CI on all PRs fail, this PR also proposes we add a `report` job to the `upstream` CI build which will automatically open an issue in the `dask-ml` issue tracker if the `upstream` CI build fails on `main`.

### DIFF
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -1,0 +1,124 @@
+name: Upstream
+
+on:
+  schedule:
+    - cron: "0 1 * * *"
+  push:
+  pull_request:
+
+jobs:
+
+  check:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event_name == 'pull_request'
+    outputs:
+      test-upstream: ${{ steps.detect-trigger.outputs.trigger-found }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - uses: xarray-contrib/ci-trigger@v1
+        id: detect-trigger
+        with:
+          keyword: "test-upstream"
+
+  build:
+    needs: check
+    runs-on: ubuntu-latest
+    if: |
+      always()
+      && (
+          needs.check.outputs.test-upstream == 'true'
+          || (github.repository == 'dask/dask-ml' && github.event_name != 'pull_request')
+      )
+
+    env:
+      COVERAGE: "true"
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+
+      - name: Setup Conda Environment
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          miniforge-variant: Mambaforge
+          miniforge-version: latest
+          use-mamba: true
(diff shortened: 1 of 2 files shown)


.github/workflows/upstream.yml
```diff
@@ -0,0 +1,124 @@
...
+                      number
+                    }
+                  }
+                }
+              }
+            }`;
+            const variables = {
+                owner: context.repo.owner,
+                name: context.repo.repo,
+                label: 'upstream',
```
jrbourbeau, line 103: Note this requires creating an `upstream` label (or some other agreed upon name) for this repository. I'm happy to do handle this, just didn't want to loose track of this TODO item

ci/install.sh
```diff
@@ -1,3 +1,24 @@
...
+        -i https://pypi.anaconda.org/scipy-wheels-nightly/simple \
+        numpy \
+        pandas \
+        scikit-learn
+
+    python -m pip install \
+        --upgrade \
+        git+https://github.com/pydata/sparse \
+        git+https://github.com/dask/dask \
+        git+https://github.com/dask/distributed
```
jrbourbeau, line 18: Looking at the output from this step in CI, the development version of `numpy` is getting installed above, but then overwritten to `numpy=1.20.3` here for some reason (see [here](https://github.com/dask/dask-ml/runs/4217350233?check_suite_focus=true#step:4:182)). All the other upstream dependencies are installed as expected

jrbourbeau: Opted to go with `dask`, `distributed`, and `scikit-learn` as the upstream packages to test against

jrbourbeau:
@TomAugspurger thoughts on adding an upstream CI build to `dask-ml`?

TomAugspurger:
Sounds good to me, especially the bit about making it a cron job! Merge whenever you're ready.

.github/workflows/upstream.yml
TomAugspurger: I just made the upstream label: https://github.com/dask/dask-ml/labels/upstream

PR #883, opened 2021-11-15
Remove unused Azure CI files

jrbourbeau:
To my knowledge these files are no longer needed after we switched to using GitHub actions for CI in https://github.com/dask/dask-ml/pull/863


TomAugspurger:
Yep, thanks.

Issue #888, opened 2021-11-27
Incremental.predict failing for scipy.sparse input


TomAugspurger:
From https://examples.dask.org/machine-learning/text-vectorization.html

```pytb
In [14]: predictions = pipe.predict(df['text'])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-14-2acfe3e16558> in <module>
----> 1 predictions = pipe.predict(df['text'])

/usr/local/Caskroom/miniconda/base/envs/dask-dev/lib/python3.8/site-packages/sklearn/utils/metaestimators.py in <lambda>(*args, **kwargs)
    111
    112             # lambda, but not partial, allows help() to work with update_wrapper
--> 113             out = lambda *args, **kwargs: self.fn(obj, *args, **kwargs)  # noqa
    114         else:
    115

/usr/local/Caskroom/miniconda/base/envs/dask-dev/lib/python3.8/site-packages/sklearn/pipeline.py in predict(self, X, **predict_params)
    468         for _, name, transform in self._iter(with_final=False):
    469             Xt = transform.transform(Xt)
--> 470         return self.steps[-1][1].predict(Xt, **predict_params)
    471
    472     @available_if(_final_estimator_has("fit_predict"))

~/gh/dask/dask-ml/dask_ml/wrappers.py in predict(self, X)
    315         if isinstance(X, da.Array):
    316             if meta is None:
--> 317                 meta = _get_output_dask_ar_meta_for_estimator(
    318                     _predict, self._postfit_estimator, X
    319                 )

~/gh/dask/dask-ml/dask_ml/wrappers.py in _get_output_dask_ar_meta_for_estimator(model_fn, estimator, input_dask_ar)
    663     # sklearn fails if input array has size size
    664     # It requires at least 1 sample to run successfully
--> 665     ar = np.zeros(
    666         shape=(1, input_dask_ar.shape[1]),
    667         dtype=input_dask_ar.dtype,

TypeError: The `like` argument must be an array-like that implements the `__array_function__` protocol.
```

Here are some values.

```
ipdb> pp input_dask_ar
dask.array<_transformer, shape=(nan, 1048576), dtype=float64, chunksize=(nan, 1048576), chunktype=scipy.csr_matrix>
ipdb> pp input_dask_ar._meta
<0x0 sparse matrix of type '<class 'numpy.float64'>'
	with 0 stored elements in Compressed Sparse Row format>
[...]

PR #889, opened 2021-11-28
[REVIEW] Add meta inference for `ParallelPostFit.predict` for sparse arrays

VibhuJawa:
This PR resolves https://github.com/dask/dask-ml/issues/888

Todo:
- [x] Add test for warning
- [x] Add test for ParallelPostFit

### DIFF
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -677,13 +677,17 @@ def _get_output_dask_ar_meta_for_estimator(model_fn, estimator, input_dask_ar):
         # Note below works for both cupy and scipy sparse matrices
         ar = type(input_meta)((1, input_dask_ar.shape[1]), dtype=input_dask_ar.dtype)
     else:
+        func_name = model_fn.__name__.strip("_")
         msg = (
-            "\nYou did not provide metadata, so Dask is running the"
-            "function on a small dataset to guess output types. "
-            "It is possible that Dask will guess incorrectly.\n"
-            "To provide an explicit output types or to silence this message, "
-            "please provide the `predict_meta`, `predict_proba_meta`,"
-            "`transform_meta` as appropiate"
+            f"Metadata for {func_name} is not provided, so Dask is "
+            f"running the {func_name} "
+            "function on a small dataset to guess output metadata. "
+            "As a result, It is possible that Dask will guess incorrectly.\n"
+            "To silence this warning, provide explicit "
+            f"`{func_name}_meta` to the dask_ml.wrapper."
+            "\nExample: \n"
+            "wrap_clf = dask_ml.wrappers.Incremental(GradientBoostingClassifier(), "
+            f"{func_name}_meta = np.array([1],dtype=np.int8))"
         )
         warnings.warn(msg)
         ar = np.zeros(shape=(1, input_dask_ar.shape[1]), dtype=input_dask_ar.dtype)
--- a/dask_ml/wrappers.py
+++ b/dask_ml/wrappers.py
@@ -663,19 +663,19 @@ def _get_output_dask_ar_meta_for_estimator(model_fn, estimator, input_dask_ar):
     """
     # sklearn fails if input array has size size
     # It requires at least 1 sample to run successfully
-    imput_meta = input_dask_ar._meta
-    if hasattr(imput_meta, "__array_function__"):
+    input_meta = input_dask_ar._meta
+    if hasattr(input_meta, "__array_function__"):
         ar = np.zeros(
             shape=(1, input_dask_ar.shape[1]),
             dtype=input_dask_ar.dtype,
-            like=imput_meta,
+            like=input_meta,
         )
-    elif "scipy.sparse" in type(imput_meta).__module__:
+    elif "scipy.sparse" in type(input_meta).__module__:
         # sparse matrices dont support
         # `like` due to non implimented __array_function__
         # Refer https:/q/github.com/scipy/scipy/issues/10362
         # Note below works for both cupy and scipy sparse matrices
-        ar = type(imput_meta)((1, input_dask_ar.shape[1]), dtype=input_dask_ar.dtype)
+        ar = type(input_meta)((1, input_dask_ar.shape[1]), dtype=input_dask_ar.dtype)
     else:
         msg = (
             "\nYou did not provide metadata, so Dask is running the"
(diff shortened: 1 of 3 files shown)


dask_ml/wrappers.py
```diff
@@ -662,9 +663,28 @@ def _get_output_dask_ar_meta_for_estimator(model_fn, estimator, input_dask_ar):
...
+    if hasattr(imput_meta, "__array_function__"):
+        ar = np.zeros(
+            shape=(1, input_dask_ar.shape[1]),
+            dtype=input_dask_ar.dtype,
+            like=imput_meta,
+        )
+    elif "scipy.sparse" in type(imput_meta).__module__:
+        # sparse matrices dont support
+        # `like` due to non implimented __array_function__
+        # Refer https:/q/github.com/scipy/scipy/issues/10362
```
GenevieveBuckley, line 676:
```suggestion
        # Refer https://github.com/scipy/scipy/issues/10362
```

Broken link is because of a typo here

dask_ml/wrappers.py
```diff
@@ -662,9 +663,28 @@ def _get_output_dask_ar_meta_for_estimator(model_fn, estimator, input_dask_ar):
     """
     # sklearn fails if input array has size size
     # It requires at least 1 sample to run successfully
-    ar = np.zeros(
-        shape=(1, input_dask_ar.shape[1]),
-        dtype=input_dask_ar.dtype,
-        like=input_dask_ar._meta,
-    )
+    imput_meta = input_dask_ar._meta
```
GenevieveBuckley, line 666:
```suggestion
    input_meta = input_dask_ar._meta
```

Is this a typo? The variable name probably makes more sense if it says "input" (instead of "imput")

dask_ml/wrappers.py
```diff
@@ -662,9 +663,28 @@ def _get_output_dask_ar_meta_for_estimator(model_fn, estimator, input_dask_ar):
...
+            dtype=input_dask_ar.dtype,
+            like=imput_meta,
+        )
+    elif "scipy.sparse" in type(imput_meta).__module__:
+        # sparse matrices dont support
+        # `like` due to non implimented __array_function__
+        # Refer https:/q/github.com/scipy/scipy/issues/10362
+        # Note below works for both cupy and scipy sparse matrices
+        ar = type(imput_meta)((1, input_dask_ar.shape[1]), dtype=input_dask_ar.dtype)
+    else:
```
GenevieveBuckley, line 679: When does this clause get triggered? Could we have a test for it?

dask_ml/wrappers.py
```diff
@@ -662,9 +663,28 @@ def _get_output_dask_ar_meta_for_estimator(model_fn, estimator, input_dask_ar):
...
+        msg = (
+            "\nYou did not provide metadata, so Dask is running the"
```
GenevieveBuckley, line 681:
```suggestion
            "\nYou did not provide metadata for the output, so Dask is running the"
```

I personally got slightly confused about which metadata was being talked about here: `input_dask_ar._meta` or metadata passed in with `predict_meta`/`predict_proba_meta`/`transform_meta`. I think this will only be confusing for developers who have read the code above this bit, since users will just see the message. In any case, it might help to make the message a bit clearer.

dask_ml/wrappers.py
```diff
@@ -662,9 +663,28 @@ def _get_output_dask_ar_meta_for_estimator(model_fn, estimator, input_dask_ar):
...
+            "function on a small dataset to guess output types. "
+            "It is possible that Dask will guess incorrectly.\n"
+            "To provide an explicit output types or to silence this message, "
```
GenevieveBuckley, line 684:
```suggestion
            "To provide an explicit output metadata to silence this message, "
```

I don't think "or" belongs in this sentence? Providing the explicit output metadata is the thing that silences this message, it's not two separate things, isn't that right? Also in this sentence it seems like `output types` => `output metadata`. 

If they are in fact two separate things, you'll need instructions on how to do both (eg: kwargs for providing explicit output type, as well as the kwargs for providing the output metadata)

GenevieveBuckley (commented):
I think the general approach here seems good. I've left a few comments, mostly about minor things.

dask_ml/wrappers.py
GenevieveBuckley: Ah, I see from your todo list you're planning on adding a test, that's great!

dask_ml/wrappers.py
VibhuJawa: Yup. That was a typo. Thanks for the review. Pushed a fix with commit . https://github.com/dask/dask-ml/pull/889/commits/a295e39d11826c60faf43e0d1d42dc66a9a0a265

dask_ml/wrappers.py
VibhuJawa: Addressed this with commit https://github.com/dask/dask-ml/pull/889/commits/48645448147126c6a343d6d085dd571be0cd1665

dask_ml/wrappers.py
VibhuJawa: Reworded it to make as clear as i could. 

This is how it will look the user now. 

```python
UserWarning: Metadata for predict_proba is not provided, so Dask is running the predict_proba function on a small dataset to guess output metadata. As a result, It is possible that Dask will guess incorrectly.
To silence this warning, provide explicit `predict_proba_meta` to the dask_ml.wrapper.
Example: 
wrap_clf = dask_ml.wrappers.Incremental(GradientBoostingClassifier(), predict_proba_meta = np.array([1],dtype=np.int8))
  warnings.warn(msg)
```

Thanks for the inputs, please  feel free to let me know if it is still confusing.

dask_ml/wrappers.py
VibhuJawa: Yup. Added a test [here](https://github.com/dask/dask-ml/blob/48645448147126c6a343d6d085dd571be0cd1665/tests/test_parallel_post_fit.py#L253-L291) . 

This basically gets triggered when using dask array wrapper around an object that does not support `__array_function__` .   A real world common example will be a Tensorflow tensor

VibhuJawa:
Thanks a lot for reviewing the PR @GenevieveBuckley and providing actionable inputs.  I  have addressed your reviews and completed my todo lists. This PR is now ready for another round of reviews.

GenevieveBuckley:
Thanks @VibhuJawa! Someone else from the maintenance team will have a look at this PR next.

TomAugspurger (approved):

TomAugspurger:
Thanks @VibhuJawa, and thanks for the review @GenevieveBuckley!

I'll make a release with this now.

PR #898, opened 2022-01-05
Specify HLG layer name during delayed object creation in fit

### LINKED ISSUE #895: Test failures with latest dask from main
**What happened**:
Multiple test failures post the merge of dask/dask@fa15d0c6d9604ca2267290942317058735a73d78 with `ValueError` around HLG graph layer names. 

here's the output from one of the tests, most errors I saw were of a similar nature. 

`test_incremental_sparse_inputs`
```python
E  ValueError: Layer ('fit-a42744f10f719e4c75eaf92f1c890e5f', 2) not in the HighLevelGraph's layers: 
['fit-a42744f10f719e4c75eaf92f1c890e5f', 'array-873f3177c716dd9595aeb50eb00dcb2b', 'array-fab8aba9efacb871881c861cdab60e29']
```

Initially popped up as part of failing tests in dask-contrib/dask-sql#350
**Environment**:

- Dask version: 2021.12.0+13.gfa15d0c6 or newer
- Python version: 3.8
- Operating System: Ubuntu 18.04

ayushdg:
Fixes #895.

Updated the `fit` method in the `_partial` module to specify the layer during delayed object creation as required after dask/dask#8452. 

Also added a commit unrelated to the issue replacing the usage of `distutils.version.LooseVersion` with `packaging.version.parse` as a side effect of needing it while testing locally. Happy to remove it or open a separate pr for those changes if preferred.

### DIFF
--- a/dask_ml/model_selection/methods.py
+++ b/dask_ml/model_selection/methods.py
@@ -2,11 +2,11 @@ from __future__ import absolute_import, division, print_function
 
 import warnings
 from collections import defaultdict
-from distutils.version import LooseVersion
 from threading import Lock
 from timeit import default_timer
 
 import numpy as np
+import packaging.version
 from dask.base import normalize_token
 from scipy import sparse
 from scipy.stats import rankdata
@@ -20,7 +20,7 @@ from .utils import _index_param_value, _num_samples, _safe_indexing, copy_estima
 
 # Copied from scikit-learn/sklearn/utils/fixes.py, can be removed once we drop
 # support for scikit-learn < 0.18.1 or numpy < 1.12.0.
-if LooseVersion(np.__version__) < "1.12.0":
+if packaging.version.parse(np.__version__) < packaging.version.parse("1.12.0"):
 
     class MaskedArray(np.ma.MaskedArray):
         # Before numpy 1.12, np.ma.MaskedArray object is not picklable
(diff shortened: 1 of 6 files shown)


dask_ml/model_selection/methods.py
```diff
@@ -20,7 +20,7 @@
 
 # Copied from scikit-learn/sklearn/utils/fixes.py, can be removed once we drop
 # support for scikit-learn < 0.18.1 or numpy < 1.12.0.
-if LooseVersion(np.__version__) < "1.12.0":
+if packaging.version.parse(np.__version__) < packaging.version.parse("1.12.0"):
```
ayushdg, line 23: These individual checks can probably be moved to the `_compat` module similar to how it's done in other parts of the codebase.

jrbourbeau (commented):
Thanks @ayushdg. @gjoseph92 would you mind taking a quick look at the changes here?

gjoseph92 (approved):
Thanks for fixing @ayushdg, LGTM. Sorry about the breaking change 😞

dask_ml/model_selection/methods.py
jrbourbeau: Alternatively, given that today the [minimum supported NumPy version for Dask is 1.18](https://github.com/dask/dask/blob/f34f1afd407d4371901a7ccb14a4a6b86c4cb4c2/setup.py#L13), we could probably just bump the minimum supported version for Dask-ML too (though let's handle that in a separate PR)

jrbourbeau (commented):
Thanks @ayushdg @gjoseph92 -- just pushed an empty commit to trigger a CI build against the `main` branch of `dask` (so we should now be exercising both the `if DASK_2022_01_0` and `else` code paths below)

jrbourbeau:
Whoops, forgot the `upstream` build is failing for unrelated reasons : / Just confirmed locally that `tests/test_incremental.py::test_incremental_sparse_inputs` passes with the current `dask` `main` branch 👍

PR #900, opened 2022-01-18
Implement check_consistent_length

### LINKED ISSUE #865: Implement `check_lengths`
https://github.com/dask/dask-ml/pull/863/files#diff-970a9a902963a060977ac1912b5a86a5177cebfd468517967f63335bb6d024eaR435-R437 added an empty `check_consistent_lengths`. We should implement a real one.

Unlike scikit-learn, we don't actually want to check the lengths when constructing the graphs. But we do want to check things like that the number of chunks match, that the chunk shapes match if present, divisions, etc.

freddyaboulton:
Fixes #865

### DIFF
--- a/dask_ml/utils.py
+++ b/dask_ml/utils.py
@@ -241,7 +241,7 @@ def check_random_state(random_state):
         raise TypeError("Unexpected type '{}'".format(type(random_state)))
 
 
-def check_matching_blocks(*arrays):
+def check_matching_blocks(*arrays, check_first_dim_only=False):
     """Check that the partitioning structure for many arrays matches.
 
     Parameters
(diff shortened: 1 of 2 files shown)


dask_ml/utils.py
```diff
@@ -241,8 +241,8 @@ def check_random_state(random_state):
         raise TypeError("Unexpected type '{}'".format(type(random_state)))
 
 
-def check_matching_blocks(*arrays):
-    """Check that the partitioning structure for many arrays matches.
+def _check_matching_blocks(*arrays, check_first_dim_only=False):
```
freddyaboulton, line 244: I thought this was the best way to re-use code between check_consistent_length and check_matching_blocks without changing the api of either function.

Issue #911, opened 2022-03-25
[BUG] Dask-ML ParallelPostFit prediction fails on empty partitions


VibhuJawa:
Dask-ML ParallelPostFit prediction fails on empty partitions

**Minimal Complete Verifiable Example**:

```python
from sklearn.linear_model import LogisticRegression
import dask.dataframe as dd
from dask_ml.wrappers import ParallelPostFit
import pandas as pd

df = pd.DataFrame({"x": [1, 2, 3, 4, 5, 6, 7, 8], "y": [True, False] * 4})
ddf = dd.from_pandas(df, npartitions=4)

clf = ParallelPostFit(LogisticRegression())
clf = clf.fit(df[["x"]], df["y"])

ddf_with_empty_part = ddf[ddf.x < 5][["x"]]
result = clf.predict(ddf_with_empty_part).compute()

expected = clf.estimator.predict(ddf_with_empty_part.compute())

assert_eq_ar(result, expected)
```
**TRACE**

```python

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [3], in <cell line: 1>()
----> 1 result.compute()

File /datasets/vjawa/miniconda3/envs/dask-ml-dev/lib/python3.8/site-packages/dask/base.py:290, in DaskMethodsMixin.compute(self, **kwargs)
    266 def compute(self, **kwargs):
    267     """Compute this dask collection
    268 
    269     This turns a lazy Dask collection into its in-memory equivalent.
   (...)
    288     dask.base.compute
    289     """
--> 290     (result,) = compute(self, traverse=False, **kwargs)
    291     return result

File /datasets/vjawa/miniconda3/envs/dask-ml-dev/lib/python3.8/site-packages/dask/base.py:573, in compute(traverse, optimize_graph, scheduler, get, *args, **kwargs)
    570     keys.append(x.__dask_keys__())
    571     postcomputes.append(x.__dask_postcompute__())
--> 573 results = schedule(dsk, keys, **kwargs)
    574 return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])

File /datasets/vjawa/miniconda3/envs/dask-ml-dev/lib/python3.8/site-packages/dask/threaded.py:81, in get(dsk, result, cache, num_workers, pool, **kwargs)
     78     elif isinstance(pool, multiprocessing.pool.Pool):
     79         pool = MultiprocessingPoolExecutor(pool)
---> 81 results = get_async(
     82     pool.submit,
[...]

PR #912, opened 2022-03-25
[WIP] Fix empty partition prediction with ParallelPostFit

VibhuJawa:


TomAugspurger:
Thanks @VibhuJawa. When looking at the traceback in #911, I see that skearn's `check_array` takes an `ensure_min_samples` parameter. If we pass `ensure_min_samples=False` there, does stuff go through properly?

VibhuJawa:
Thanks for reviewing the issue tom. 

> When looking at the traceback in https://github.com/dask/dask-ml/issues/911, I see that skearn's check_array takes an ensure_min_samples parameter. If we pass ensure_min_samples=False there, does stuff go through properly?

So i tried exploring a clean way to expose the `self._validate_data`  [parameter](https://github.com/scikit-learn/scikit-learn/blob/bf0ece817df520ec7e036d12a8021ff52ea9ff47/sklearn/linear_model/_base.py#L428-L428) but came up with nothing. Open to any ideas on that front.  

I think the problem is that each family of models in `sklearn`  calls it with different parameters, (See below) and I am also not sure if all models will just work even if we can some how coerce it to accept them.    (See related discussion [here](https://github.com/scikit-learn/scikit-learn/issues/22442#issuecomment-1035578069)) . 

1. [_base_chain](https://github.com/scikit-learn/scikit-learn/blob/26e1930abf5ad637e44573a152dadbd45fb5db0b/sklearn/multioutput.py#L621)

2. [naive_bayes](https://github.com/scikit-learn/scikit-learn/blob/3d4ee9e9dc384ea11724e270b51bd26a1f227f0c/sklearn/naive_bayes.py#L249)

3. [kmeans](https://github.com/scikit-learn/scikit-learn/blob/09a72930b8ec32efbfb5f9ad14701e52f482a996/sklearn/cluster/_kmeans.py#L889-L896)

4. [_kernel_pca](https://github.com/scikit-learn/scikit-learn/blob/bbdb2eff9b877c0ae00ed9854099b92119504f62/sklearn/decomposition/_kernel_pca.py#L481
)
 

Please let me know if the approach i am taking in this PR is not feasible. I will  try to explore other ways to go about solving this problem.

mmccarty:
@TomAugspurger - Any further feedback or guidance here? I don't see a way to expose `check_array`'s kwargs without changes to `sklearn`. IMO, it seems reasonable to handle this case in `dask-ml`.

cc @jrbourbeau @betatim for vis

betatim:
I think what is happening is that `predict()` (and co) are being called with an empty input that contains zero samples. It seems sensibly for the scikit-learn estimators to consider that an error. So I agree with Mike that this is something that dask-ml should handle. Probably by not calling `predict()`, `transform()` etc when a partition is empty and instead returning what ever is the expected result of making a prediction on an empty array (I'm not sure what this should be, `None` also an empty array?).

VibhuJawa:
> instead returning what ever is the expected result of making a prediction on an empty array (I'm not sure what this should be, `None` also an empty array?).

In this PR, 
1. If the output is supposed to be arrays  , I create empty arrays (for both sparse and dense arrays)

2. If the output is supposed to be dataframes in this PR For dataframes I return an empty dataframes 

https://github.com/dask/dask-ml/blob/28b97e0e903e09b3af4ee3c1dbbeadc7b0b57915/dask_ml/wrappers.py#L661-L688

This also matches what cuML does  (which returns an empty series) . See below:
```python
>>> type(reg)
<class 'cuml.linear_model.logistic_regression.LogisticRegression'>
>>> reg.predict(X_new.iloc[:0])
Series([], dtype: float32)

```

Issue #860, opened 2021-10-11
DaskML vs. Sklearn LogisticRegression. Coefficients seem different. 


koaning:
The coefficients from the `daskml.linear_model.LogisticRegression` seem different from the `sklearn.linear_model.LogisticRegression`. 

The script below demonstrates a benchmark that shows this. 

```python
# %pip install dask_ml memo
import time
import numpy as np
from dask_ml.datasets import make_classification
from dask_glm.linear_model import LogisticRegression as DaskLogReg
from sklearn.linear_model import LogisticRegression as SklearnLogReg 
from memo import grid, memlist, memfunc

data_hindsight = []

@memfunc(print)
@memlist(data=data_hindsight)
def run_experiment(size=1_000, chunksize=1_000):
    X, y = make_classification(n_samples=size, n_features=20, n_informative=10, chunksize=chunksize)

    # Run everything in Dask
    tic = time.time()
    mod_dask = DaskLogReg()
    mod_dask.fit(X, y)
    dask_time = time.time() - tic 

    # Run everything in Scikit-Learn
    X_np, y_np = np.asarray(X), np.asarray(y)
    mod_sklearn = SklearnLogReg()
    tic = time.time()
    mod_sklearn.fit(X_np, y_np)
    sklearn_time = time.time() - tic
    
    # Return stats that are of interest
    return {
        'time_dask': dask_time,
        'time_sklearn': sklearn_time, 
        'acc_dask': (mod_dask.predict(X) == y).mean().compute(),
        'acc_sklearn': (mod_sklearn.predict(X_np) == y_np).mean(),
        'coef_diff': np.abs(mod_sklearn.coef_ - mod_dask.coef_).sum() # This is what I'm interested in understanding.
    }

for size in [1000, 2000]:
    for chunksize in [500, 1000]:
        run_experiment(size=size, chunksize=chunksize)
```

The output in `data_hindsight` looks like this: 

```
{'size': 1000, 'chunksize': 500, 'time_dask': 1.2568867206573486, 'time_sklearn': 0.003863811492919922, 'acc_dask': 0.795, 'acc_sklearn': 0.795, 'coef_diff': 7.47177509238394}
{'size': 1000, 'chunksize': 1000, 'time_dask': 0.09700465202331543, 'time_sklearn': 0.005501270294189453, 'acc_dask': 0.784, 'acc_sklearn': 0.784, 'coef_diff': 8.645457699298978}
[...]

PR #914, opened 2022-04-01
Update coefficient assignment

kchare:
## What
This PR contains three primary additions:

1. A bug fix for coefficient assignment in `dask_ml.linear_model.utils` to update the ``add_intercept`` functionality to be consistent between NumPy and Dask arrays. Previously, this was done differently, resulting in `.coef_` and `.intercept_` terms that were not equal for models fit on the same data with a NumPy array or Dask array.
2. Tests to ensure that models fit on the same data in NumPy array and Dask arrays will yield the same result.
3. Tests of model coefficients against `scikit-learn` for this same data, ensuring that coefficients match for in-memory data.

## Why
#860 
This issue traces the bug, and (a) suggests a fix, (b) proposes tests to tie out Dask ML linear models between input data types, and (c) tests for equality of model coefficients between Dask ML and ``scikit-learn``.


tests/linear_model/test_glm.py
```diff
@@ -203,3 +220,79 @@ def test_logistic_predict_proba_shape():
...
+    ]
+)
+def test_model_against_sklearn(reg_type, skl_reg_type, skl_params, data_generator, request):
+    """
+    Tests accuracy of model predictions
+
+    Notes:
+        1. Dask ML 'newton' solver does not implement a penalty, only
+            set arbitarily high for clarity of non-regularization.
+        2. This method only works for `fit_intercept=False`
```
stsievert, line 254: Why? `fit_intercept=True` below.

tests/linear_model/test_glm.py
```diff
@@ -203,3 +220,79 @@ def test_logistic_predict_proba_shape():
...
+    (LogisticRegression, sklLogisticRegression, {'penalty' : 'none'}, "single_chunk_classification"),
+    (PoissonRegression, sklPoissonRegressor, {'alpha' : 0}, "single_chunk_count_classification")
+    ]
+)
+def test_model_against_sklearn(reg_type, skl_reg_type, skl_params, data_generator, request):
+    """
+    Tests accuracy of model predictions
+
+    Notes:
+        1. Dask ML 'newton' solver does not implement a penalty, only
```
stsievert, line 252: Why does `solver='newton'` have to be used? `penalty='l2'` is passed to the Dask-ML optimization but [is unused in Dask-GLM][2] [and Dask-ML][3].

I think testing linear models without regularization is fine. In that case, the `penalty` and `C` values can be ignored for `solver='newton'`.

[2]:https://github.com/dask/dask-glm/blob/1daf4c5de0f44eb9405a40afb5506d4d821d407f/dask_glm/algorithms.py#L143
[3]:https://github.com/dask/dask-ml/blob/1e811ce48951a45a54859aca387510abdc085722/dask_ml/linear_model/glm.py#L147-L148

tests/linear_model/test_glm.py
```diff
@@ -203,3 +220,79 @@ def test_logistic_predict_proba_shape():
...
+    X, y = dataset_function(n_samples=100, n_features=20, n_informative=10, chunks=100, random_state=0)
+    X_np, y_np = X.compute(), y.compute()
+    
+    skl_model = skl_reg_type(fit_intercept=True, **skl_params)
+    skl_model.fit(X_np, y_np)
+
+    dask_ml_model = reg_type(fit_intercept=True, solver='newton', penalty='l2', C=1e8)
+    dask_ml_model.fit(X, y)
+
+    np.testing.assert_allclose(skl_model.coef_.flatten(), dask_ml_model.coef_, rtol=1e-3, atol=1e-3)
```
stsievert, line 287: Maybe add a test to make sure that both elements are approximately equal?

```suggestion
    np.testing.assert_allclose(skl_model.coef_.flatten(), dask_ml_model.coef_, rtol=1e-3, atol=1e-3)
    
    # Test that both coefs are within an epsilon-ball
    est, truth = dask_mo_model.coef_, skl_model.coef_.flatten()
    rel_error = np.norm(est - truth) / np.norm(truth)
    assert rel_error < 1e-6
```

 Relative error is a weaker condition often used in convergence proofs to ensures that both elements are close enough. I think that's what this test expresses but it'd be nice to see it more explicitly.

tests/linear_model/test_glm.py
```diff
@@ -203,3 +220,79 @@ def test_logistic_predict_proba_shape():
...
+
+def test_poisson_regressor_against_sklearn(single_chunk_count_classification):
+    X, y = single_chunk_count_classification
+    skl_model = sklPoissonRegressor(alpha=0, fit_intercept=True)
```
stsievert, line 291: Is `alpha = 0` to make sure that the regression is not regularized for both Dask-ML and Scikit-learn?

(not important: I think `C = 1 / alpha` or `C = 1 / (n * alpha)` with `len(y) == n`).

tests/linear_model/test_glm.py
```diff
@@ -203,3 +220,79 @@ def test_logistic_predict_proba_shape():
...
+    skl_model.fit(X, y)
+    dask_ml_model = PoissonRegression(fit_intercept=True, solver='newton', penalty='l2', C=1e8)
+    dask_ml_model.fit(X, y)
+    np.testing.assert_allclose(skl_model.coef_.flatten(), dask_ml_model.coef_, rtol=1e-3, atol=1e-3)
+    
+    
+
```
stsievert, line 298: NIt: I think the python linter Black needs to be run (I think the CI tests will fail).

tests/linear_model/test_glm.py
```diff
@@ -173,6 +182,14 @@ def test_add_intercept_raises_chunks():
 
     assert m.match("Chunking is only allowed")
 
+def test_add_intercept_ordering():
```
stsievert, line 185: :+1: :+1:

tests/linear_model/test_glm.py
```diff
@@ -6,6 +6,11 @@
 from dask.dataframe.utils import assert_eq
 from dask_glm.regularizers import Regularizer
 from sklearn.pipeline import make_pipeline
+from sklearn.linear_model import (
```
stsievert, line 9: Style nit: `import sklearn.linear_model as sml`?

stsievert (commented):
:+1: :+1: I'm glad to see this fix. I'm really glad to see these tests! They're long overdue.

I have some questions about the tests (and some style nits):

tests/linear_model/test_glm.py
kchare: Yes, the goal of this test is to ensure that the coefficients are equal with some small amount of relative error. Two differences between the `np.testing.assert_allclose` and relative error calculation that you suggest is that (1) the NumPy utility computes the error element-wise and bounds the maximum error for each element, and (2) the `atol` condition allows for an absolute tolerance when the relative tolerance comes from numeric instability (i.e. when the difference is compared to a near-zero coefficient and has high relative error but absolute error at an order of magnitude of `1e-10`).

I am happy to update the relative error to be calculated via that vector norm rather than element-wise, but wanted to call out that difference before updating the implementation.

tests/linear_model/test_glm.py
kchare: Yes, this is exactly the reasoning in this test. The `scikit-learn` documentation indicates that `skl.linear_model.PoissonRegressor` with `alpha=0` is equivalent to the unpenalized GLM ([link](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.PoissonRegressor.html)).

tests/linear_model/test_glm.py
kchare: Done.

tests/linear_model/test_glm.py
kchare: I ran the Black linter over these files before my last commit, and will continue to do so for future updates.

tests/linear_model/test_glm.py
kchare: This has been removed and updated in the test code.

tests/linear_model/test_glm.py
stsievert: I think `N` should be much larger than `p`. `N / p == 5` seems reasonable. Having `N == p` sounds like a bad idea, especially without regularization and only 10 informative features.

I don't think tests should use a special dataset. I think changing the error tolerance is better, especially if the relative error is used as an error condition (I can't find the comment, but it's `rel_error = LA.norm(est - truth) / LA.norm(truth)`). Relative error is used as a convergence criteria in most textbooks, so I think it's fine to include it in the tests too.

tests/conftest.py
```diff
@@ -146,3 +146,37 @@ def scheduler(request):
...
+
+@pytest.fixture
+def medium_size_regression():
+    """X, y pair for regression with N >> p.
+
+    There are many more samples in this problem than there are
+    features. Useful for testing stability of solutions.
+    """
+    X, y = make_regression(
+        chunks=100, n_samples=500, n_features=100, n_informative=10, random_state=0
```
stsievert, line 159:
```suggestion
        chunks=100, n_samples=500, n_features=100, n_informative=100, random_state=0
```

This seems like a better idea without regularization. It also wouldn't be a bad idea to reduce the number of features.

tests/conftest.py
```diff
@@ -146,3 +146,37 @@ def scheduler(request):
...
+    greater stability of the solutions. Useful for testing
+    the accuracy of solvers.
+    """
+    sample_size = 2_000
+    n_features = 100
+    X, y = make_counts(
+        chunks=100,
+        n_samples=sample_size,
+        n_features=n_features,
+        n_informative=n_features // 10,
```
stsievert, line 178:
```suggestion
        n_informative=n_features,
```

tests/linear_model/test_glm.py
```diff
@@ -203,3 +218,115 @@ def test_logistic_predict_proba_shape():
...
+def test_poisson_regressor_against_sklearn(single_chunk_count_classification):
+    X, y = single_chunk_count_classification
+    skl_model = slm.PoissonRegressor(alpha=0, fit_intercept=True)
+    skl_model.fit(X, y)
+    dask_ml_model = PoissonRegression(
+        fit_intercept=True, solver="newton", penalty="l2", C=1e8
+    )
+    dask_ml_model.fit(X, y)
+    np.testing.assert_allclose(
+        skl_model.coef_.flatten(), dask_ml_model.coef_, rtol=1e-3, atol=1e-3
```
stsievert, line 331: When I look at the tests, I can't really tell if the coefficients are "close enough" with the specification of `atol` and `rtol` (e.g., what if all the coefficients are less than `atol`?).

It'd also be great if you could add some relative error condition in your other tests (it's also the test used in many optimization textbooks). This code would convince me that the coefficients are "close enough":

``` python
import numpy.linalg as LA
est = np.hstack((dask_ml_model.coef_, dask_model.intercept_))
truth = np.hstack((skl_model.coef_, slm.intercept_))

rel_error = LA.norm(est - truth) / LA.norm(truth)
assert rel_error < 1e-3
```

kchare: That makes sense to me. I have converted all of the tests to use the relative error, as you describe above. 

There is one remaining issue that seems to be related to the gradient descent results that affects both the `PoissonRegressor` results and the `LogisticRegression` results.

First, When running the test with a relative error tolerance of `1e-3`, it typically does not pass for the `PoissonRegressor` with gradient descent (Newton's method and LBFGS achieve the same results). I have spent a while debugging the data creation functions, but have not been able to reliably get the test to pass with the `1e-3` relative error tolerance. I suspect that this is due to having a large number of features, as increasing `N / p` from `5` to `10` does decrease the error, and potentially due to a non-convex loss function for this data leading to difficulty for the gradient descent algorithm. `scikit-learn` defaults to LBFGS for `sklearn.linear_model.PoissonRegressor` with no alternative option. For the `LogisticRegression`, the test of the relative error of the coefficients using `1e-3` is similarly too large to pass (as the test is written, it is approx. `0.14`) when using gradient descent. It converges to the `scikit-learn` results for both Newton's method and LBFGS.

Maybe the best strategy is to remove gradient descent from the tests and supply some sort of convergence warning when using vanilla GD?

I appreciate all of the time that you have put into looking over these results!

tests/linear_model/test_glm.py
```diff
@@ -203,3 +219,121 @@ def test_logistic_predict_proba_shape():
...
+    """Tests that models return same coefficients and intercepts with array types"""
+    X, y = request.getfixturevalue(data_generator)
+    np_mod, da_mod = reg_type(fit_intercept=True), reg_type(fit_intercept=True)
+    da_mod.fit(X, y)
+    np_mod.fit(X.compute(), y.compute())
+    da_coef = np.hstack((da_mod.coef_, da_mod.intercept_))
+    np_coef = np.hstack((np_mod.coef_, np_mod.intercept_))
+
+    rel_error = LA.norm(da_coef - np_coef) / LA.norm(np_coef)
+    assert rel_error < 1e-8
```
stsievert, line 242: Thanks for including the relative error. I think it'd be best to include the relative error test alongside the `allclose` test; that way it's pretty hard to question.

tests/linear_model/test_glm.py
```diff
@@ -203,3 +219,121 @@ def test_logistic_predict_proba_shape():
...
+
+
+@pytest.mark.parametrize("solver", ["newton", "gradient_descent", "lbfgs"])
+@pytest.mark.parametrize("fit_intercept", [True, False])
+@pytest.mark.parametrize(
+    "reg_type, skl_reg_type, skl_params, data_generator",
+    [
+        (LinearRegression, slm.LinearRegression, {}, "medium_size_regression"),
```
stsievert, line 250: Style nit: maybe this would be a bit simpler and a bit shorter:

``` python
# fmt: off
@pytest.mark.parametrize("est, skl_params, data", [
    ("LinearRegression", {}, medium_size_regression()),
    ("LogisticRegression", {"penalty": None}, single_chunk_classification()),
])
# fmt: on
def test_model_against_sklearn(est, skl_params, data):
    X, y = data
    EstDask = getattr(dask_ml.linear_regression, est)
    EstSklearn = getattr(scikit_learn.linear_model, est)
    dask_est = EstDask(...)
    skl_est = EstSklearn(...)
    ...
```

tests/linear_model/test_glm.py
```diff
@@ -203,3 +219,121 @@ def test_logistic_predict_proba_shape():
...
+        (LinearRegression, slm.LinearRegression, {}, make_regression),
+        (
+            LogisticRegression,
+            slm.LogisticRegression,
+            {"penalty": "none"},
+            make_classification,
+        ),
+    ],
+)
+def test_model_coef_against_sklearn(
```
stsievert, line 298: Maybe this test should be combined with `test_model_against_sklearn`?

Also, relative error tests are best suited for tests on the coefficients (and not so much for binary predictions).

tests/linear_model/test_glm.py
```diff
@@ -203,3 +219,121 @@ def test_logistic_predict_proba_shape():
...
+
+    est, truth = np.hstack((dask_ml_model.intercept_, dask_ml_model.coef_)), np.hstack(
+        (skl_model.intercept_, skl_model.coef_.flatten())
+    )
+    rel_error = LA.norm(est - truth) / LA.norm(truth)
+    assert rel_error < 1e-3
+
+
+@pytest.mark.parametrize("solver", ["newton", "gradient_descent", "lbfgs"])
+def test_poisson_regressor_against_sklearn(single_chunk_count_classification, solver):
```
stsievert, line 326: Why isn't this test part of `test_model_coef_against_sklearn`?

dask_ml/datasets.py
```diff
@@ -64,12 +64,14 @@ def make_counts(
     rng = dask_ml.utils.check_random_state(random_state)
 
     X = rng.normal(0, 1, size=(n_samples, n_features), chunks=(chunks, n_features))
-    informative_idx = rng.choice(n_features, n_informative, chunks=n_informative)
-    beta = (rng.random(n_features, chunks=n_features) - 1) * scale
+    informative_idx = rng.choice(
+        n_features, n_informative, chunks=n_informative, replace=False
+    )
+    beta = (rng.random(n_features, chunks=n_features) - 0.5) * 2 * scale
```
stsievert, line 70: Why is this change required?

stsievert (commented):
This PR looks pretty good. A couple nitpicky comments below.

I'm glad to see the tests in this PR. They ensure that `add_intercept` works the same for NumPy and Dask arrays, and also that various Dask-ML linear estimators produce the same results as the corresponding Scikit-learn estimators.

dask_ml/datasets.py
kchare: I had changed this due to some issues with convergence for the Poisson Regressor. I am not an expert in Poisson regression, but this previously had `beta` values between `[-1, 0)` and the update brought that to `[-1, 1]`. Reverting it back makes no difference in the updated tests, so I will revert the change.

tests/linear_model/test_glm.py
kchare: Added back.

tests/linear_model/test_glm.py
kchare: Great. I have combined the tests into a single parameterized test.

tests/linear_model/test_glm.py
kchare: I adopted this set of tests with one minor difference -- when combining the tests for coefficient as well as `PoissonRegression`, there is actually a naming difference between `dask-ml` and `scikit-learn`. I have added a second parameter and a `try-except` block to handle that difference.

tests/linear_model/test_glm.py
kchare: Combined into a single test.

kchare:
As I consolidated the tests to address your comments, I realized that `scikit-learn` actually assigns an intercept value of `0.0` when `fit_intercept=False` (see [here](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/linear_model/_base.py#L388)). To ensure that the results are the same between `scikit-learn` and `dask-ml`, I have also addressed that difference in the latest commit.

dask_ml/datasets.py
```diff
@@ -64,12 +64,14 @@ def make_counts(
...
-    informative_idx = rng.choice(n_features, n_informative, chunks=n_informative)
+    informative_idx = rng.choice(
+        n_features, n_informative, chunks=n_informative, replace=False
+    )
     beta = (rng.random(n_features, chunks=n_features) - 1) * scale
 
     informative_idx, beta = dask.compute(informative_idx, beta)
 
-    z0 = X[:, informative_idx].dot(beta[informative_idx])
+    z0 = X[:, informative_idx].dot(beta[informative_idx]) + 0.5
```
stsievert, line 74: Why are changes to this file required?

tests/linear_model/test_glm.py
```diff
@@ -42,6 +45,11 @@ def get_params(self, deep=True):
 X, y = make_classification(chunks=50)
 
 
+@pytest.fixture()
+def glm_regression_data():
```
stsievert, line 49: I don't think this is ever used?

stsievert (commented):
This is looking pretty good! I've got a couple nits to pick.

tests/linear_model/test_glm.py
```diff
@@ -203,3 +220,79 @@ def test_logistic_predict_proba_shape():
     lr.fit(X, y)
     prob = lr.predict_proba(X)
     assert prob.shape == (100, 2)
+
+
+@pytest.mark.parametrize(
+    "reg_type,data_generator",
+    [
+        (LinearRegression, "single_chunk_regression"),
```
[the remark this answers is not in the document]
kchare: When I tried updating to this, I received an error due to the fact that e.g. `single_chunk_regression` is a fixture and cannot be called directly. I followed the development pattern [here](https://github.com/pytest-dev/pytest/issues/349#issuecomment-1072791330). This same thread has some external packages that could be options to implement this behavior such as `pytest-cases`, but did not want to introduce an additional dependency.

dask_ml/datasets.py
kchare: This adds an explicit intercept term, as the `X` array does not have a constant term for the intercept. I am happy to remove it, however, as it does not change the results.

tests/linear_model/test_glm.py
kchare: Thanks for pointing that out. I have now removed it.

tests/linear_model/test_glm.py
```diff
@@ -203,3 +220,79 @@ def test_logistic_predict_proba_shape():
...
+    ``np.testing.assert_allclose``. This ensures that the model coefficients
+    match up with SK Learn.
+    """
+    X, y = request.getfixturevalue(data_generator)
+    EstDask = getattr(dask_ml.linear_model, est)
+    # sklearn uses 'PoissonRegressor' while dask-ml uses 'PoissonRegression'
+    try:
+        EstSklearn = getattr(sklearn.linear_model, est)
+    except AttributeError:
+        EstSklearn = getattr(sklearn.linear_model, skl_reg_type)
```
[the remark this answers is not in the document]
kchare: I have updated the code to reflect this change. I agree that it is still a bit hacky. I have not thought of a better way to cleanly handle the difference in names, however. This method seems to be more efficient than a dictionary with keys / values as the `EstDask` and `EstSklearn`. If the set of GLMs included were to be expanded though, that may not be the case.

dask_ml/datasets.py
```diff
@@ -64,12 +64,14 @@ def make_counts(
     rng = dask_ml.utils.check_random_state(random_state)
 
     X = rng.normal(0, 1, size=(n_samples, n_features), chunks=(chunks, n_features))
-    informative_idx = rng.choice(n_features, n_informative, chunks=n_informative)
+    informative_idx = rng.choice(
+        n_features, n_informative, chunks=n_informative, replace=False
```
stsievert, line 68: This is a good change.

I think this PR will get merge faster if all the changes in this file are left to another PR.

tests/linear_model/test_glm.py
```diff
@@ -203,3 +212,78 @@ def test_logistic_predict_proba_shape():
...
+    da_mod.fit(X, y)
+    np_mod.fit(X.compute(), y.compute())
+    da_coef = np.hstack((da_mod.coef_, da_mod.intercept_))
+    np_coef = np.hstack((np_mod.coef_, np_mod.intercept_))
+
+    rel_error = LA.norm(da_coef - np_coef) / LA.norm(np_coef)
+    assert rel_error < 1e-8
+
+
+# fmt: off
```
stsievert, line 238:
```suggestion
```

stsievert (approved):
LGTM after the changes below are implemented!

dask_ml/datasets.py
stsievert: I think it'd be good to do some better error handling in this function (like making sure that `n_informative < n_features`). I think a future PR would be a great place for that alongside workarounds for the items in https://github.com/dask/dask-ml/pull/914#discussion_r854348163 if they're still relevant.

Past the changes in this file, this PR LGTM!

kchare: Awesome, thank you! I have reverted the two changes from the `datasets.py` file. When I did so, I had to make one small change to the `atol` condition of the tests to make it pass for the `PoissonRegressor`. The original condition checked to the `1e-4`, and the maximum with the dataset update was ~0.00014, so I updated the condition to restrict only to `2e-4`.

A future PR for the other changes to the `datasets.py` file sounds like a good idea to me. I would be happy to work on that at some point, but may not be able to do so in the short term.

stsievert:
The tests pass with the 3.7 and 3.8 environments. The doc error isn't relevant (unexpected warning), and the 3.9 environment is failing with these (not relevant) errors/warnings:

* `TypeError: _fit() got an unexpected keyword argument 'return_counts'`
* AttributeError: 'OneHotEncoder' object has no attribute '_infrequent_indices'
* `FutureWarning: if_delegate_has_method was deprecated in version 1.1 and will be removed in version 1.3. Use if_available instead.`
* `FutureWarning: The loss 'log' was deprecated in v1.1 and will be removed in version 1.3. Use `loss='log_loss'` which is equivalent.
* `ValueError: ndarray is not C-contiguous`

stsievert:
I've verified that the CI errors on this PR are not relevant (see https://github.com/dask/dask-ml/pull/914#issuecomment-1133519952). I will squash and merge this PR next weekend unless I hear otherwise.

TomAugspurger:
Let's merge this and see if CI passes on `main`.

PR #910, opened 2022-03-24
Replace _is_pairwise import with _is_pairwise_metric for compatibility with scikit-learn 1.1

cmarmo:
Dear maintainers,

this pull request replaces the import of `_is_pairwise` with the import of `_is_pairwise_metric` for compatibility with scikit-learn 1.1 (hopefully to be released soon). About backward compatibility: `_is_pairwise` was deprecated in 0.24 (December 2020).
Fixes #908.

Thanks for considering it.

PR #921, opened 2022-04-19
Fix precomputed kernel test with _safe_tags.

cmarmo:
@TomAugspurger sorry to bother you again.
I realized that one of the failures in #910 was indeed related to my changes.
The failure is related to #771 and its fix in  #772.
As the tag "pairwise" is not defined the test fails again.
I have replaced ``is_pairwise_metric`` with the ``_safe_tags()`` function (see [here](https://github.com/scikit-learn/scikit-learn/blob/e61040c14e59bec3b0d275d40a43e10cf641af32/sklearn/utils/_tags.py#L25)) which falls back to default when the tags is not defined.
Sorry for not noticing that before.

### DIFF
--- a/dask_ml/model_selection/_search.py
+++ b/dask_ml/model_selection/_search.py
@@ -15,12 +15,7 @@ from dask.delayed import delayed
 from dask.distributed import as_completed
 from dask.utils import derived_from
 from sklearn import model_selection
-from sklearn.base import (
-    BaseEstimator,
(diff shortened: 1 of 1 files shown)


cmarmo:
I don't understand what is wrong with with black....

TomAugspurger:
Thanks, I think I saw another PR fixing the linting issue.

PR #924, opened 2022-04-28
[WIP]Fix Dask-ML CI

VibhuJawa:
This PR fixes CI 

- [ ] Fix Linting
- [x] Fix Docs Build

**Error Details :** 

This code (see [link](https://github.com/dask/dask-ml/blob/80eb0d81b22df3762706384c3ca1c0be4137e591/dask_ml/datasets.py#L373-L375))   in dask-ml   calls dask.compute  with scheduler="single-threaded"  .  

 This  causes failures in  dask-ml docs  builds due to it raising the below warning  ```UserWarning: Running on a single-machine scheduler when a distributed client is active might lead to unexpected results  and thus is blocking the CI ```
 
**Fix Details:**
Remove the `dask.compute` call  as we dont really need to compute it right away

**Todo:**
- [ ] Address the `test_train_test_split_dask_dataframe_rng` test that popped up

PR #909, opened 2022-03-24
Black formatting updates

mmccarty:

PR #929, opened 2022-05-20
Fixing CI

mmccarty:
This PR follows on from #909 and fixes the CI build with following changes

- [x] adds `:okwarning:` and `:okexcept:` options to sphinx ipython blocks
- [x] ~~pin versions of `sklearn` and `scipy` that are currently compatible with `dask-ml`~~
- [x] upgrade test and associated code to support both sklearn 1.0 and 1.1
- [x] removes `xgboost` from `ci/environment-docs.yml` 

If this PR is acceptable, #924 and #931 can be closed.


mmccarty:
Need to add checks to swtich between scikit-learn 1.0.* and 1.1.

mmccarty:
Need to add checks to swtich between scikit-learn 1.0.* and 1.1.

dask_ml/metrics/regression.py
```diff
@@ -16,7 +17,9 @@ def _check_sample_weight(sample_weight: Optional[ArrayLike]):
 def _check_reg_targets(
     y_true: ArrayLike, y_pred: ArrayLike, multioutput: Optional[str]
 ):
-    if multioutput is not None and multioutput != "uniform_average":
+    if multioutput is not None and (
+        is_dask_collection(multioutput) or multioutput != "uniform_average"
+    ):
```
charlesbluca, line 22: Not immediately clear to me why we need this extra check? Can there be a case where `multioutput` is a Dask collection that is also equal to `"uniform_average"`?

dask_ml/_compat.py
```diff
@@ -29,6 +29,12 @@
 PANDAS_1_2_0 = PANDAS_VERSION > packaging.version.parse("1.2.0")
 WINDOWS = os.name == "nt"
 
+# 'log_loss' is preferred as of scikit-learn 1.1
+if SK_VERSION >= packaging.version.parse("1.1"):
```
charlesbluca, line 33: Wonder if we should set the result of this conditional to a variable (i.e. `SK_1_1`) if we intend to use it elsewhere in the codebase?

ci/environment-3.8.yaml
```diff
@@ -1,4 +1,4 @@
-name: dask-ml-3.8
+name: dask-ml-latest
```
wence-, line 1: Is this name change intentional, and if so, why not for the 3.9 environment yaml?

wence- (commented):
Minor quibble

ci/environment-3.8.yaml
mmccarty: Nope. Thanks!

dask_ml/_compat.py
jrbourbeau: +1 -- let's add an `SKLEARN_1_1_X` variable a lines above with the other similar variables (e.g. `PANDAS_1_2_0`). Also, for these version compatibility variables it helps to use the full X.Y.Z version -- that way it's more clear when we can remove related compatibility code

docs/source/conf.py
```diff
@@ -46,11 +46,11 @@
 ]
 
 intersphinx_mapping = {
-    "python": ("https://docs.python.org/3.6", None),
-    "sklearn": ("http://scikit-learn.org/stable/", None),
+    "python": ("https://docs.python.org/3.10", None),
```
jrbourbeau, line 49: I _think_ this should be slightly more future proof by pointing to whatever the latest Python release is 

```suggestion
    "python": ("https://docs.python.org", None),
```

dask_ml/metrics/regression.py
mmccarty: This is an argument for [sklearn.metric.r2_score](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html#sklearn-metrics-r2-score), where `multioutput` can be a str or array-like. In Dask-ML there is [a test](https://github.com/dask/dask-ml/blob/main/tests/metrics/test_regression.py#L113) that specifically passes a Dask Array to `multioutput` and it was failing because Dask collections are ambiguous for truthy.

docs/source/hyper-parameter-search.rst
```diff
@@ -420,6 +420,7 @@ This section uses :class:`~dask_ml.model_selection.HyperbandSearchCV`, but it ca
 also be applied to to :class:`~dask_ml.model_selection.IncrementalSearchCV` too.
 
 .. ipython:: python
+   :okwarning:
```
VibhuJawa, line 423: We should open issues corresponding to the problems that we are ignoring by this. Essentially to ensure that we have it in the backlog. 


```
informative_idx, beta = dask.compute(
        informative_idx, beta, scheduler="single-threaded"
    )
```

VibhuJawa, line 423: We should open issues corresponding to the problems that we are ignoring by this. Essentially to ensure that we have it in the backlog. 


```
informative_idx, beta = dask.compute(
        informative_idx, beta, scheduler="single-threaded"
    )
```

mmccarty: :100: agreed. @VibhuJawa - Do you have a reproducer? I wasn't able to get one outside of sphinx ipython directives.

wence-:
```python
import dask.distributed # sets dask.config["scheduler"] == "dask.distributed"
import dask.array as da
import dask

a = dask.compute(da.arange(10), scheduler="single-threaded")
```

`make_classification` explicitly calls (as @VibhuJawa indicates) `dask.compute` setting the scheduler to `"single-threaded"`, which then mismatches with the scheduler set up by importing `dask.distributed`.

Plausibly this is the correct fix

```diff
diff --git a/dask_ml/datasets.py b/dask_ml/datasets.py
index a561ee0d..3a5f5e99 100644
--- a/dask_ml/datasets.py
+++ b/dask_ml/datasets.py
@@ -370,9 +370,7 @@ def make_classification(
     informative_idx = rng.choice(n_features, n_informative, chunks=n_informative)
     beta = (rng.random(n_features, chunks=n_features) - 1) * scale
 
-    informative_idx, beta = dask.compute(
-        informative_idx, beta, scheduler="single-threaded"
-    )
+    informative_idx, beta = dask.compute(informative_idx, beta)
 
     z0 = X[:, informative_idx].dot(beta[informative_idx])
     y = rng.random(z0.shape, chunks=chunks[0]) < 1 / (1 + da.exp(-z0))
```

Although that backs out https://github.com/dask/dask-ml/pull/288/commits/cb5c2eec456e70a4fb39ab5d0a1c10ab28e63d86 which looks like it was a deliberate choice at the time.

dask_ml/_compat.py
```diff
@@ -29,6 +29,14 @@
 PANDAS_1_2_0 = PANDAS_VERSION > packaging.version.parse("1.2.0")
 WINDOWS = os.name == "nt"
 
+SKLEARN_1_1_X = SK_VERSION >= packaging.version.parse("1.1")
```
wence-, line 32: This name is misleading, since it is `True` for 1.2 (say). I realise this mimics the naming above, so I guess fine.

wence- (approved):
Modulo minor comments, I think this (from my inexpert eye) looks fine.

docs/source/hyper-parameter-search.rst
VibhuJawa: @mmccarty, **Minimal Example**

```python

from dask.distributed import Client
import dask_ml.datasets
 
client = Client()


X,y = dask_ml.datasets.make_classification_df(chunks=100)
```
```python

/datasets/vjawa/miniconda3/envs/dask-ml-dev/lib/python3.8/site-packages/dask/base.py:1282: UserWarning: Running on a single-machine scheduler when a distributed client is active might lead to unexpected results.
  warnings.warn(


```


Please note that there is a call in  `k-means` that will need to be eventually fixed too. 

https://github.com/dask/dask-ml/blob/30b3deab19755ecd45156ef9cac9fee646615100/dask_ml/cluster/k_means.py#L472-L475

VibhuJawa: @wence- , I tried fixing it by removing the explicit cals with https://github.com/dask/dask-ml/pull/924 but that causes issues with key-names downstream [here](https://github.com/dask/dask-ml/blob/67d28b15dfff7869e9a04def203aa129a3540b27/tests/model_selection/test_split.py#L225-L226) . I suggest going with @mmccarty's fix to unblock the CI and then we can tackle this issue from the backlog.

mmccarty: @VibhuJawa - Ah, of course, the warning is reproducible. Sorry, I was thinking about the key error. Opened an issue #933

dask_ml/_compat.py
TomAugspurger: This is the convention used by packages like `six`.

TomAugspurger (approved):

TomAugspurger:
Thanks!

mmccarty:
Thanks @TomAugspurger! We have one issue #933 to follow up, which I'll look at soon. It may lead to more issues but we can track them there.

PR #957, opened 2023-01-10
Fix tests for latest versions

mmccarty:
adding temporary upper pin on scikit-learn


jrbourbeau (commented):
Thanks for working on this @mmccarty! Feel free to ping me if I can be of use

mmccarty:
@jrbourbeau - Thanks! Ready for a review.

.github/workflows/tests.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
       matrix:
         # os: ["windows-latest", "ubuntu-latest", "macos-latest"]
         os: ["ubuntu-latest"]
-        python-version: ["3.7", "3.8", "3.9"]
+        python-version: ["3.8", "3.9", "3.10"]
```
jrbourbeau, line 13: I think we can do this too 

```suggestion
        python-version: ["3.8", "3.9", "3.10", "3.11"]
```

ci/environment-3.7.yaml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - pytest-cov
```
jrbourbeau, line 17: Since we're dropping Python 3.7 from CI we should be able to safely drop this file too

ci/environment-3.8.yaml
```diff
@@ -17,7 +17,7 @@ dependencies:
   - pytest-cov
   - pytest-mock
   - python=3.8.*
-  - scikit-learn>=1.0.0
+  - scikit-learn <=1.1.3
```
jrbourbeau, line 20: Why this version constraint?

ci/environment-docs.yaml
```diff
@@ -21,7 +21,7 @@ dependencies:
   - psutil
   - python=3.8
   - sortedcontainers
-  - scikit-learn>=1.0.0
+  - scikit-learn<=1.1.*
```
jrbourbeau, line 24: This version constraint is different than the `scikit-learn <=1.1.3` one used elsewhere. Just wanted to double check this is intentional

jrbourbeau (commented):
Thanks @mmccarty! Just a few minor comments / questions

.github/workflows/tests.yaml
mmccarty: Not yet. We need numba/numba#8304

ci/environment-3.8.yaml
mmccarty: scikit-learn 1.2.0 has 123 test failures. I'm going follow up is that.

ci/environment-docs.yaml
mmccarty: Nope! Thanks

ci/environment-3.8.yaml
jrbourbeau: Seems like a good reason! Thanks for clarifying

.github/workflows/tests.yaml
jrbourbeau: Ah, I see. In that case it seems strange to me that we're adding a `ci/environment-3.11.yaml` file.

mmccarty: Good point. I can remove it from this PR.

mmccarty:
@jrbourbeau Anything further? I'm now working on scikit-learn 1.2.0 support.

jrbourbeau (approved):
LGTM, thanks again @mmccarty

hungbui-io:
Hi dask-ml team, I am wondering when dask-ml new version will be released, the one that supports python 3.10? I am trying to use latest dask-ml, dask, and dask distributed with python 3.10, but facing some library importing issues.
Any information would be much appreciate!

PR #959, opened 2023-01-13
updates for scikit-learn 1.2.0

mmccarty:
- [x] Update tests and code to handle `scikit-learn==1.2.0` deprecation warnings
- [x] Update environment files
- [x] Update package requirements

### DIFF
--- a/dask_ml/cluster/k_means.py
+++ b/dask_ml/cluster/k_means.py
@@ -75,7 +75,7 @@ class KMeans(TransformerMixin, BaseEstimator):
         by `np.random`.
 
     n_init : 'auto' or int, default=10
-        Number of time the k-means algorithm will be run with different
+        Number of times the k-means algorithm will be run with different
         centroid seeds. The final results will be the best output of
         n_init consecutive runs in terms of inertia.
         When `n_init='auto'`, the number of runs will be 10 if using
--- a/dask_ml/cluster/k_means.py
+++ b/dask_ml/cluster/k_means.py
@@ -74,6 +74,17 @@ class KMeans(TransformerMixin, BaseEstimator):
         If None, the random number generator is the RandomState instance used
         by `np.random`.
 
+    n_init : 'auto' or int, default=10
+        Number of time the k-means algorithm will be run with different
+        centroid seeds. The final results will be the best output of
+        n_init consecutive runs in terms of inertia.
+        When `n_init='auto'`, the number of runs will be 10 if using
+        `init='random'`, and 1 if using `init='kmeans++'`.
+        .. versionadded:: 1.2
+           Added 'auto' option for `n_init`.
+        .. versionchanged:: 1.4
+           Default value for `n_init` will change from 10 to `'auto'` in version 1.4.
+
     Attributes
     ----------
     cluster_centers_ : np.ndarray [n_clusters, n_features]
@@ -142,6 +153,7 @@ class KMeans(TransformerMixin, BaseEstimator):
         n_jobs=1,
         algorithm="full",
         init_max_iter=None,
+        n_init="auto",
     ):
         self.n_clusters = n_clusters
         self.init = init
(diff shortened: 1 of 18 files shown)


mmccarty:
Ready for review @jrbourbeau cc @betatim

dask_ml/cluster/k_means.py
```diff
@@ -74,6 +74,17 @@ class KMeans(TransformerMixin, BaseEstimator):
         If None, the random number generator is the RandomState instance used
         by `np.random`.
 
+    n_init : 'auto' or int, default=10
+        Number of time the k-means algorithm will be run with different
```
betatim, line 78:
```suggestion
        Number of times the k-means algorithm will be run with different
```

dask_ml/cluster/k_means.py
```diff
@@ -315,6 +331,16 @@ def k_init(
     oversampling_factor : int, optional
         Only used for ``init='k-means||`''. Controls the additional number of
         candidate centers in each iteration.
+    n_init : 'auto' or int, default=10
+        Number of time the k-means algorithm will be run with different
```
betatim, line 335:
```suggestion
        Number of times the k-means algorithm will be run with different
```

betatim:
The changes looks reasonable to me. Only question I have is regarding backwards compatibility and such. Does dask-ml make promises about changes of parameter default values?

betatim (approved):

dask_ml/decomposition/pca.py
```diff
@@ -327,11 +327,11 @@ def _fit(self, X):
         # Compute noise covariance using Probabilistic PCA model
         # The sigma2 maximum likelihood (cf. eq. 12.46)
         self.noise_variance_ = 0.0
-        if n_components < min(self.n_features_, self.n_samples_):
+        if n_components < min(self.n_features_in_, self.n_samples_):
             if solver == "randomized":
                 self.noise_variance_ = (
                     total_variance - self.explained_variance_.sum()
-                ) / (min(self.n_features_, self.n_samples_) - n_components)
+                ) / (min(self.n_features_in_, self.n_samples_) - n_components)
```
stsievert, line 334: Should this renaming of `n_features_` to `n_features_in_` trigger a deprecation warning?

dask_ml/preprocessing/_encoders.py
```diff
@@ -116,22 +116,29 @@ def __init__(
         categorical_features: Optional[pd.Categorical] = None,
         categories: Union[str, ArrayLike] = "auto",
         drop: Optional[bool] = None,
-        sparse: bool = True,
+        sparse_output: bool = True,
```
stsievert, line 119: Should  this renaming of `sparse` to `sparse_output` trigger a deprecation warning?

mmccarty:
> The changes looks reasonable to me. Only question I have is regarding backwards compatibility and such. Does dask-ml make promises about changes of parameter default values?

No promises that I'm aware of. From my experience, the project seems to follow scikit-learn for API changes. I don't intend on adding backwards compatibility with older version of scikit-learn in this PR. Others are welcome to follow up.

dask_ml/decomposition/pca.py
mmccarty: scikit-learn is handling that. In this case, `sklearn.preprocessing.OneHotEncoder` will raise a warning.

dask_ml/preprocessing/_encoders.py
mmccarty: `sklearn.preprocessing.OneHotEncoder` will raise a warning.

TomAugspurger:
In general, my policy was to match the lowest-supported version of
scikit-learn, but to pretty aggressively bump that minimum supported
version when it became at all challenging to support both versions. If
pople are able to upgrade their version of scikit-learn to the latest, then
they ought to be able to upgrade dask-ml as well (and vice-versa).

mmccarty:
I'm not opposed to this. #957 requires 1.1.3 and is merged. Added #961 to track.

betatim (approved):

mmccarty:
@jrbourbeau - Any comments here?

jrbourbeau (commented):
Apologies for the delayed response. I've not had a change to look at the changes here yet and probably won't be able to get to it for a bit. If you and @betatim are happy with the changes here, I'd recommend moving forward with them. No need to block on me

betatim:
Let's merge this then. I would have clicked the button, but I can't :D

Issue #964, opened 2023-02-08
sklearn handles text labels differently than ml_dask on OneHotEncoding


brian-methodical:
**Describe the issue**:
[OneHotEncoding](https://github.com/dask/dask-ml/blob/ba2024d68557fade43f4dc2fa369445d01629923/dask_ml/preprocessing/_encoders.py#L16) defaults to sets `catagories_` to a numpy dtype='<U1' when explicitly given an array of text [["a", "b", "c"]] sklearn sets `catagories_` dtype `object`.
Which is the correct behavior:

1. either cast `catagories_` to dtype=object from OneHotEncoding on dask-ml when explicitly setting arrays (auto[1] is off)
2. or fix the test so it does not care about `catagories_` dtype mismatch OR explicitly set dtype as argument in tests

[1] please note this issue is only with categories explicitly set and not when set to 'auto'.

When set to 'auto' dtype is numpy.float64 array, which is a third behavior. However we are matching what is done by sklearn there.

**Minimal Complete Verifiable Example**:

[see test_basic_array ](https://github.com/dask/dask-ml/blob/7b8c8e0f2d3fb029850e9bf07203b8f7176bdf3e/tests/preprocessing/test_encoders.py#L24)

**Anything else we need to know?**:

This is illustrated in the [failing test](https://github.com/dask/dask-ml/actions/runs/4083895663/jobs/7039952888#step:5:1050)

Sklearn documentation found [here](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html) shows dtype as object where dask-ml shows dtype as `<U1` found [here](https://github.com/dask/dask-ml/blob/80eb0d81b22df3762706384c3ca1c0be4137e591/dask_ml/preprocessing/_encoders.py#L104)

**Environment**:

- Dask version:  dask-ml-3.8 conda env  dask 2023.1.1
- Python version: 3.8, 3.9, 3.10
- Operating System:  ubuntu-latest
- Install method (conda, pip, source): conda

narnia24:
hey can i work on this?

ruthvik0612:
Hi 👋  
I’m a new contributor and a recent graduate working with ML preprocessing tools.  
I’m interested in working on this issue to understand and align (or document) the behavior of OneHotEncoder with text labels between dask-ml and scikit-learn.  

Before starting, I wanted to ask:
- Is this issue still valid?
- Would you prefer a behavioral fix, additional tests, or documentation clarification?

[...]

Issue #963, opened 2023-02-08
HashingVectorizer behaves differently from FeatureHasher


brian-methodical:
**Describe the issue**:
[HashingVectorizer](https://ml.dask.org/modules/generated/dask_ml.feature_extraction.text.HashingVectorizer.html#dask-ml-feature-extraction-text-hashingvectorizer) behaves differently from FeatureHasher,  HashingVectorizer can work off a Sting like

```
JUNK_FOOD_DOCS = (
    "the pizza pizza beer copyright",
    "the pizza burger beer copyright",
    "the the pizza beer beer copyright",
    "the burger beer beer copyright",
    "the coke burger coke copyright",
    "the coke burger burger",
)

```

but [FeatureHasher](https://ml.dask.org/modules/generated/dask_ml.feature_extraction.text.FeatureHasher.html#dask-ml-feature-extraction-text-featurehasher) expects an iterable of strings like:

```
JUNK_FOOD_DOCS = [["the", "pizza", "pizza", "beer", "copyright"],
   ["the", "coke", "burger", "burger"] ]

```
Which is the correct behavior:

1. expect the hasher to parse strings into vectors
2. fix the test by sending list of lists of strings to FeatureHasher instead list of strings, like the other hasher expects

**Minimal Complete Verifiable Example**:

[See dask-ml/tests/feature_extraction/test_text.py: test_basic() ](https://github.com/dask/dask-ml/blob/ba2024d68557fade43f4dc2fa369445d01629923/tests/feature_extraction/test_text.py#L33)

**Anything else we need to know?**:

This is illustrated in the [failing test](https://github.com/dask/dask-ml/actions/runs/4083895663/jobs/7039952888#step:5:817)

**Environment**:

- Dask version:  dask-ml-3.8 conda env  dask 2023.1.1
- Python version: 3.8, 3.9, 3.10
- Operating System:  ubuntu-latest
- Install method (conda, pip, source): conda

brian-methodical:
Here is a possible fix to this test if this is the route we wish to go: https://github.com/dask/dask-ml/compare/main...brian-methodical:dask-ml:fix-tests#diff-824171fe718be1c9bd2d722b5ebc30f71b6dd402568282716078a1c5ec25db1f

PR #965, opened 2023-02-10
Fix tests

brian-methodical:
resolving:

sklearn handles text labels differently than ml_dask on OneHotEncoding #964

HashingVectorizer behaves differently from FeatureHasher #963

@mmccarty who can review?


mmccarty:
Thanks @brian-methodical !

PR #994, opened 2024-06-08
modified cross_validation.rst file with a k-fold cross validation example

npk7:
Dear Dask team,

Currently, cross validation page consists only of train_test_split() method. Here is the link: https://ml.dask.org/cross_validation.html

Dask ML comes with powerful kFold method, so created a simple example to illustrate the kFold cross validation method. 

My hope is this would contribute positively to the documentation. Please advice. Thank you for the opportunity of contributing to this amazing community driven ML library.

Here is the link I used for the same: https://ml.dask.org/modules/generated/dask_ml.model_selection.KFold.html

### DIFF
--- a/docs/source/cross_validation.rst
+++ b/docs/source/cross_validation.rst
@@ -4,7 +4,28 @@ Cross Validation
 See the `scikit-learn cross validation documentation`_ for a fuller discussion of cross validation.
 This document only describes the extensions made to support Dask arrays.
 
-The simpler way to split a dataset into k-fold is with :func:`https://ml.dask.org/modules/generated/dask_ml.model_selection.KFold.html`
+The simplest way to split one or more Dask arrays is with :func:`dask_ml.model_selection.train_test_split`:
+
+.. ipython:: python
+
+   import dask.array as da
+   from dask_ml.datasets import make_regression
+   from dask_ml.model_selection import train_test_split
+
+   X, y = make_regression(n_samples=125, n_features=4, random_state=0, chunks=50)
+   X
+
+The interface for splitting Dask arrays is the same as scikit-learn's version.
+
+.. ipython:: python
+
+   X_train, X_test, y_train, y_test = train_test_split(X, y)
+   X_train  # A dask Array
+
+   X_train.compute()[:3]
+
+Here is another illustration of performing k-fold cross validation purely in Dask. Here a link to gather more information on k-fold cross validation :func:`https://ml.dask.org/modules/generated/dask_ml.model_selection.KFold.html`:
+
 .. ipython:: python
 
    import dask.array as da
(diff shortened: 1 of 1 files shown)


docs/source/cross_validation.rst
```diff
@@ -24,6 +24,47 @@ The interface for splitting Dask arrays is the same as scikit-learn's version.
 
    X_train.compute()[:3]
 
+Here is another illustration of performing k-fold cross validation purely in Dask. Here a link to gather more information on k-fold cross validation :func:`https://ml.dask.org/modules/generated/dask_ml.model_selection.KFold.html`:
```
TomAugspurger, line 27:
```suggestion
Here is another illustration of performing k-fold cross validation purely in Dask. Here a link to gather more information on k-fold cross validation :class:`dask_ml.model_selection.KFold`:
```

I think the sphinx reference should be to function / class, and it'll translate to the HTML link when built.

docs/source/cross_validation.rst
```diff
@@ -24,6 +24,47 @@ The interface for splitting Dask arrays is the same as scikit-learn's version.
...
+				 chunks=20) # partitions to be made 
+
+   train_scores: list[int] = []
+   test_scores: list[int] = []
+
+   model = LinearRegression()
+
+The Dask kFold method splits the data into k consecutive subsets of data. Here we specify k to be 5, hence, 5-fold cross validation
+
+.. ipython:: python
```
TomAugspurger, line 49:
```suggestion

.. ipython:: python
```

I believe you need a newline below the directive.

docs/source/cross_validation.rst
```diff
@@ -24,6 +24,47 @@ The interface for splitting Dask arrays is the same as scikit-learn's version.
...
+   kf = KFold(n_splits=5)
+
+   for i, j in kf.split(X):
+      X_train, X_test = X[i], X[j]
```
TomAugspurger, line 53: Can you check that the formatting here is correct? (might need another space)

docs/source/cross_validation.rst
TomAugspurger, line 27: All of this should go at the bottom of the document (the next two paragraphs are discussing train_test_split still)

docs/source/cross_validation.rst
npk7: you are correct, it needed another space. updated the code to reflect the same.

docs/source/cross_validation.rst
npk7: incorporated your suggestions, moved k-fold to the bottom of the document

docs/source/cross_validation.rst
```diff
@@ -36,4 +35,48 @@ However, if there's a strong pattern in your data, you'll want to perform a full
...
+
+   model = LinearRegression()
+
+The Dask kFold method splits the data into k consecutive subsets of data. Here we specify k to be 5, hence, 5-fold cross validation
+
+
+.. ipython:: python
+   kf = KFold(n_splits=5)
+
+   for i, j in kf.split(X):
```
TomAugspurger, line 64: Looks like the doc build failed here: https://github.com/dask/dask-ml/actions/runs/9434455032/job/26270343032?pr=994

Is that from the lack of a newline after the `ipython` directive?

npk7: thanks you for spotting that, added a new line

PR #997, opened 2024-07-20
CI Fixups

TomAugspurger:
- numpy 2 combat
- changed error messgae

PR #998, opened 2024-07-21
Fixing incremental_pca

fujiisoup:
Trial to fixing #997

Let me work in this PR. The test was failing in the CI but passing in my computer.


fujiisoup:
Closing as it is an upstream issue.

Issue #1003, opened 2024-11-25
Tests fail with dask 2024.11.2


GaetanLepage:
**Describe the issue**:

Some tests are failing since [we bumped dask to 2024.11.2](https://github.com/NixOS/nixpkgs/pull/354636):

<details>

```
=========================== short test summary info ============================
FAILED tests/model_selection/test_hyperband.py::test_basic[numpy-ConstantFunction-20] - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-74...
FAILED tests/model_selection/test_hyperband.py::test_hyperband_patience - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-78...
FAILED tests/model_selection/test_hyperband.py::test_same_random_state_same_params - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-dd...
FAILED tests/model_selection/test_incremental.py::test_explicit - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-ba...
FAILED tests/model_selection/test_incremental.py::test_search_basic - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-b2...
FAILED tests/model_selection/test_incremental.py::test_search_plateau_patience - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-e9...
FAILED tests/model_selection/test_incremental.py::test_search_plateau_tol - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-55...
FAILED tests/model_selection/test_incremental.py::test_gridsearch - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-30...
FAILED tests/model_selection/test_incremental.py::test_numpy_array - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-12...
FAILED tests/model_selection/test_incremental.py::test_transform - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-4a...
FAILED tests/model_selection/test_incremental.py::test_small - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-16...
FAILED tests/model_selection/test_incremental.py::test_smaller - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-11...
FAILED tests/model_selection/test_incremental.py::test_high_performing_models_are_retained_with_patience - RuntimeError: Not enough arguments provided: missing keys {'_partial_fit-1d...
[...]

PR #1004, opened 2024-11-26
Avoid persisting Futures

TomAugspurger:
Closes https://github.com/dask/dask-ml/issues/1003

With this, `test_incremental.py` passes for me locally.


TomAugspurger:
This was included in #1008

Issue #1012, opened 2025-02-16
Tests failing with `ValueError: cannot broadcast shape (nan,) to shape (nan,)`


GaetanLepage:
**Describe the issue**:

When updating `dask` to 2025.2.0, the following `dask-ml` tests start failing:
```
=========================== short test summary info ============================
FAILED tests/ensemble/test_blockwise.py::TestBlockwiseVotingRegressor::test_fit_array - ValueError: cannot broadcast shape (nan,) to shape (nan,)
FAILED tests/ensemble/test_blockwise.py::TestBlockwiseVotingRegressor::test_fit_frame - ValueError: cannot broadcast shape (nan,) to shape (nan,)
FAILED tests/linear_model/test_glm.py::test_lr_score - ValueError: cannot broadcast shape (nan,) to shape (nan,)
FAILED tests/metrics/test_regression.py::test_ok[r2_score-True] - ValueError: cannot broadcast shape (nan,) to shape (nan,)
FAILED tests/metrics/test_regression.py::test_ok[r2_score-False] - ValueError: cannot broadcast shape (nan,) to shape (nan,)
FAILED tests/preprocessing/test_data.py::TestQuantileTransformer::test_fit_transform_frame - IndexError: boolean index shape (nan,) must match indexed array's (nan,).
FAILED tests/test_incremental.py::test_scoring_string[threads-r2] - ValueError: cannot broadcast shape (nan,) to shape (nan,)
FAILED tests/test_incremental.py::test_scoring_string[distributed-r2] - ValueError: cannot broadcast shape (nan,) to shape (nan,)
FAILED tests/test_parallel_post_fit.py::test_laziness - ValueError: cannot broadcast shape (nan,) to shape (nan,)
= 9 failed, 585 passed, 116 skipped, 1 deselected, 8 xfailed, 6 xpassed, 312 warnings in 204.79s (0:03:24) =
```

**Minimal Complete Verifiable Example**:

```console
$ pytest
```

**Anything else we need to know?**:

Context: bumping dask in nixpkgs: https://github.com/NixOS/nixpkgs/pull/381896

**Environment**:

- Dask version: 2025.2.0
- Python version: 3.12.8
- Operating System: NixOS
- Install method (conda, pip, source): nix

matteociccozzi-ah:
@TomAugspurger 
```
if __name__ == "__main__":
    from dask.distributed import Client, LocalCluster

    cluster = LocalCluster(n_workers=4)
[...]

PR #1013, opened 2025-03-03
fix: prevent broadcasting errors in r2_score using da.where()

wietzesuijker:
Closes https://github.com/dask/dask-ml/issues/1012

First PR here. Curious to hear your feedback.

**Problem**
After updating to Dask 2025.2.0, tests fail with a ValueError due to changes in chunk size handling.

**Solution**
Refactor r2_score() to use da.where() for correct broadcasting.

**Testing**
Test added to ensure r2_score() works correctly with arrays that have different chunk configurations.

### DIFF
--- a/dask_ml/metrics/regression.py
+++ b/dask_ml/metrics/regression.py
@@ -162,18 +162,13 @@ def r2_score(
     numerator = (weight * (y_true - y_pred) ** 2).sum(axis=0, dtype="f8")
     denominator = (weight * (y_true - y_true.mean(axis=0)) ** 2).sum(axis=0, dtype="f8")
 
-    nonzero_denominator = denominator != 0
-    nonzero_numerator = numerator != 0
-    valid_score = nonzero_denominator & nonzero_numerator
-    output_chunks = getattr(y_true, "chunks", [None, None])[1]
-    output_scores = da.ones([y_true.shape[1]], chunks=output_chunks)
-    with np.errstate(all="ignore"):
-        output_scores[valid_score] = 1 - (
-            numerator[valid_score] / denominator[valid_score]
-        )
-        output_scores[nonzero_numerator & ~nonzero_denominator] = 0.0
+    score = da.where(
+        numerator == 0,
+        1.0,
+        da.where(denominator != 0, 1 - numerator / denominator, 0.0),
+    )
 
-    result = output_scores.mean(axis=0)
+    result = score.mean(axis=0)
     if compute:
         result = result.compute()
     return result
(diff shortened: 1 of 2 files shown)


dask_ml/ensemble/_blockwise.py
```diff
@@ -54,6 +54,7 @@ def _predict(self, X):
             dtype = "float64"
 
         if isinstance(X, da.Array):
+            X = X.compute_chunk_sizes()  # Ensure integer chunks to avoid broadcasting errors.
```
TomAugspurger, line 57: I'd pretty strongly like to avoid this if it's possible. Computing the chunk sizes requires executing the graph up to this point, which could be very expensive.

wietzesuijker: Thanks for your feedback @TomAugspurger. Sorry for the delay. I've revised the changes. Could you review them again?

TomAugspurger:
Thanks.

I'm not entirely sure what the best action is, but I think we ought to avoid anything that triggers computation unnecessarily, including `len`.

Can you say a bit more about getting `n_samples` is needed in blockwise?

wietzesuijker:
Thanks @TomAugspurger. `n_samples` (now obtained via X.shape[0]) lets us determine the rechunking size without forcing computation. It splits the test data into one block per trained estimator, ensuring alignment and preventing broadcast errors. Combined with the da.where() update in r2_score, these changes maintain laziness and correct behavior with mismatched chunks.

TomAugspurger:
I'm probably missing something, but why do we care that the size of the test dataset matches the size of the training dataset (`_n_samples`)? I'd expect us to just care that the number of samples in `X_train` and `y_train` to match, and separately that the number of samples in `X_test`, `y_test` match.

wietzesuijker:
> why do we care that the size of the test dataset matches the size of the training dataset (`_n_samples`)?

The goal is not for the test dataset to match the training dataset's overall size. The focus is ensuring each estimator, trained on a specific data block, receives a matching block from the test set. `X.shape[0]` is used (as `n_samples`) to compute the optimal test data block size, dividing the test set into blocks equal to the number of estimators. This aligns predictions and prevents broadcast errors, regardless of training and test dataset sizes.

wietzesuijker:
Thanks for triggering the tests, Tom. Is there anything I can/should do to fix the failing runs? The errors seem unrelated and similar to other recent runs e.g. https://github.com/dask/dask-ml/actions/runs/14849752234/job/41691028477.

TomAugspurger:
I'll take a look in the next few days.

TomAugspurger:
Thanks @wietzesuijker. There's still one error in `tests/preprocessing/test_data.py::TestQuantileTransformer::test_fit_transform_frame` but the rest from https://github.com/dask/dask-ml/issues/1012 are fixed. I'll fix that in another PR.

PR #1014, opened 2025-03-11
add TF-IDF Transformer

pr38:
Hi,

I am trying to add TfidfTransformer to dask. I am going to assume that adding TfidfVectorizer can be easily done later down the line, by stacking CountVectorizer and TfidfTransformer .

 This is not the first attempt to add tfidf to dask-ml(https://github.com/dask/dask-ml/pull/869/files#diff-421c76129d7900a3ae83237eab5785858c0652d9f550d92e0202e45ebdcb2977). I can't comment on the differences between my attempt and previous attempts; but I can guaranty that my implementation only use's the dask array api, not no other api.

### DIFF
--- a/dask_ml/feature_extraction/text.py
+++ b/dask_ml/feature_extraction/text.py
@@ -3,6 +3,8 @@ Utilities to build feature vectors from text documents.
 """
 
 import itertools
+import math
+from typing import Literal, Optional
 
 import dask
 import dask.array as da
@@ -13,9 +15,13 @@ import numpy as np
 import scipy.sparse
 import sklearn.base
 import sklearn.feature_extraction.text
+import sparse
 from dask.delayed import Delayed
 from distributed import get_client, wait
+from sklearn.base import BaseEstimator, OneToOneFeatureMixin, TransformerMixin
 from sklearn.utils.validation import check_is_fitted
+
+from ..utils import check_array
 
 
 class _BaseHasher(sklearn.base.BaseEstimator):
@@ -280,3 +286,93 @@ def _merge_vocabulary(*vocabularies):
         )
     }
     return vocabulary
+
+
+def _handle_zeros_in_scale(scale: da.Array) -> da.Array:
+
+    constant_mask = scale < 10 * np.finfo(scale.dtype).eps
+
+    scale[constant_mask] = 1.0
+
+    return scale
+
+
+def _normalize(X: da.Array, norm: Literal["l2", "l1", "max"] = "l2") -> da.Array:
+    if norm not in ("l1", "l2", "max"):
+        raise ValueError("'%s' is not a supported norm" % norm)
+
+    if norm == "l1":
+        norms = da.abs(X).sum(axis=1)
+    elif norm == "l2":
+        norms = da.sqrt(da.square(X).sum(axis=1))
+    elif norm == "max":
+        norms = da.max(da.abs(X), axis=1)
+
+    norms = _handle_zeros_in_scale(norms)
+    X = X / norms[:, np.newaxis]
+
+    return X
+
+
+class TfidfTransformer(
+    OneToOneFeatureMixin, TransformerMixin, BaseEstimator, auto_wrap_output_keys=None
+):
+
+    def __init__(
+        self,
+        *,
+        norm: Optional[Literal["l2", "l1"]] = "l2",
+        use_idf: bool = True,
+        smooth_idf: bool = True,
+        sublinear_tf: bool = False,
+    ):
+        self.norm = norm
+        self.use_idf = use_idf
+        self.smooth_idf = smooth_idf
+        self.sublinear_tf = sublinear_tf
+
+    def fit(self, X, y=None):
+        X = check_array(X)
+
+        if self.use_idf:
+
+            X = X.map_blocks(lambda a: sparse.as_coo(a).astype(np.float64))
+
+            if math.isnan(
+                X.shape[0]
+            ):  # if number is rows is not currently know, get it
+                X.compute_chunk_sizes()
+
+            n_samples = X.shape[0]
+
+            df = da.count_nonzero(X, axis=0)
+
+            if self.smooth_idf:
+                n_samples = n_samples + 1
+                df = df + 1
+
+            idf_ = da.log(n_samples / df) + 1
+
+            self.idf_ = idf_.compute()
+
+        return self
+
+    def transform(self, X):
+
+        if self.use_idf:
+            check_is_fitted(self, "idf_")
+
+        X = X.map_blocks(lambda a: sparse.as_coo(a).astype(np.float64))
+
+        if self.sublinear_tf:
+            X = da.where(X != 0, da.log(X) + 1, X)
+
+        if self.use_idf:
+            tf_idf = X * self.idf_
+        else:
+            tf_idf = X
+
+        if self.norm:
+            tf_idf = _normalize(tf_idf, norm=self.norm)
+
+        return tf_idf
--- a/tests/feature_extraction/test_text.py
+++ b/tests/feature_extraction/test_text.py
@@ -188,3 +188,19 @@ def test_count_vectorizer_remote_vocabulary():
         )
         m.fit_transform(b)
         assert m.vocabulary_ is remote_vocabulary
+
+@pytest.mark.parametrize("norm", [None, "l2","l1"])
+@pytest.mark.parametrize("smooth_idf", [True, False])
+@pytest.mark.parametrize("sublinear_tf", [True, False])
+@pytest.mark.parametrize("use_idf", [True, False])
+def test_tf_idf(data,norm,smooth_idf,sublinear_tf,use_idf):
+    x = sklearn.feature_extraction.text.CountVectorizer().fit(JUNK_FOOD_DOCS)
+
+    d_tf_idf =  dask_ml.feature_extraction.text.TfidfTransformer(norm=norm,smooth_idf=smooth_idf,sublinear_tf=sublinear_tf,use_idf=use_idf)
+    d_tf_idf_result =  d_tf_idf.fit_transform(x).compute()
+
+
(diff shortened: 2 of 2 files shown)


dask_ml/feature_extraction/text.py
```diff
@@ -280,3 +286,90 @@ def _merge_vocabulary(*vocabularies):
...
+            ):  # if number is rows is not currently know, get it
+                X.compute_chunk_sizes()
+
+            n_samples = X.shape[0]
+
+            df = da.count_nonzero(X, axis=0)
+
+            if self.smooth_idf:
+                n_samples = n_samples + 1
+                df = df + 1
```
stsievert, line 350:
```suggestion
                df += 1
```

dask_ml/feature_extraction/text.py
```diff
@@ -280,3 +286,90 @@ def _merge_vocabulary(*vocabularies):
...
+                X.shape[0]
+            ):  # if number is rows is not currently know, get it
+                X.compute_chunk_sizes()
+
+            n_samples = X.shape[0]
+
+            df = da.count_nonzero(X, axis=0)
+
+            if self.smooth_idf:
+                n_samples = n_samples + 1
```
stsievert, line 349:
```suggestion
                n_samples += 1
```

stsievert (commented):
Thanks for the contribution! At first glance, this looks pretty good. I'd appreciate some more review.

All these comments and linting errors are minor. If you want to run the linting yourself, look at https://ml.dask.org/contributing.html#style

pr38:
ok, I am ready for more feedback.

I should mention that I am running `compute_chunk_sizes ` when the number of rows are not know. I have seen maintainers complain about other's doing that here. If you guys don't want me to run `compute_chunk_sizes `, maybe we can make sure that CountVectorizer keeps chunk_sizes, to avoid foot guns.

dask_ml/feature_extraction/text.py
```diff
@@ -280,3 +284,87 @@ def _merge_vocabulary(*vocabularies):
...
+
+    def fit(self, X, y=None):
+        if self.use_idf:
+
+            X = X.map_blocks(lambda a: sparse.as_coo(a).astype(np.float64))
+
+            if math.isnan(
+                X.shape[0]
+            ):  # if number is rows is not currently know, get it
+                X.compute_chunk_sizes()
```
stsievert, line 340: This shouldn't be called implicitly because it immediately triggers computation. I think we should continue to raise, or at least warn. There's a warning you can use at https://github.com/dask/dask/blob/0b526507ab7254f435039536d058c7c5b958ec4b/dask/array/core.py#L105-L112

pr38: ok, I think I found a 'best of all worlds' solution here. I am now using lazy row counts when chunk sizes as not know. The code is largely dug up from dask core's 'compute_chunk_sizes'. I am of the opinion that you guys should add lazy chunk sizes to dask array core.

dask_ml/feature_extraction/text.py
```diff
@@ -280,3 +284,103 @@ def _merge_vocabulary(*vocabularies):
...
+def _get_chunk_len(a):
+    s = np.asarray(a.shape[0], dtype=int)
+    return s.reshape(1, 1, 1)
+
+
+def _get_lazy_row_count_as_dask_array(a: da.Array) -> da.Array:
+    chunk_len = a.map_blocks(
+        _get_chunk_len,
+        dtype=int,
+        chunks=tuple(len(c) * (1,) for c in a.chunks) + ((a.ndim,),),
```
stsievert, line 324: Why do chunks have `(a.ndim,)` tacked on? Why does `_get_chunk_len` return a 3-D ndarray?

Those questions might be moot, I was able to make `_get_chunk_len` return a 2-D array:

``` python
def _get_lazy_row_count_as_dask_array(a: da.Array) -> da.Array:
    chunk_len = a.map_blocks(
        lambda a: np.asarray(a.shape[0], dtype=int).reshape(1, 1),
        dtype=int,
        chunks=tuple(len(c) * (1,) for c in a.chunks),
    )
    return chunk_len.sum()
```

This works with 2-D arrays. Not sure how it will work with >2-D arrays, also not sure that matters in the text use case.

dask_ml/feature_extraction/text.py
```diff
@@ -280,3 +284,103 @@ def _merge_vocabulary(*vocabularies):
...
+        if self.use_idf:
+
+            X = X.map_blocks(lambda a: sparse.as_coo(a).astype(np.float64))
+
+            if math.isnan(
+                X.shape[0]
+            ):  # if number is rows is not currently know, get a lazy reference to row counts
+                n_samples = _get_lazy_row_count_as_dask_array(X)
+            else:
+                n_samples = X.shape[0]
```
stsievert, line 358: Style nit:

``` python
            # if number is rows is not currently know, get a lazy reference to row counts
            n_samples = X.shape[0] if not math.isnan(X.shape[0]) else _lazy_n_samples(X)
```

dask_ml/feature_extraction/text.py
```diff
@@ -280,3 +284,103 @@ def _merge_vocabulary(*vocabularies):
...
+        smooth_idf: bool = True,
+        sublinear_tf: bool = False,
+    ):
+        self.norm = norm
+        self.use_idf = use_idf
+        self.smooth_idf = smooth_idf
+        self.sublinear_tf = sublinear_tf
+
+    def fit(self, X, y=None):
+        if self.use_idf:
```
stsievert, line 349: Style nit: some indentation can be removed with a return statement at the top:

``` python
class TfidfTransformer:
    def fit(self, X, y=None):
        if not self.use_idf:
            return self
        ...

File: dask_ml/feature_extraction/text.py (hunk 4 above)
Comment (pr38) [reply dask_ml/feature_extraction/text.py:324]: Well done! map_blocks's shapes had been one of the harder parts of dask.

Comment (pr38) [reply dask_ml/feature_extraction/text.py]: Assuming lazy_row_counts is going to only be used by tf-idf, we should be fine.

**Comment by pr38** [comment]
There are a couple of extra things I want to point out. While scikit-learn returns a scipy.sparse.coo_array while I return a sparse.COO. I don't know how well the rest of dask-ml plays with sparse.COO, and I am not sure how committed to 1 to 1 matching scikit-learn dask-ml is.

I have also seen two other implementations of '_handle_zeros_in_scale' in the dask-ml code base. I would have used the existing implementations but none used epsilon to catch near zero values.

File: dask_ml/feature_extraction/text.py [hunk 7]
```diff
@@ -280,3 +284,92 @@ def _merge_vocabulary(*vocabularies):
...
+    return X
+
+
+def _get_lazy_row_count_as_dask_array(a: da.Array) -> da.Array:
+    chunk_len = a.map_blocks(
+        lambda a: np.asarray(a.shape[0], dtype=int).reshape(1, 1),
+        dtype=int,
+        chunks=tuple(len(c) * (1,) for c in a.chunks),
+    )
+    return chunk_len.sum()
```
stsievert, line 321: Dask Arrays are split into blocks. A 1000x30 array, might be split it into four 500x15 blocks. So the `map_blocks` applies a given function to each of those four blocks. If that function is `len`, the length of each block would be computed and result in an array `[[500, 500], [500, 500]]`. The `chunk_len.sum()` line would sum those 4 values, and overestimate the length of the array to be 2000 instead of 1000. This suggestion fixes that:

```suggestion
    return chunk_len[:, 0].sum()  # in case each row split into >=2 chunks
```

This is an issue, the input is user supplied. I think the test should be modified:

``` python
def test_tf_idf(norm, smooth_idf, sublinear_tf, use_idf, rechunk):
    corpus = db.from_sequence(JUNK_FOOD_DOCS)
    x_d = dask_ml.feature_extraction.text.CountVectorizer().fit_transform(corpus)
    if rechunk:
        x_d.compute_chunk_sizes()
        x_d = x_d.rechunk(2, 3)
    ...
```

Below is some code that illustrates the issue, perhaps more clearly. All versions of the `test_tf_idf` test pass with this code:

``` python
def _get_lazy_row_count_as_dask_array(a: da.Array) -> da.Array:
    a.compute_chunk_sizes()
    a = a.rechunk(chunks=(2, 3))
    assert a.shape[0] == 6
    chunk_len = a.map_blocks(
        lambda a: np.asarray(a.shape[0], dtype=int).reshape(1, 1),
        dtype=int,
        chunks=tuple(len(c) * (1,) for c in a.chunks),
    )
    assert chunk_len.compute().tolist() == [[2, 2], [2, 2], [2, 2]]
    assert chunk_len.compute().sum() == 12 != 6  # not 6 as expected!
    assert chunk_len[:, 0].sum().compute() == 6
    # return chunk_len.sum()  # all tests fail when I uncomment this line
    return chunk_len[:, 0].sum()
```

stsievert (commented):
You've also got some style errors. Could you look at run the linting commands at https://ml.dask.org/contributing.html#style ?

dask_ml/feature_extraction/text.py
pr38: ok, I understand; I had assumed chunks only existed in the first axis(like in analytical/dataframe use cases), not just any axis. Updated tests and code to capture this scenario.

pr38:
@stsievert what else would you need for this pr to be merged/approved?

pr38:
@stsievert are you willing to take a second look at this, hoping to push this push this through. If you are busy, would you happen to know of another maintiner we can bring in?

stsievert:
This looks good. A couple suggestions:

- Could you add some tests to make sure Dask-ML's TfidfTransformer is the same as sklearn's TfidfTransformer, that it has the same attributes/etc? I think `dask_ml.utils.assert_estimator_equal` would help with this (and simplify your test)
- Linting is still failing. Could you correct that?

pr38:
@stsievert

I have added the 'dask_ml.utils.assert_estimator_equal' to the tests. 

Regarding linting, it seems that flake8 and black are in conflict. Would you like me to prioritize one over the other?

Issue #1016, opened 2025-03-25
RuntimeError: Attempting to use an asynchronous Client in a synchronous context of `dask.compute`


GaetanLepage:
**Describe the issue**:

When bumping `dask` to `2025.3.0`, almost all tests in those files
```
tests/model_selection/test_incremental.py
tests/model_selection/test_successive_halving.py
tests/model_selection/test_hyperband.py
```
sarted failing with:
```
RuntimeError: Attempting to use an asynchronous Client in a synchronous context of `dask.compute`
```

**Minimal Complete Verifiable Example**:

```console
$ pytest
```

**Anything else we need to know?**:

**Environment**:

- Dask version: 2025.3.0
- Python version: 3.12
- Operating System: NixOS
- Install method (conda, pip, source): nix

TomAugspurger:
@stsievert any chance you could take a look at this?

The immediate error is from us using `dask.persist(...)` instead of `client.persist(...)` in hyperband. That's relatively easy to fix.

However, there's been some changes in distributed to disallow persisting collections of Futures, so `client.persist({0: Future})` doesn't work anymore. I haven't been able to spot an easy fix yet :/

PR #1018, opened 2025-05-10
CI Fixes

TomAugspurger:
This gets CI passing again.

- 0b11c327e7ba8b0a1b2a343c62683d5c6c26d7c3 skips that test on older versions of dask.https://github.com/dask/dask/pull/11943 should fix it upstream.
- ca71efb9e9a526932b79776798e92495ee5e390f skips all the hyperband tests. I couldn't figure it out. More at https://github.com/dask/dask-ml/issues/1016#issuecomment-2868934974.


TomAugspurger:
There's some more tech debt in the sphinx setup. Currently pinned to sphinx<5, which isn't working on python>=3.13.

PR #1022, opened 2026-04-08
Addition of flexible chunking support to dask_ml.datasets

### LINKED ISSUE #65: Make chunks optional in the datasets.make_* methods

Jarvis2001:
### Summary
Addition of flexible chunking support to `dask_ml.datasets` generators by introducing three chunking modes:
- explicit `chunks=` input (old behaviour)
- `chunks="auto"` / `chunks=None` for Dask automatic chunking
- `chunks="cores"` for CPU-core-based row chunk heuristics

### Changes
- Added `_normalize_chunks()` helper in datasets.py
- Updated `make_counts`, `make_blobs`, `make_regression`, and `make_classification` to use the helper
- Updated `make_classification_df` so generated date columns follow the same row partitioning as `X_array`
- Extended generator docstrings to document the new `chunks="auto"` and `chunks="cores"` options
- Added tests in test_datasets.py covering `make_counts` with `None`, `"auto"`, and `"cores"`

### Why?
The previous static default (`chunks=100`) was brittle and not adaptive to dataset size or hardware. The new behaviour preserves explicit chunk control while offering a safer default and a CPU-aware heuristic.

### Notes
- `chunks="auto"` defers chunk layout to Dask internals
- `chunks="cores"` picks row partitioning based on `os.cpu_count()`
- Existing explicit chunk behaviour remains unchanged

Closes: #65

### DIFF
--- a/dask_ml/datasets.py
+++ b/dask_ml/datasets.py
@@ -1,3 +1,4 @@
+import os
 import numbers
 from datetime import timedelta
 
@@ -23,12 +24,23 @@ def _check_axis_partitioning(chunks, n_features):
         raise ValueError(msg)
 
 
+def _normalize_chunks(chunks, shape, dtype=None):
+    if chunks is None or chunks == "auto":
+        chunks = ("auto", shape[1])
+    elif chunks == "cores":
+        n_cores = os.cpu_count() or 1
+        target_blocks = max(1, min(shape[0], n_cores * 4))
+        row_chunk = max(1, shape[0] // target_blocks)
+        chunks = (row_chunk, shape[1])
+    return da.core.normalize_chunks(chunks, shape, dtype=dtype)
+
+
 def make_counts(
     n_samples=1000,
     n_features=100,
     n_informative=2,
     scale=1.0,
-    chunks=100,
+    chunks=None,
     random_state=None,
 ):
     """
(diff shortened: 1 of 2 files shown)


TomAugspurger:
Thanks or the PR.

Can you say a bit more about the motivation? In my experience, these dataset generators are primarily used for testing and demonstration. Are you using them for something else?

> The previous static default (chunks=100) was brittle

What do you mean by brittle?

IMO, this doesn't seem worth the complexity.

Jarvis2001:
Thanks for the review!

> Can you say a bit more about the motivation? In my experience, these dataset generators are primarily used for testing and demonstration. Are you using them for something else?

You're right that these generators are primarily used for testing and demos. However, the original issue (from 2017) was that users found it cumbersome always to specify `chunks=100` or similar. The goal was to provide convenience without hardcoding a potentially poor default.

>What do you mean by brittle?

By "brittle," I meant that `chunks=100` doesn't adapt well:
- For large datasets (e.g., `n_samples=1e6`), 100 rows per chunk creates too many small tasks
- For small datasets or memory-constrained environments, it might be unnecessarily large
- There's no single number that works well across all scenarios

Even for testing/demos, the new options provide value:
- `chunks="auto"` lets Dask choose reasonable defaults based on data size
- `chunks="cores"` provides a hardware-aware heuristic (e.g., more chunks on multi-core systems)
- `chunks=None` (default) maintains the same behaviour as before.

> IMO, this doesn't seem worth the complexity.

The implementation is minimal:
- One helper function (`_normalize_chunks`)
- Backwards-compatible API
- No breaking changes to existing code

If you'd prefer to keep it simple, we could add `chunks=None` with Dask auto chunking and skip the "cores" option. But I think the current approach strikes a good balance between convenience and control.

